PR #4, opened 2016-05-27
Suggestion to add replace option to Projector#attach()

novemberborn:
Perhaps it'd be useful to expose [Maquette's `replace` method](http://maquettejs.org/docs/typedoc/interfaces/_maquette_.projector.html#replace) to the projector. Changes `Projector#attach()` to take an options object.

Ideally we remove the previous implementation.

I'm not sure what (if anything) the `append` option to `createProjector()` does. It doesn't seem to be referenced in the `attach` method.

### DIFF
--- a/src/projector.ts
+++ b/src/projector.ts
@@ -13,6 +13,8 @@ import createParentMixin, { ParentMixin, ParentMixinOptions, Child } from './mix
 /* maquette polyfills changed from 2.2 to 2.3 */
 global.requestAnimationFrame = global.requestAnimationFrame || global.window.requestAnimationFrame;
 
+export type AttachType = 'append' | 'merge' | 'replace';
+
 export interface ProjectorOptions extends ParentMixinOptions<Child>, EventedOptions {
 	/**
 	 * The root element for the projector
@@ -20,16 +22,14 @@ export interface ProjectorOptions extends ParentMixinOptions<Child>, EventedOpti
 	root?: Element;
 
 	/**
-	 * If `true`, automatically attach to the DOM during creation
+	 * If set, automatically attach to the DOM during creation
 	 */
-	autoAttach?: boolean;
+	autoAttach?: boolean | AttachType;
+}
 
-	/**
-	 * If `true`, append instead of merge when attaching to the projector to the DOM
-	 *
-	 * Only applies if `autoAttach` is `true`
-	 */
-	append?: boolean;
+export interface AttachOptions {
+	type?: AttachType;
+	tagName?: string;
 }
 
 export interface ProjectorMixin {
@@ -49,7 +49,7 @@ export interface ProjectorMixin {
 	 * @param tagName If `append` is `true` then `tagName` will be used to determine what tag name
 	 *                is used to append to the root element. Defaults to `div`.
 	 */
-	attach(append?: boolean, tagName?: string): Handle;
+	attach(options?: AttachOptions): Handle;
 
 	/**
 	 * Inform the projector that it is in a dirty state and should re-render.  Calling event handles will automatically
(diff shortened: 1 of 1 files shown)


novemberborn:
Used in <https://github.com/dojo/app/pull/5>.

src/projector.ts
```diff
@@ -29,9 +29,17 @@ export interface ProjectorOptions extends ParentMixinOptions<Child>, EventedOpti
 	 *
 	 * Only applies if `autoAttach` is `true`
 	 */
+	// FIXME Is this used? What about replace?
```
kitsonk: Yes it is used in the initializer... it gets passed to `attach`, though potentially this should be changed to be `AttachOptions` now.

src/projector.ts
```diff
@@ -49,7 +57,10 @@ export interface ProjectorMixin {
 	 * @param tagName If `append` is `true` then `tagName` will be used to determine what tag name
 	 *                is used to append to the root element. Defaults to `div`.
 	 */
+	attach(options?: AttachOptions): Handle;
+	// FIXME Remove old attach() overload
 	attach(append?: boolean, tagName?: string): Handle;
+	attach(append?: boolean | AttachOptions, tagName?: string): Handle;
```
kitsonk: Go ahead and break this API backwards... just go with `options?: AttachOptions` as the signature.

src/projector.ts
```diff
@@ -29,9 +29,17 @@ export interface ProjectorOptions extends ParentMixinOptions<Child>, EventedOpti
 	 *
 	 * Only applies if `autoAttach` is `true`
 	 */
+	// FIXME Is this used? What about replace?
 	append?: boolean;
 }
 
+export interface AttachOptions {
+	append?: boolean;
```
kitsonk: Why not do this:

```typescript
export enum AttachType {
    Append,
    Merge,
    Replace
}

export interface AttachOptions {
    type: AttachType;
    tagName?: string;
}

/* or */
export interface AttachOptions {
    type: 'append' | 'merge' | 'replace';
    tagName?: string;
}
```

src/projector.ts
kitsonk: Or `ProjectorOptions` also extends `AttachOptions`, which would flatten the options.

src/projector.ts
novemberborn: Importing the enums seems annoying, so I prefer the second suggestion. If the options are optional though then `type` should be optional too (but default to `merge`).

src/projector.ts
novemberborn: `autoAttach` could be set to `'append' | 'merge' | 'replace'` and do its thing when set?

src/projector.ts
kitsonk: Yes, a default behaviour makes more sense.  I guess logically you could want to supply a `tagName` without being explicit about the attachment type.

src/projector.ts
kitsonk: or `boolean | 'append' | 'merge' | 'replace'`? Where `true` and `'merge'` are equivalent?

novemberborn: That's a lot of options…

kitsonk: well, there aren't boolean literals (yet, they just said they were going to add them to TS2 today), so the only way to be explicit about `autoAttach` without `boolean` would be to do `autoAttach: undefined` and even then, `strictNullChecks` in TS2 would disallow that.  It is nice to have the option to be explicit instead of implicit.

novemberborn:
@kitsonk updated.

kitsonk:
I am unclear now if this is ready for merging... You have 3 commits post saying it was updated.

The main reason we are still getting CI errors has to do with SauceLabs Edge being slow and blowing up the async tests.  In my branch, I have extended the timeouts while we evaluate the approach for some of the tests.  So basically the failures you are seeing are not the droids you are looking for.

kitsonk:
I suspect you pushed those commits to the wrong branch, because dealing with also changing the polyfill behaviour should be a separate PR... 😄

novemberborn:
@kitsonk focus restored 😄

Issue #6, opened 2016-06-09
Classes on CachedRenderMixin


kitsonk:
CachedRenderMixin does not properly toggle classes... so if you drop a class out of the `string[]` it will drop it out of the VNode hash, but Maquette will not realise it has gone missing on subsequent renders... the CachedRenderMixin needs to persist those across renders and drop them out by setting them to false.

PR #7, opened 2016-06-13
Support removing classes from Widgets 

agubler:
Add support to remove classes previously applied to widgets via the widget's state.

Addresses issue #6

### DIFF
--- a/src/mixins/createCachedRenderMixin.ts
+++ b/src/mixins/createCachedRenderMixin.ts
@@ -104,6 +104,11 @@ const shadowClasses = new WeakMap<CachedRenderMixin<CachedRenderState>, string[]
  */
 const shadowStyles = new WeakMap<CachedRenderMixin<CachedRenderState>, StylesHash>();
 
+/**
+ * A weak map of the historic classes associated to a specific widget
+ */
+const widgetClassesMap = new WeakMap<CachedRenderMixin<CachedRenderState>, string[]>();
+
 const createCachedRenderMixin: CachedRenderFactory = createStateful
 	.mixin(createRenderable)
 	.mixin({
@@ -121,9 +126,15 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
 					props[key] = cachedRender.listeners[key];
 				}
 				const classes: { [index: string]: boolean; } = {};
+				const widgetClasses: string[] = widgetClassesMap.get(cachedRender) || [];
+
+				widgetClasses.forEach((c) => classes[c] = false);
+
 				if (cachedRender.classes) {
 					cachedRender.classes.forEach((c) => classes[c] = true);
 				}
+
+				widgetClassesMap.set(cachedRender, cachedRender.classes || []);
 				props.classes = classes;
 				props.styles = cachedRender.styles || {};
 				props.key = cachedRender;
--- a/src/mixins/createCachedRenderMixin.ts
+++ b/src/mixins/createCachedRenderMixin.ts
@@ -115,6 +115,11 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
 		mixin: createVNodeEvented,
 		initialize(instance: CachedRenderMixin<CachedRenderState>) {
 			instance.own(instance.on('statechange', () => { instance.invalidate(); } ));
+			instance.own({
+				destroy() {
+					widgetClassesMap.delete(instance);
+				}
+			});
 		}
 	})
 	.mixin({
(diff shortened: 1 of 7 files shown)


src/mixins/createCachedRenderMixin.ts
```diff
@@ -121,9 +126,15 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
...
 				const classes: { [index: string]: boolean; } = {};
+				const widgetClasses: string[] = widgetClassesMap.get(cachedRender) || [];
+
+				widgetClasses.forEach((c) => classes[c] = false);
+
 				if (cachedRender.classes) {
 					cachedRender.classes.forEach((c) => classes[c] = true);
 				}
+
+				widgetClassesMap.set(cachedRender, cachedRender.classes || []);
```
kitsonk: Because you are `|| []` above, it will never be `|| []` here, right?

src/mixins/createCachedRenderMixin.ts
```diff
@@ -121,9 +126,15 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
 					props[key] = cachedRender.listeners[key];
 				}
 				const classes: { [index: string]: boolean; } = {};
+				const widgetClasses: string[] = widgetClassesMap.get(cachedRender) || [];
```
kitsonk: You could set this in the initializer (where you know it should be just `[]` as well as `.own` a handle to remove it).

src/mixins/createCachedRenderMixin.ts
agubler: The above `|| []` is for the `widgetClasses` not the `cachedRender.classes` which is externally defined in the widget's state (could be undefined), this could however be moved into the `if (cachedRender.classes)` above.

src/mixins/createCachedRenderMixin.ts
agubler: Instead of a `WeakMap` let each widget instance manage and own their array of previous classes?

kitsonk: No, just in the initializer do something like this:

```typescript
(instance, options) => {
    widgetClassesMap.set(instance, []);
    instance.own({
        destroy() {
            widgetClassesMap.delete(instance);
        }
    });
}
```

And then you can always just do the following here:

```typescript
const widgetClasses = widgetClassesMap.get(cachedRender);
```

And then you don't ever need to do a `widgetClassesMap.set` below, because you are dealing with a primitive.

src/mixins/createCachedRenderMixin.ts
kitsonk: Ah, I didn't catch that, but don't you want the list to continue to persist from render to render... You never want it to drop out, so shouldn't you be re-using the array you got from above and just pushing any uniques onto it?

src/mixins/createCachedRenderMixin.ts
agubler: That's actually what I'd done yesterday, watch this space.

src/mixins/createCachedRenderMixin.ts
agubler: I couldn't think of a reason to keep anything other than the active classes from the previous render.

src/mixins/createCachedRenderMixin.ts
```diff
@@ -121,9 +131,15 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
 					props[key] = cachedRender.listeners[key];
 				}
 				const classes: { [index: string]: boolean; } = {};
+				const widgetClasses: string[] = widgetClassesMap.get(cachedRender);
```
kitsonk: Why the type assertion here?

src/mixins/createCachedRenderMixin.ts
```diff
@@ -121,9 +131,15 @@ const createCachedRenderMixin: CachedRenderFactory = createStateful
...
+
+				widgetClasses.forEach((c) => classes[c] = false);
+
 				if (cachedRender.classes) {
 					cachedRender.classes.forEach((c) => classes[c] = true);
+					widgetClassesMap.set(cachedRender, cachedRender.classes);
```
kitsonk: The classes should persist on every render once they are declared, otherwise I think we will run into issues with maquette and I don't think this reflects this... for example current behaviour (where each line is a render):

```
[ 'foo' ] => classes: { 'foo': true }
[ 'bar' ] => classes: { 'foo': false, 'bar': true }
[ 'baz' ] => classes: { 'bar': false, 'baz': true }
[ 'foo' ] => classes: { 'baz': false, 'foo': true }
```

But I think we need:

```
[ 'foo' ] => classes: { 'foo': true }
[ 'bar' ] => classes: { 'foo': false, 'bar': true }
[ 'baz' ] => classes: { 'foo': false, 'bar': false, 'baz': true }
[ 'foo' ] => classes: { 'foo': true, 'bar': false, 'baz': false }
```

src/mixins/createCachedRenderMixin.ts
agubler: oops

src/mixins/createCachedRenderMixin.ts
agubler: I don't think that we need to track the history prior to `render()` - 1, as the classes have already been removed and therefore are not required in the `classes` object and the first example will behave as we expect.

I wrote a functional test to prove this given the scenario above that classes are added and  removed as expected, please have a look and let me know any thoughts.

kitsonk: Ok, if it doesn't cause a problem, it doesn't cause a problem. 👍

PR #9, opened 2016-06-14
Add npm-debug.log to .gitignore

agubler:


kitsonk:
These sort of things should be added to your own `git` `core.excludesfile` and if we were to add it, it should go into `dojo2-package-template`.

agubler:
okay then I think we should add it to the `dojo2-package-template` then

PR #12, opened 2016-06-15
Return a promise from Projector#attach(), support afterCreate option

novemberborn:
See usage in <https://github.com/dojo/app/pull/5/commits/8181b989d6194588c9ee5d6b46859b84a3cfe2e9>.

When the promise fulfils you know the projector has actually been attached to the DOM. Previously it would return a handle but attach asynchronously.

You can specify an `afterCreate` callback which is called synchronously after the projector has been attached, allowing you to make further changes before the DOM is painted.

### DIFF
--- a/src/projector.ts
+++ b/src/projector.ts
@@ -5,6 +5,7 @@ import global from 'dojo-core/global';
 import { Handle } from 'dojo-core/interfaces';
 import { assign } from 'dojo-core/lang';
 import { queueTask } from 'dojo-core/queue';
+import Promise from 'dojo-core/Promise';
 import WeakMap from 'dojo-core/WeakMap';
 import createVNodeEvented, { VNodeEvented } from './mixins/createVNodeEvented';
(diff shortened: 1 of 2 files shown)


kitsonk:
Thinking about this, would it be a better way to provide the `afterCreate` is to emit a `aftercreate` event, similiar to the `schedulerender`?  You could always then provide the the listener to Maquette and then anyone interested in listening for it could register for it.  Since the Evented mixin is already part of this, from a consumption, you would just do a `projector.on('aftercreate', listener);`.

novemberborn:
> Thinking about this, would it be a better way to provide the afterCreate is to emit a aftercreate event, similiar to the schedulerender?

Better than the `afterCreate` option, you mean?

kitsonk:
> Better than the `afterCreate` option, you mean?

Yes.  I think it is somewhat limiting to have to decide that at creation (and no way to "stop" listening) and the `Evented` model is more aligned to the other patterns we have been using.

novemberborn:
@kitsonk went for an `attach` event.

PR #17, opened 2016-06-23
Fix issue when rendering a falsey form widget value

### LINKED ISSUE #16: Clearing value a widgets' store does not reflect in the rendered widget
Unable to clear the `value` of when using `createTextInput` in combination with a `store` using `store.patch`

Example: `.patch({ id: 'widget-id', value: '' });`

Does not clear the existing value in the widget on the `render()`

kitsonk:
Fixes #16


agubler:
👍

PR #19, opened 2016-06-28
Expect a registry provider to provide a registry

novemberborn:
Rather than creating widgets with a particular registry, expect a registry provider to be passed. Widgets can access the required registry via this provider.

See also <https://github.com/dojo/app/pull/18>.

### DIFF
--- a/src/mixins/interfaces.ts
+++ b/src/mixins/interfaces.ts
@@ -1,4 +1,5 @@
 import { Handle } from 'dojo-core/interfaces';
+import Promise from 'dojo-core/Promise';
 import { List, Map } from 'immutable';
 import { Renderable } from './createRenderable';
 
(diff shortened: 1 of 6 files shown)


src/mixins/interfaces.ts
```diff
@@ -20,3 +21,38 @@ export interface Parent {
...
+/**
+ * Provides access to read-only registries.
+ */
+export interface RegistryProvider<T> {
+	/**
+	 * Get a registry. Different widgets expect different types of registries.
+	 *
+	 * Implementations should throw if the given type is not supported.
+	 */
+	get(type: 'actions' | 'stores' | 'widgets'): Registry<T>;
```
novemberborn: @kitsonk rather than making typing assumptions for actions, stores and widgets, this approach lets each mixin define what kind of registry provider it receives. Downside though is that without using union types the mixin can get only one kind of registry out of the provider, and even if union types are used then the mixin needs to typecast each registry…

Perhaps each type could have its interface, and union typing to combine provider functionality for mixins that need multiple registries? I'm also not sure whether the current approach precludes stateful children and listeners mixins to be combined… will investigate.

novemberborn: > I'm also not sure whether the current approach precludes stateful children and listeners mixins to be combined… will investigate.

Seems like that's still allowable.

kitsonk:
👍

PR #24, opened 2016-07-10
Add createChild() to createStatefulChildrenMixin

kitsonk:
This PR allows instances of `createStatefulChildrenMixin` to be able to create children that are intended to be managed externally (like via `dojo/app`).  Specifically it assumes that `widget` registries return not only contain `get` and `identify` but also `create()` with the signature of:

```typescript
export interface CreatableRegistry<T extends Child> extends Registry<T> {
	/**
	 * Realize a child of the specified parent, returning a promise which resolves with
	 * a tuple that contains the ID and the realized instance.
	 *
	 * @param parent The parent where the realized child should be attached to
	 * @param factory The factory that should be used to realize the child
	 * @param options Any options that should be passed to the factory when realizing the child
	 */
	create<U extends T, O>(parent: { children: Map<string, Child> | List<Child>; }, factory: ComposeFactory<U, O>, options?: O): Promise<[ string | symbol, U ]>;
}
```

It then add the method `createChild()` which has the following signature:

```typescript
export type StatefulChildren<C extends Child, S extends StatefulChildrenState> = Stateful<S> & {
	/**
	 * Creates an instance based on the supplied factory and adds the child to this parent
	 * returning a promise which resolves with the ID and the instace.
	 */
	createChild<D extends C, O>(factory: ComposeFactory<D, O>, options?: O): Promise<[string | symbol, D]>;
}
```

It is mainly indented to be used during construction when creating a widget parent which is tightly coupled to it's child widget but is being managed by something like `dojo/app`.  For example, the following can be done in an initalizer:

```typescript
{
	mixin: createStatefulChildrenMixin,
	initialize(instance) {
		instance.createChild(createRenderable, <RenderableOptions> {
			render() {
				return h('div');
			}
		});
	}
}
```

### DIFF
--- a/src/mixins/createStatefulChildrenMixin.ts
+++ b/src/mixins/createStatefulChildrenMixin.ts
@@ -1,6 +1,6 @@
 import compose, { ComposeFactory } from 'dojo-compose/compose';
 import createEvented from 'dojo-compose/mixins/createEvented';
-import createStateful, { Stateful, StatefulOptions, StateChangeEvent } from 'dojo-compose/mixins/createStateful';
+import createStateful, { Stateful, StatefulOptions, StateChangeEvent, State } from 'dojo-compose/mixins/createStateful';
 import Map from 'dojo-shim/Map';
 import Promise from 'dojo-shim/Promise';
 import WeakMap from 'dojo-shim/WeakMap';
@@ -22,6 +22,10 @@ export interface StatefulChildrenOptions<C extends Child, S extends StatefulChil
 	registryProvider?: RegistryProvider<C>;
 }
 
+export interface CreateChildrenMap<C extends Child> {
+	[label: string]: [ComposeFactory<C, StatefulOptions<State>>, StatefulOptions<State>];
+}
+
 export type StatefulChildren<C extends Child, S extends StatefulChildrenState> = Stateful<S> & {
 	/**
 	 * The children that are associated with this widget
@@ -31,8 +35,20 @@ export type StatefulChildren<C extends Child, S extends StatefulChildrenState> =
 	/**
 	 * Creates an instance based on the supplied factory and adds the child to this parent
 	 * returning a promise which resolves with the ID and the instace.
+	 *
+	 * @param factory The factory to use to create the child
+	 * @param options Any options that should be used when creating the child
 	 */
-	createChild<D extends C, O extends { id?: string }>(factory: ComposeFactory<D, O>, options?: O): Promise<[string, D]>;
+	createChild<D extends C, O extends StatefulOptions<S>, S extends State>(
+		factory: ComposeFactory<D, O>, options?: O): Promise<[string, D]>;
+
+	/**
+	 * Creates a set or map of instances based upon the supplied data
+	 *
+	 * @param children the set or map of children factories and options
+	 */
+	createChildren(children: [ComposeFactory<C, StatefulOptions<State>>, StatefulOptions<State>][]): Promise<[string, C][]>;
+	createChildren(children: CreateChildrenMap<C>): Promise<{ [label: string]: [string, C]}>
 
 	/**
 	 * The ID for this widget
--- a/src/mixins/createStatefulChildrenMixin.ts
+++ b/src/mixins/createStatefulChildrenMixin.ts
@@ -1,11 +1,11 @@
-import { ComposeFactory } from 'dojo-compose/compose';
+import compose, { ComposeFactory } from 'dojo-compose/compose';
 import createEvented from 'dojo-compose/mixins/createEvented';
 import createStateful, { Stateful, StatefulOptions, StateChangeEvent } from 'dojo-compose/mixins/createStateful';
 import Map from 'dojo-shim/Map';
 import Promise from 'dojo-shim/Promise';
 import WeakMap from 'dojo-shim/WeakMap';
 import { List, Map as ImmutableMap } from 'immutable/immutable';
-import { Child, ChildListEvent, Registry, RegistryProvider } from './interfaces';
+import { Child, ChildListEvent, CreatableRegistry, RegistryProvider } from './interfaces';
 import { isList } from '../util/lang';
 
 export interface StatefulChildrenState {
@@ -17,18 +17,27 @@ export interface StatefulChildrenOptions<C extends Child, S extends StatefulChil
 }
 
 export type StatefulChildren<C extends Child, S extends StatefulChildrenState> = Stateful<S> & {
+	/**
+	 * The children that are associated with this widget
+	 */
 	children: List<C> | ImmutableMap<string, C>;
+
+	/**
+	 * Creates an instance based on the supplied factory and adds the child to this parent
+	 * returning a promise which resolves with the ID and the instace.
+	 */
+	createChild<D extends C, O>(factory: ComposeFactory<D, O>, options?: O): Promise<[string | symbol, D]>;
 }
 
-export interface StatefulChildrenMixinFactory extends ComposeFactory<Stateful<StatefulChildrenState>, StatefulChildrenOptions<Child, StatefulChildrenState>> {
-	<C extends Child>(options?: StatefulChildrenOptions<C, StatefulChildrenState>): Stateful<StatefulChildrenState>;
+export interface StatefulChildrenMixinFactory extends ComposeFactory<StatefulChildren<Child, StatefulChildrenState>, StatefulChildrenOptions<Child, StatefulChildrenState>> {
+	<C extends Child>(options?: StatefulChildrenOptions<C, StatefulChildrenState>): StatefulChildren<C, StatefulChildrenState>;
 }
 
 interface ManagementState {
 	cache?: Map<string, Child>;
 	generation?: number;
 	current?: List<string>;
-	registry: Registry<Child>;
+	registry: CreatableRegistry<Child>;
 }
 
 /**
@@ -135,7 +144,17 @@ function manageChildrenState(evt: ChildListEvent<any, Child>) {
 	}
 }
 
-const createStatefulChildrenMixin: StatefulChildrenMixinFactory = createStateful
+const createStatefulChildrenMixin = compose({
+		createChild<C extends Child, O>(factory: ComposeFactory<C, O>, options?: O): Promise<[string | symbol, C]> {
+			const stateful: StatefulChildren<Child, StatefulChildrenState> = this;
+			if (managementMap.has(stateful)) {
+				const { registry } = managementMap.get(stateful);
+				return registry.create(stateful, factory, options);
+			}
+			return Promise.reject(new Error('Unable to create child, unable to resolve registry'));
+		}
+	})
+	.mixin(createStateful)
 	.mixin({
 		mixin: createEvented,
 		initialize(instance: StatefulChildren<Child, StatefulChildrenState>, { registryProvider }: StatefulChildrenOptions<Child, StatefulChildrenState> = {}) {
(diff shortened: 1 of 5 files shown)


kitsonk:
@agubler @matt-gadd @novemberborn can I get a 👍 or feedback on this?

agubler:
@kitsonk to be honest the principal seems sound, I was waiting for @novemberborn to review the proposal and the how it will be supported within `dojo/app`.

kitsonk:
Ok, cool... I shall just await @novemberborn's magic.

novemberborn:
I don't get it. It seems to me the parent is passing the factory and the options to the registry, and itself… why can't it create and attach the child itself?

> create children that are intended to be managed externally (like via dojo/app).

Widget factories in `dojo/app` can only ever create a single instance (which is given the `id` the factory is registered with). Custom element factories can create multiple instances, but `dojo/app` doesn't really manage them.

Also, `dojo/app` will lazily create the widget the first time you `get()` it.

kitsonk:
> why can't it create and attach the child itself?

Then the app would have no visibility of it for maintain the application description (e.g. setup state in the widget state store and generate an ID).  The `app` (or registry if you will) should generate an ID and append that ID to the state of the passed parent.  The parent will then request that widget when it manages it children.

> Also, `dojo/app` will lazily create the widget the first time you `get()` it.

We could either consider `create()` an implicit `get()` and only return the `id` instead of the instance.

novemberborn:
Ah, so the parent knows it needs to create children, but it doesn't have IDs for them. `Registry#create()` would be an inversion of control, allowing the registry to generate an ID, create the instance, and track it internally. In the case of `dojo/app` it could also link the `stateFrom` option to the default store. Importantly the parent controls the factory so it doesn't have to be registered separately.

If that's the right interpretation then that makes sense to me. This is also a good use case for Symbol identifiers, except that AFAICT that doesn't work with `createStateful`?

kitsonk:
> If that's the right interpretation then that makes sense to me.

Yes, your summary is much better than mine, but that is essentially it.  Nor does the parent "care" what the ID is.  It is simply trying to express that it wants to create something that it expects to be "managed" for it.  The problem we ran into with ToDOMVC was that there child widgets were isolated and the parent had to do all the management, like figure out what store to put the state into, etc...  All things it shouldn't concern itself with.

> This is also a good use case for Symbol identifiers, except that AFAICT that doesn't work with `createStateful`?

The challenge is that Symbols are not serializable to JSON.  I think it is bigger than just `createStateful` persay, it would be how could we reliably persist symbols that they could be referencable between sessions, right?

It brings up an interesting situation, if we persist the state of the application some place and restore it later, we wouldn't want widgets to recreate their children (or we would want `app` to ignore them).  Does this mean that the parent should reference its children via a local reference?

novemberborn:
OK I've updated https://github.com/dojo/app/issues/22 with requirements.

Can't `create()` just return the instance. The ID is available through `.id` right?

---

> The challenge is that Symbols are not serializable to JSON. I think it is bigger than just createStateful persay, it would be how could we reliably persist symbols that they could be referencable between sessions, right?

Ah right. Why are we bothering supporting symbols in registries then?

> It brings up an interesting situation, if we persist the state of the application some place and restore it later, we wouldn't want widgets to recreate their children (or we would want app to ignore them). Does this mean that the parent should reference its children via a local reference?

I suppose any state with dynamically generated IDs shouldn't be persisted. Symbols would help signal that of course.

kitsonk:
> Can't `create()` just return the instance. The ID is available through `.id` right?

I guess it could, but we are then tightly coupling the widget state ID to the registry ID.  To this point, it was only depended upon [here](https://github.com/dojo/widgets/blob/75524333478f9809b571025a47452efc1929a231/src/mixins/createCachedRenderMixin.ts#L207).

> Ah right. Why are we bothering supporting symbols in registries then?

Because I am ignorant and went symbol happy?!

> I suppose any state with dynamically generated IDs shouldn't be persisted. Symbols would help signal that of course.

Well... Lets think about it, if we wanted to "deflate" the state of an active application and "inflate" it later (potentially in a different device) wouldn't we want to preserve the state of these dynamically created states?  Which then led me to the conjecture of the app/registry using that state instead of creating a new one, which would mean the registry API would need to support some sort of label for these to figure out what state to re-use?

novemberborn:
> I guess it could, but we are then tightly coupling the widget state ID to the registry ID.

That's what the app factory already does though.

> Because I am ignorant and went symbol happy?!

I'm happy to be symbol happy.

> Well... Lets think about it, if we wanted to "deflate" the state of an active application and "inflate" it later (potentially in a different device) wouldn't we want to preserve the state of these dynamically created states? Which then led me to the conjecture of the app/registry using that state instead of creating a new one, which would mean the registry API would need to support some sort of label for these to figure out what state to re-use?

Yea… and the parent needs to know there is preserved state, and things suddenly get very difficult. I don't think that would work without the parent controlling the identifiers.

kitsonk:
> I don't think that would work without the parent controlling the identifiers.

The only challenge with that though is how does the parent know they are unique identifiers, unless the app factory some how scoped the IDs for the created widgets and then left the ones it created in its scope.

Could we use a magic delimiter?  Like maybe `:` or `/`?  It would increase the complexity of resolving IDs for the app factory, but it would allow something like this:

```typescript
interface WidgetRegistry {
	get(id: string | symbol, reference?: Parent): Promise<T>;
}

const widgetRegistry = registryProvider.get('widgets');
const parent = createWidget({ id: 'parent' });
widgetRegistry.create(parent, 'widget', createWidget)
    .then((widget) => {
        console.log(widget.id); // logs `parent:widget`
        return widgetRegistry.get('widget', parent);
    })
    .then((widget) => {
        console.log(widget.id); // logs `parent:widget`
    });
```

That also means it could be represented in the state like this:

```typescript
{ id: 'parent', children: [ 'widget' ] }
```

The only thing that is missing here though is how to serialize the factory in a meaningful way so that everything could be re-instantiated again.

novemberborn:
> The only challenge with that though is how does the parent know they are unique identifiers

I don't know. That's up to the developer to get right. Scoping would help, yes, but can just be a development pattern.

My take is that if a widget wants to create children as part of its initialization, and it wants their state to survive serialization, it needs to manage the IDs of those children, and perhaps even store state as to which children to instantiate. After deserialization the widget must still be responsible for restoring its children. Anything else is magic and too complicated.

The `create()` proposal at the top here would allow all this, assuming the widget passes the desired ID in the `options` object.

novemberborn:
@kitsonk why type the parent with `{ children: Map<string, Child> | List<Child>; }`? Wouldn't it be better to type it with the `append()` method? That way the registry doesn't have to know how to correctly add the new widget.

kitsonk:
> Wouldn't it be better to type it with the `append()` method?

A two step process?

```typescript
widgetRegistry.create(createWidget)
    .then((child) => instance.append(child));
```

How would this effect the ability of the app factory to assign a unique ID and manage it?

novemberborn:
I was suggesting the  app factory would call `append()`, instead of messing with `children`. It does seem nice though if the parent takes care of appending. Order might be important.

kitsonk:
The only problem with that, is that `append()` only accepts a reference, which currently then results in a `registry.identify(child)` due to only trusting the registry to know the ID.  There was no direct assumption that `widget.state.id === registry.identify(widget)`.

novemberborn:
Turns out the app factory doesn't set `widget.state.id` to the generated ID, see <https://github.com/dojo/app/blob/3e6bc359aff913885a98b986f71540ff128c30a1/src/_realizeCustomElements.ts#L384:L386>. There's a `TODO` question around that but the behavior seems sensible.

So I'd be OK with returning the registry ID along with the child instance. And then the parent should be able to use `append()`, right?

src/mixins/createStatefulChildrenMixin.ts
```diff
@@ -135,18 +188,52 @@ function manageChildrenState(evt: ChildListEvent<any, Child>) {
...
+					/* depending upon the construction lifecycle, the this.id may not have been properly set and will
+					 * auto-generate an ID, therefore we have copied the ID out of options, if it was present and will
+					 * use that as a base for autogenerating the child widget's ID */
+					options.id = `${id || this.id}-child-${++management.childrenUID}`;
+				}
+				return registry.create(factory, options)
+					.then(([ id, child ]): [ string, C ] => {
+						/* Because this mixin manages children in the state, we will append the child to the state
+						 * of the parent */
+						this.setState({ children: (this.state.children || []).concat([ id ]) });
```
agubler: do you think this is clearer?

```ts
const children: string[] = [...(this.state.children || []), id];
this.setState({ children });
```

src/mixins/createStatefulChildrenMixin.ts
```diff
@@ -135,18 +188,52 @@ function manageChildrenState(evt: ChildListEvent<any, Child>) {
...
+			{ registryProvider, id }: StatefulChildrenOptions<Child, StatefulChildrenState> = {}
+		) {
 			if (registryProvider) {
 				const registry = registryProvider.get('widgets');
-				managementMap.set(instance, { registry });
+				managementMap.set(instance, {
+					registry,
+					generation: 0,
+					childrenUID: 0,
+					id: id
```
agubler: just `id`?

src/mixins/createStatefulChildrenMixin.ts
matt-gadd: my 2 cents, I always prefer to hoist the logical:
```javascript
const children: string[] = this.state.children || [];
this.setState({ children: [...children, id] });
```

agubler: I think both of these are pretty clear 😄

kitsonk: And better than what I had there... I will choose!

src/mixins/createStatefulChildrenMixin.ts
kitsonk: Yup, thanks.

src/mixins/createStatefulChildrenMixin.ts
kitsonk: I chose something slightly different, because a) the contextual typing works at this code point and b) it avoids creating an unnecessary array when guarding against `undefined`.

tests/unit/mixins/createStatefulChildrenMixin.ts
```diff
@@ -349,5 +356,102 @@ registerSuite({
...
+					get(type: string) {
+						return type === 'widgets' ? registry : null;
+					}
+				}
+			});
+			assert(foo);
+			assert(p);
+			return p.then((result) => {
+				const [ id, widget ] = result;
+				assert.include(id, 'undefined-child');
```
agubler: Do we want to generate an `id` that contains `undefined`? Does this mean that there could be `id` collisions between widgets where the parent does not contain `id`?

agubler:
Just one question but otherwise 👍

tests/unit/mixins/createStatefulChildrenMixin.ts
kitsonk: This is a side effect of stubbing things... because we aren't using a parent that has an ID, it gets back `undefined` for the parent ID, but then the `id` generated here is unique in the sense of `undefined-child-1` and any other children created from this parent will be unique.  The mixin which provides widgets with an ID when none is supplied will generate them uniquely.  But I can see where it looks a bit confusing, so I will update the test to stub a widget with an ID.

matt-gadd:
👍

agubler:
just needs conflicts resolving then 👍

PR #29, opened 2016-08-03
Add placeholder HTML attribute support to TextInput

bitpshr:
This PR is more to start a conversation around HTML attribute support on various input types. While it accomplishes the goal of adding placeholder support to TextInput, I'm not sure it's the right way to go. There will be other input types that can have a placeholder, so maybe it should be a mixin. I'm also not sure the aspect-based method used here is what we will settle on.

### DIFF
--- a/src/createTextInput.ts
+++ b/src/createTextInput.ts
@@ -1,3 +1,4 @@
+import { VNodeProperties } from 'maquette/maquette';
 import { ComposeFactory } from 'dojo-compose/compose';
 import createWidget, { Widget, WidgetOptions } from './createWidget';
 import createFormFieldMixin, { FormFieldMixin, FormFieldMixinState, FormFieldMixinOptions } from './mixins/createFormFieldMixin';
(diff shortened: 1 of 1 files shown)


kitsonk:
Yes, I think it goes better in a mixin.  The [FormFieldMixin](https://github.com/dojo/widgets/blob/d4be191a9a0ddb6136f627bd4c122853b068b40d/src/mixins/createFormFieldMixin.ts#L105) might be the place, if it is reasonable to expect every FormField to take a placeholder, you can see a few other attributes that are "rendered" in there.

On the other hand, FormFields might actually be the most "variable" of the rendering of elements, and it might make sense to destructure the definition of the child attributes into something that is easier to maintain.  Just like [`render()`](https://github.com/dojo/widgets/blob/d4be191a9a0ddb6136f627bd4c122853b068b40d/src/mixins/createCachedRenderMixin.ts#L187) calls `getChildrenNodes()` and `getNodeAttributes()`, maybe there is an easier way to express functions that would define node attributes that could be "collected" by `getNodeAttributes()` and then we would avoid aspecting this function.

Maybe a widget could have an array of `nodeAttributes` where were defined something like this:

```typescript
interface NodeAttributeFunction {
    (attributes: VNodeProperties): [ key: string, value: string];
}

interface NodeAttributeMixin {
    nodeAttributes: NodeAttributeFunction[];
}
```

The only thing is that we don't have a way of statically defining these so that if you had several mixins, that you could concatenate the `nodeAttributes` into a single array.  It would all have to be managed on the initialize functions, pushing these static functions onto the array.

Of course we could change compose so that certain class properties don't get "overwritten" but get "merged", but we don't have that feature, yet.

kitsonk:
I am going to raise another PR that implements what I suggested in the above comment.

PR #31, opened 2016-08-04
Migration to TS2.0

### LINKED ISSUE #1: TypeScript 2.0 for dojo/widgets

agubler:
Dependant on has, shim, core & compose TS2.0 updates

 * [x] Upgrade to TS2.0 (beta)
 * [x] Convert to use `paths` and `baseUrl` for UMD type discovery
 * [ ] Add `strictNullChecks`
 * [x] Add `noImplicitThis `

Fixes #1

### DIFF
--- a/src/createDijit.ts
+++ b/src/createDijit.ts
@@ -119,22 +119,21 @@ function constructDijitWidget(dijit: Dijit<DijitWidget>, srcNodeRef: Node): Prom
  * is added to the flow of the real DOM
  * @param element The DOM Element that is being
  */
-function afterCreate(element: HTMLElement) {
-	const dijit: Dijit<DijitWidget> = this;
-	const dijitData = dijitDataWeakMap.get(dijit);
+function afterCreate(this: Dijit<DijitWidget>, element: HTMLElement) {
+	const dijitData = dijitDataWeakMap.get(this);
 	if (dijitData.dijitWidget) {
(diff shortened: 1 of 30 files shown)


agubler:
I'm going to say that perhaps this should go in without the `strictNullChecks` compiler option and create a task to complete this

src/createDijit.ts
```diff
@@ -166,7 +165,7 @@ function resolveCtor<D extends DijitWidget>(Ctor: DijitWidgetConstructor<D> | st
 				handle.remove();
 				reject(error);
 			});
-			const mid = Ctor;
+			const mid = <string> Ctor;
```
kitsonk: Hmmm... I am surprised with flow control that this hadn't been narrowed to `string` here.

kitsonk:
outside of the one comment, which I don't think we have to address right now 👍

src/createDijit.ts
agubler: Lol, so the `next` TS version `v2.1.0-dev.20160815` doesn't narrow the type correctly but the version `v2.1.0-dev.20160817` does.

So I updated this and it worked locally as it used the `20160817` version but for some reason travis used `20160815` version and failed.

kitsonk: As mentioned in other channels `node_modules` is cached on Travis.  So once you have a branch that has cached, you would have to clear it to get the same behaviour.

PR #35, opened 2016-08-22
Basic theme and CSS module support

bitpshr:
Figured we can start this conversation...
* PostCSS is used to generate CSS modules
* CSS module JSON is generated as TS modules
* Added height support to ResizePanel
* Added intelligent aspect to Button to allow custom classes during creation

Notes:
* Generated CSS lives at `src/themes/structural/_generated/`
* Generated CSS modules live at `src/themes/structural/modules/`
* I had to comment out the `tslint` task for now; we either need to generate lint-able modules or ignore generated TS modules.


src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
 
 const createButton: ButtonFactory = createWidget
 	.mixin({
-		mixin: createFormFieldMixin
+		mixin: createFormFieldMixin,
+		aspectAdvice: {
+			before: {
+				getNodeAttributes(this: Button, overrides: VNodeProperties = {}) {
```
kitsonk: Did we want to wait until #40 is landed and refactor this?  It should be heck of a lot easier.

src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
...
+					if (this.state.classes) {
```
kitsonk: Don't you want this to run, even if `this.state.classes` isn't true (e.g. always have `classes[css.button] = true;`)?  The code coverage analysis shows that the block below isn't being covered.

src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
...
+						const classes: { [index: string]: boolean; } = {};
+						this.state.classes.forEach((c) => classes[c] = true);
+						classes[ css.button ] = true;
```
kitsonk: Need to make it more explicit in our style guide, but array literals are spaced, but index properties are not, so `classes[css.button] = true`

src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
...
+						overrides.classes = classes;
+					}
+
+					return [overrides];
```
kitsonk: where ideally this would be spaced, `[ overrides ]`.

src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
...
+				}
+			}
+		}
 	})
 	.extend({
 		tagName: 'button',
-		type: 'button'
+		type: 'button',
+		classes: [ css.button ]
```
kitsonk: I need to think about how this works... because I can see the desire to work here, but I need to check to see if this blows away if they are set in the state, like above.  I guess it must have worked for you.

src/mixins/createTabbedMixin.ts
```diff
@@ -195,9 +196,9 @@ const createTabbedMixin: TabbedMixinFactory = createCachedRenderMixin
 
 			function getTabChildVNode(tab: TabbedChild): (VNode | string)[] {
 				const tabListeners = getTabListeners(tabbed, tab);
-				const nodes = [ h('div.tab-label', { onclick: tabListeners.onclickTabListener }, [ tab.state.label ]) ];
+				const nodes = [ h('div.' + css['tab-label'], { onclick: tabListeners.onclickTabListener }, [ tab.state.label ]) ];
```
kitsonk: String literal would be preferred over concatenation (I know we aren't 100% consistent on that yet)...

```typescript
const nodes = h(`div.${css['tab-label']}`)
```

src/mixins/createTabbedMixin.ts
```diff
@@ -222,22 +223,22 @@ const createTabbedMixin: TabbedMixinFactory = createCachedRenderMixin
 
 			children.forEach(([ , tab ], idx) => {
 				const isActiveTab = tab === activeTab;
-				if (isActiveTab || (childrenNodes[idx] && childrenNodes[idx].properties.classes['visible'])) {
+				if (isActiveTab || (childrenNodes[idx] && childrenNodes[idx].properties['data-visible'])) {
```
kitsonk: 👍

src/themes/structural/_generated/Button.css
```diff
@@ -0,0 +1,5 @@
+._button_312p3_1 {
```
kitsonk: If these are generated and dynamic, they shouldn't be checked into the repo.

src/themes/structural/modules/Button.ts
```diff
@@ -0,0 +1 @@
+export default {"button":"_button_312p3_1"};
```
kitsonk: Same for these, they shouldn't be checked in...

Also, to get around the tslint issue, if you emit:

```typescript
/* tslint:disable:object-literal-key-quotes quotemark */
export default {"button":"_button_312p3_1"};

```

It should get through tslint.

src/createButton.ts
agubler: Can you just override `getClasses` instead? adding your class to the array?

src/createButton.ts
```diff
@@ -12,11 +14,26 @@ export interface ButtonFactory extends ComposeFactory<Button, ButtonOptions> { }
...
 const createButton: ButtonFactory = createWidget
 	.mixin({
-		mixin: createFormFieldMixin
+		mixin: createFormFieldMixin,
+		aspectAdvice: {
+			before: {
+				getNodeAttributes(this: Button, overrides: VNodeProperties = {}) {
+					if (this.state.classes) {
+						const classes: { [index: string]: boolean; } = {};
+						this.state.classes.forEach((c) => classes[c] = true);
```
agubler: why do we need this again when it is in the base `getNodeAttributes`?

src/createButton.ts
agubler: I think this would get overwritten if set classes was called or classes existed in the state, but would need to check.

src/createResizePanel.ts
```diff
@@ -24,14 +26,20 @@ export interface ResizePanelMixin {
...
+	height: string;
 }
 
 export type ResizePanel = Widget<ResizePanelState> & ParentListMixin<Child> & Destroyable & ResizePanelMixin;
 
 export interface ResizePanelFactory extends ComposeFactory<ResizePanel, ResizePanelOptions> { }
 
 const resizeNodePropertiesMap = new WeakMap<ResizePanel, VNodeProperties>();
-const resizingMap = new WeakMap<ResizePanel, { width: number, clientX: number }>();
+const resizingMap = new WeakMap<ResizePanel, {
```
agubler: Think we should have type/interface for this?

Gruntfile.js
```diff
@@ -37,12 +43,41 @@ module.exports = function (grunt) {
...
+				processors: [
+					require('postcss-simple-vars')({
+						variables: require('./src/themes/structural/common')
+					}),
+					require('postcss-modules')({
+						getJSON: function(cssFileName, json) {
+							var filename = path.basename(cssFileName, '.css');
+							fs.writeFileSync(
+								'src/themes/structural/modules/' + filename + '.ts',
+								'export default ' + JSON.stringify(json) + ';'
```
bitpshr: Switched to string literals so concatenation is no longer necessary and added a new line to each generated file.

Gruntfile.js
```diff
@@ -37,12 +43,41 @@ module.exports = function (grunt) {
...
+					dest: 'src/themes/structural/_generated/'
+				} ]
+			}
 		}
 	});
 
 	grunt.registerTask('dev', [
 		'typings',
-		'tslint',
+		// TODO: uncomment this, figure out a way to ignore generated CSS TS modules
```
bitpshr: Added `/* tslint:disable:object-literal-key-quotes quotemark whitespace */` to the beginning of each generated file so we could enable the `tslint` task again.

Gruntfile.js
```diff
@@ -53,7 +88,8 @@ module.exports = function (grunt) {
 
 	grunt.registerTask('dist', [
 		'typings',
-		'tslint',
+		// TODO: uncomment this, figure out a way to ignore generated CSS TS modules
```
bitpshr: Added `/* tslint:disable:object-literal-key-quotes quotemark whitespace */` to the beginning of each generated file so we could enable the `tslint` task again.

src/createButton.ts
bitpshr: This should go away when #40 lands. #45 created so nothing gets lost.

src/createButton.ts
bitpshr: This should go away when #40 lands. #45 created so nothing gets lost.

src/createButton.ts
bitpshr: Noted.

src/createButton.ts
bitpshr: Noted.

src/createButton.ts
kitsonk: @agubler and @matt-gadd talked about this.  We think that actually using the feature in compose (see: dojo/compose#46) that for classes that are "bound" to the factory/class, that this is the right way to add these and they will should be merged and added to every render.  What we need to do though is make `classes` `readonly` so that you cannot modify instance `classes` in this fashion and only look for instance bound classes in `.state` when rendering.

So for now this is the right way, though we need to make other changes to preserve the functionality...

kitsonk: See #46 for more details/discussion

bitpshr: Noted.

src/themes/structural/_generated/Button.css
bitpshr: I ignored generated CSS files and added the `postcss` grunt task to both `dev` and `dist` to make sure it's run before use.

bitranch:
Can someone explain what problem this is trying to solve?

(Since I know my graphic artist will complain about trying to override dynamic css classes, I'd like to know that there's some benefit to this.)

src/themes/structural/modules/Button.ts
bitpshr: I ignored generated modules and added the `postcss` grunt task to both `dev` and `dist` to make sure it's run before use. I also added `/* tslint:disable:object-literal-key-quotes quotemark whitespace */` to the beginning of each generated module.

kitsonk:
> Can someone explain what problem this is trying to solve?

@bitranch First, glad to see you chiming in on Dojo 2!!!!

The problem it is trying to solve is to allow a widget developer to express dependency on any structural CSS that it needs, and ensure that CSS can be built in an optimised way (e.g. if you build a layer, you only include the structural CSS for widgets you include).

At the same time, from a widget developer point of view, we wanted them to be able to express the CSS in a modular way, just like their TypeScript code and import it in a similar way and have the tooling deal with ensuring that any internal class names don't impact the overall application.  This is why we looked to CSS Modules and PostCSS to generate them, but then are adding a bit more to the tooling so they integrate well into a TypeScript development environment that allow you to import the classes and integrate them into your widget just like other code.

This touches on that there are three concepts of CSS for Dojo 2:

- *Inline Styles* - These should be avoided at all costs, the only exception to the rule is when a widget has no choice but to change the value of a CSS property that cannot be easily described by swapping a class name.  The only case we have found so far is a panel that needs to resize, as it needs to calculate its width and height dynamically at runtime.  But the first question to ask yourself is "is there any other possible way to achieve this!?"  These are expressed as functional logic in TypeScript as part of the render process of a widget.
- *Structural* - CSS that is necessary for the widget to exhibit the right behaviours.  These should be encapsulated and not "leak" outside.  Their selectors should be written rooted around a class name which will be "namespaced" upon build.  These are expressed as CSS Modules which generate a TypeScript module which resolves the CSS Module class name to the build time class name and are added programatically to the widget module.
- *Theming* - This CSS deals with the overall look and feel.  Selectors for these that want to change a particular class of widget should be rooted on the custom tag name or `is` attribute of the widget class (e.g. `widget-button, [is="widget-button"]`).  Theme developers should avoid targeting any of the structural class names (or really avoid depending too much of the internal structure of a widget).  These are just CSS files which can be integrated into a build pipeline, using any sort of pre-post processors desired.  The Dojo 2 themes are likely to continue to be built using Stylus though SASS is also a consideration, but we don't have a strong opinion about what others will use.

We expect to provide more clear and detailed guidelines, but there were lots of challenges with Dijit that we wanted to overcome, like the CSS was all or nothing in a lot of cases, except when you did custom widgets and too many challenges were "forgetting" that sort of structural CSS, or custom widget developers not really following that pattern, making the widget CSS unmaintainable.  But we think we can provide a system that if followed will allow widget developers to work along side the UI/UX people, largely independent, but still have something that works at the end.

So if everyone does it right, your graphic designer won't ever have to be aware there are dynamically generated class names and will still be able to deliver you the look and feel you and your customers desire.

bitpshr:
Rebased and all merge conflicts resolved. Should be ready for another review.

src/createResizePanel.ts
kitsonk: Yeah, but should be dealt with as a separate issue.  I think there is other private state here that should be rationalised into a single weakmap and interface.

PR #37, opened 2016-09-03
Migrate to @types where possible

kitsonk:
This PR migrates what we can to @types, though because we still have to use "classic" module resolution, there is some strangeness which requires that dependencies to be declared in `typings.json`.

This also required a couple changes in `dojo/typings`, in particular so that Intern was better integrated with the `@types/chai`.


matt-gadd:
On the module resolution front, their isn't a lot stopping us from moving to `node` module resolution now given we've moved to the new packaging style. The only places where it would currently break are where we haven't used relative paths and instead have mapped `src` for the tests, which should be easily fixed?

kitsonk:
Fair point...

kitsonk:
Ok, this has been updated.  The other thing I realised, is we could also add `paths` to the `tsconfig.json` to map `src/**/*` to the `src` path as well to fix it, though I guess that leaks non-distribution configuration into the `tsconfig.json`, so I guess the relative paths work.  Technically we can drop the `src` from the package maps now as well.

agubler:
👍

Issue #56, opened 2016-09-28
Widget Transitions / Animations


agubler:
Maquette supports both programmatic and CSS animations so am opening this thread to discuss the best approach initially for Dojo2 within the widget architecture.

The Maquette documentation provides an example for both styles:
1. Programmatically requires the use of a 3rd party library such as Velocity.js (although has broken typings) but in the future could use the Web Animations API.
2. To support CSS animations, the consumer needs to load the `css-animations.js` provided by Maquette and then just use CSS as usual. `updateAnimations` are not supported by `maquette` when using `cssTransitions`.

Initially I think it'll will be simpler to support css transition by providing a `mixin` in Widgets that adds the transition classes to the node attributes which Maquette needs in order to perform the animations? Subsequently can look to support the programmatic API; adding sugar around which ever animations implementation is chosen? Question as to whether D2 should bundle an animation library such as `Velocity` or allow consumers to include their own write their own animation function that get attached to the Maquette Vdom?

Example code using Velocity's API

``` ts
import createWidget from 'dojo-widgets/createWidget';
import { animate } from 'velocity-animate';

const createWidgetAnimation = createWidget
    .extend({
        nodeAttributes: [
            function(this: any): VNodeProperties {
                return {
                    enterAnimation: function (domNode: any) {
                        animate(domNode, 'slideDown', 1500, 'ease-in-out');
                    }
                };
            }
        ]
    });
```

Possible mixin solution for supporting CSS transitions:

``` ts
import { VNodeProperties } from 'maquette';
import compose, { ComposeFactory } from 'dojo-compose/compose';
import { NodeAttributeFunction } from './createRenderMixin';
import createStateful, { State, Stateful, StatefulOptions } from 'dojo-compose/mixins/createStateful';

export type CssTransitionMixinState = State & {
    enterAnimation?: string;
    exitAnimation?: string;
}

export interface CssTransition {
    nodeAttributes: NodeAttributeFunction[];
}

export type CssTransitionMixin<S extends CssTransitionMixinState> = CssTransition & Stateful<S>;

export interface CssTransitionMixinFactory extends ComposeFactory<CssTransitionMixin<CssTransitionMixinState>, StatefulOptions<CssTransitionMixinState>> {};

const createCssTransitionMixin: CssTransitionMixinFactory = compose({
    nodeAttributes: [
        function (this: CssTransitionMixin<CssTransitionMixinState>): VNodeProperties {
            const { enterAnimation, exitAnimation } = this.state;
[...]

PR #57, opened 2016-09-28
Feature: Support CSS Transitions with Maquette Projector and Widget Mixin

agubler:
**Type:** feature

**Description:**

Widget mixin to support CSS transitions within the widget architecture.

Using css transitions with Maquette requires loading Maquette's provided `css-transitions.js` in the page, this is then used within the Maquette projector construction assuming the `cssTransitions` options has been set to `true`.

Currently due to limitations in the provided `css-transitions.js` script, `exitAnimation` is only supported for CSS Transitions and not CSS Animations (like those provided by `animate.css`), currently the element is not being correctly removed from the DOM on completion of the exitAnimation.

Related to #56

### DIFF
--- a/src/mixins/createCssTransitionMixin.ts
+++ b/src/mixins/createCssTransitionMixin.ts
@@ -1,0 +1,40 @@
+import { VNodeProperties } from 'maquette';
+import compose, { ComposeFactory } from 'dojo-compose/compose';
+import { NodeAttributeFunction } from './createRenderMixin';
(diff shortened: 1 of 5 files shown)


kitsonk:
The defect with exit animations is fixed in Maquette 2.4.0 correct?

Also, while it isn't critical for this PR, as the unit tests are sufficient, we should consider some sort of functional tests around this so make sure the whole thing "hangs together".

PR #59, opened 2016-09-29
Label all factories

### LINKED ISSUE #58: Integrate dojo/compose debugging features
**Enhancement** 

We should leverage the debugging features that are proposed in dojo/compose#65

kitsonk:
**Type:** feature

- [X] There is a related issue
- [X] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
- [X] Unit or Functional tests are included in the PR

**Description:**

This PR labels all factories in widgets, which can then be used for diagnostic purposes.  In many run-time environments, logging an instance to a console will be logged with the appropriate name and factories will often appear similar to the way classes appear.  Environments that do not support the ES6 for `function.name` to be configurable, they won't be renamed, and in environments that do not support `function.name` at all, the console output will not be impacted, but you can still interrogate the instances and factories directly.

~~Needs dojo/compose#65~~
Resolves #58

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,6 @@
     "jsdom": "^9.5.0",
     "remap-istanbul": "^0.6.4",
     "tslint": "^3.11.0",
-    "typescript": "^2.0.3"
+    "typescript": "~2.0.3"
   }
 }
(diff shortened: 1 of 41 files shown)


agubler:
This is extremely outdated now, closing. Needs to be re-addressed.

PR #61, opened 2016-10-01
VNodeProperties disabled attribute is typed as boolean

agubler:
**Type:** bug

**Description:**

Lint error reporting `disabled` attribute of `VNodeProperties` is typed as `boolean` not string.

Lint Error: `src/mixins/createFormFieldMixin.ts(120,6): error TS2322: Type 'string' is not assignable to type 'boolean'.`

### DIFF
--- a/src/mixins/createFormFieldMixin.ts
+++ b/src/mixins/createFormFieldMixin.ts
@@ -116,9 +116,8 @@ const createFormMixin: FormMixinFactory = compose({
 				if (this.state && this.state.name) {
 					props.name = this.state.name;
 				}
-				if (this.state.disabled) {
-					props['disabled'] = 'disabled';
-				}
+
+				props.disabled = Boolean(this.state.disabled);
 
 				return props;
 			}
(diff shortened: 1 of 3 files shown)


kitsonk:
Yeah, I caught this too when the version of Maquette gets updated: https://github.com/dojo/widgets/pull/59/files#diff-76ffc0d3dc592af5a6984bd567d24ec9R120

I was just having issues with CI on android for some reason with the changes that I am trying to chase down.

src/mixins/createFormFieldMixin.ts
```diff
@@ -116,9 +116,8 @@ const createFormMixin: FormMixinFactory = compose({
 				if (this.state && this.state.name) {
 					props.name = this.state.name;
 				}
-				if (this.state.disabled) {
-					props['disabled'] = 'disabled';
-				}
+
+				props['disabled'] = !!this.state.disabled;
```
kitsonk: Since these are now typed, you can do `props.disabled` which is better.  Also why not cast as `Boolean(this.state.disabled)` to be more explicit?

agubler: Sure, can use `Boolean` just by default use `!!`, will update.

agubler: Fixed

PR #63, opened 2016-10-03
Work around VNodeProperties attributes being marked as readonly

### LINKED ISSUE #62: Maquette 2.4.X Regressions
**Bug**

It appears that Maquette 2.4.X has introduced some regressions in our code that was unexpected.  Upon compiling Widgets now we get the following errors:

```
Using tsc v2.0.3
src/mixins/createFormFieldMixin.ts(112,6): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createFormFieldMixin.ts(115,5): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createFormFieldMixin.ts(117,6): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createFormFieldMixin.ts(120,5): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createRenderMixin.ts(230,6): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createRenderMixin.ts(231,6): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/mixins/createRenderMixin.ts(232,6): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/projector.ts(154,5): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/projector.ts(160,4): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
src/projector.ts(161,4): error TS2450: Left-hand side of assignment expression cannot be a constant or a read-only property.
```

The functional tests also appear to fail, but it is hard to tell if it is related to these complication errors or further breaking changes in Maquette.

Reverting to `maquette@2.3.7` appears to alleviate the problem.

Package Version: 2.0.0-alpha

**Code**

```
> npm install maquette@latest
```

**Expected behavior:**

It works, without issue...

**Actual behavior:**

It has issues.

agubler:
In the latest release of Maquette all the properties of `VNodeProperties` are marked as `readonly` - these changes prevent the regressions noted in issue #62.

We could also maintain a local interface the represents `VNodeProperties` but that might become a pain to maintain (although I added a partial interface for the `projector`, so am open to all suggestions)

resolves #62

### DIFF
--- a/src/mixins/createFormFieldMixin.ts
+++ b/src/mixins/createFormFieldMixin.ts
@@ -106,20 +106,10 @@ const createFormMixin: FormMixinFactory = compose({
 
 		nodeAttributes: [
 			function (this: FormFieldMixin<any, FormFieldMixinState<any>>): VNodeProperties {
-				const props: VNodeProperties = {};
+				const { type, value, state } = this;
+				const { disabled, name } = state;
 
-				if (this.type) {
-					props['type'] = this.type;
-				}
-				/* value should always be copied */
-				props.value = this.value;
-				if (this.state && this.state.name) {
-					props.name = this.state.name;
-				}
-
-				props.disabled = Boolean(this.state.disabled);
-
-				return props;
+				return { type, value, name, disabled: Boolean(disabled) };
 			}
 		]
 	}, (instance: FormField<any>, { type } = <any> {}) => {
--- a/src/mixins/createRenderMixin.ts
+++ b/src/mixins/createRenderMixin.ts
@@ -212,10 +212,8 @@ const createRenderMixin = createStateful
 
 			nodeAttributes: [
 				function (this: RenderMixin<RenderMixinState>): VNodeProperties {
-					const props: VNodeProperties = this.state && this.state.id
-						? { 'data-widget-id': this.state.id }
-						: {};
-
+					const baseIdProp = this.state && this.state.id ? { 'data-widget-id': this.state.id } : {};
+					const styles = this.styles || {};
 					const classes: { [index: string]: boolean; } = {};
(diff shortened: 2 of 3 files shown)


src/mixins/createFormFieldMixin.ts
```diff
@@ -106,20 +106,10 @@ const createFormMixin: FormMixinFactory = compose({
...
-				/* value should always be copied */
-				props.value = this.value;
-				if (this.state && this.state.name) {
-					props.name = this.state.name;
-				}
-
-				props.disabled = Boolean(this.state.disabled);
-
-				return props;
+				return { type, value, name, disabled: Boolean(disabled) };
```
kitsonk: Super 👍

src/mixins/createRenderMixin.ts
```diff
@@ -226,10 +224,7 @@ const createRenderMixin = createStateful
 						widgetClassesMap.set(this, this.classes);
 					}
 
-					props.classes = classes;
-					props.styles = this.styles || {};
-					props.key = this;
-					return props;
+					return Object.assign(baseIdProp, { key: this, classes, styles });
```
kitsonk: Ideally should use `dojo-core/lang#assign()`.

agubler: You're absolutely right (failing tests confirm that)!

agubler:
@kit I think it would be prudent to do both, work around in the short/medium term and raise an issue with Maquette highlighting the difficulties "adding" to an existing `VNodeProperties` object.

PR #74, opened 2016-10-10
Support concept of 'widgetClasses'

### LINKED ISSUE #46: Styles and Classes on Widgets
There are two main types of styles and classes that need to be dealt with when rendering a widget:

* Those bound to the class/factory
* Those that are instance based

There actually should be no class bound styles, as any styles that are provided on render should be calculated from the `instance.state` and not from the class.

Therefore to facilitate this, we need to do the following:

* Remove `.styles` from widget interfaces
* Make `.classes` a `readonly string[]` with a static value of the merged array of string using the merge array feature of `dojo/compose` (see: dojo/compose#46)
* Change the default render logic to use the `.classes` property as a base and mixin any classes provided in `instance.state.classes`.

Then when creating a class or mixin, the developer would just supply any class bound classes to be mixed in as a `string[]` in the prototype.

### LINKED ISSUE #69: Separate "class" classes and "stateful" classes
**Enhancement**

Currently, this is no clear separation between classes that are statically bound and should be included in every render and classes that are "stateful" and should change based on the widget's state.

We should consider separating these clearly in widgets and using the array compostability features of `dojo-compose` to make it easy for a widget developer to allow classes to be mixed into the prototype for the widget.

Package Version: master

agubler:
**Type:** feature

**Description:**

Added a property to support the concept of `widgetClasses`, these classes will always be applied to the any `widget` has them or extends/mixes in a `widget` that itself has `widgetClasses`.

The `widgetClasses` arrays of all the factories mixed in will be merged into a final `widgetsClasses` array that is used by `createCachedRender` to apply the classes on `render`.

Resolves #69
Resolves #46

### DIFF
--- a/src/createButton.ts
+++ b/src/createButton.ts
@@ -18,7 +18,7 @@ const createButton: ButtonFactory = createRenderMixin
 	.extend({
 		tagName: 'button',
 		type: 'button',
-		classes: [css.button]
+		widgetClasses: [ css.button ]
 	});
 
 export default createButton;
--- a/src/createButton.ts
+++ b/src/createButton.ts
@@ -18,7 +18,7 @@ const createButton: ButtonFactory = createRenderMixin
 	.extend({
 		tagName: 'button',
 		type: 'button',
-		widgetClasses: [ css.button ]
+		classes: [ css.button ]
 	});
 
 export default createButton;
--- a/src/createResizePanel.ts
+++ b/src/createResizePanel.ts
@@ -206,7 +206,7 @@ const createResizePanel: ResizePanelFactory = createWidget
 	})
 	.extend({
 		tagName: 'dojo-panel-resize',
-		classes: [ css.panel ]
+		widgetClasses: [ css.panel ]
 	})
 	.mixin({
 		mixin: createDestroyable,
--- a/src/createResizePanel.ts
+++ b/src/createResizePanel.ts
@@ -206,7 +206,7 @@ const createResizePanel: ResizePanelFactory = createWidget
 	})
 	.extend({
 		tagName: 'dojo-panel-resize',
-		widgetClasses: [ css.panel ]
+		classes: [ css.panel ]
 	})
 	.mixin({
 		mixin: createDestroyable,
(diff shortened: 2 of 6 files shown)


kitsonk:
Do you not agree with what is proposed in [dojo/interfaces/widgetBases](https://github.com/dojo/interfaces/blob/8c8e81c00a79bc974ceed14f4f87e2009302c0e2/src/widgetBases.d.ts#L385)

In that there is no longer any "coupling/shadowing" between `widget.classes` and `widget.state.classes` and that `widget.classes` are essentially static and the only way to change classes on an instance would be to modify `widget.state.classes`.

The proposal also eliminated the getter/setter `widget.styles` due to the confusion as well... essentially no shadowing of properties.

Personal opinion is I would rather take the hit on refactoring anything that is setting classes via `widget.classes` now and get rid of what I have realised is dangerous functionality in retrospect.  Introducing a new property feels like to me like an arms race and we will end up with fun stuff like `getElementById` APIs sooner or later, so if it isn't too confusing, I would prefer something terse.

kitsonk:
Also, mentioning that issue #46 described some of what I had thought was discussed verbally.  If we don't want to do that, we should describe what we would like to do.

agubler:
To be honest, I was looking at providing non-breaking functionality in order to support the concept of `widgetClasses` (or static classes) - rather than wipe everything out all at the same time.

However if you are happy to do that, I can easily remove the connection between `widget.classes` and `widget.state.classes` (and remove `widget.styles` at the same time).

The concepts are the same and will be provided to Maquette in the same way just will remove the `widgetClasses` attribute and re-use `classes`.

I actually didn't see #46, I was just working from #69 - I think functionally for "widget classes" this covers the proposed functionality... The only difference is leveraging the Maquette API that enables static classes to be appended to the `selector`, rather than some logic in the `getNodeAttributes` and to me feels cleaner as far as separation is concerned.

kitsonk:
I forgot abut #46 too... :-) until I was like "oh, wait, didn't @matt-gadd" and I discuss this and said yeah, that I was "stupid" and we should just blat that shadowing of state, because it was "dangerous".  If we have anything that is doing `instance.classes = ...` instead of patching state, I would rather flush that out into the open now.  Same for `instance.styles`.

Totally 👍 to the concept of putting those on the VNode selector.

agubler:
Okay, watch this space 👀

PR #82, opened 2016-10-18
Add createI18nMixin.

### LINKED ISSUE #81: Incorporate `dojo-i18n` into the widgeting system.
Implement a `createI18nMixin` that incorporates the `dojo-i18n` functionality into the widgeting system:

``` typescript
import { Bundle, Messages } from 'dojo-i18n/i18n';

interface I18nMixin {
  // Array of bundles to register at instantiation.
  bundles?: Bundle<Messages>;
  // Return a localized dictionary for the current locale.
  localizeBundle(bundle: Bundle<Messages>): Messages;
  // Register a bundle with the widget.
  registerBundle(bundle: Bundle<Messages>): void;
}
```

Widgets like `dojo-widgets/createButton` can specify the bundles they expect within their definitions, or during instantiation:

``` typescript
import greetings from 'nls/greetings';

createRenderMixin
  .extend({
    bundles: [ greetings ]
  });

// More explicitly with `registerBundle`
createRenderMixin
  .mixin({
    initialize(instance) {
      instance.registerBundle(greetings);
    }
  });

// At instantiation
createWidget({
  bundles: [ greetings ]
});
```

Likewise, the dictionary messages used by a widget can be specified either in `state.labels` loaded directly (e.g., within `getChildrenNodes`) via the `localizeBundle` method:

``` typescript
import greetings from 'nls/greetings';
import navigation from 'nls/navigation';

// When registering a single bundle...
createWidget({
  bundles: [ greetings ],
  labels: {
    // Once the locale-specific dictionary is loaded, `setState` will be called with
    // `{ label: `${greetings.hello}` }`
    label: 'hello'
  }
});

// When registering multiple bundles
createWidget({
  bundles: [ greetings, navigation ],
  labels: {
    // When multiple bundles have been registered, labels must be prefaced with
    // `${bundlePath}:` in order to ensure the correct bundle is used.
    label: 'nls/greetings:hello'
  }
});

// From within `getChildrenNodes`
createRenderMixin
  .mixin({
    mixin: {
      getChildrenNodes(): VNode[] {
        const messages = this.localizeBundles(greetings);
        return h('h1', { 'class': 'title' }, [ messages.hello ]);
      }
    }
  });
```

mwistrand:
Resolve #81 and #199, and also resolve dojo/i18n#16.

Implementation notes:
- `createButton` is the only component that includes `createI18nMixin` by default.
- If a widget's `state.locale` is unspecified, the `state.locale` of each parent is looked up until one is found. Otherwise, `i18n.locale` is used to determine which messages to load.
- If a widget's `state.labels` is specified, but it has not registered any bundles, then the widget's state is not updated.
- The text direction is set via `state.rtl`. If `true`, then the widget's node will have a `dir="rtl"` attribute. If `false`, the node will have a `dir="ltr"` attribute. If not a boolean value, then the node will not have a `dir` attribute at all.
- The widget node is also given a `data-locale` attribute set to the widget's `state.locale` (if exists). This can facilitate locale-specific styling for any application that may need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ For more background on dojo-widgets, there is a document describing the [widgeti
     	- [Widgets with Children](#widgets-with-children)
     - [Authoring Custom Widgets](#authoring-custom-widgets)
     - [Projector](#projector)
+    - [Internationalization](#internationalization)
     - [Dojo Widget Components](#dojo-widget-components)
 - [How Do I Contribute?](#how-do-i-contribute)
     - [Installation](#installation)
@@ -432,6 +433,57 @@ const app = createApp();
 
 app.append().then(() => {
 	console.log('projector is attached');
+});
+```
+
+### Internationalization
+
+Widgets can be internationalized by mixing in `dojo-widgets/mixins/createI18nMixin`. [Message bundles](https://github.com/dojo/i18n) are localized by passing them to `localizeBundle`. If the bundle supports the widget's current locale, but those locale-specific messages have not yet been loaded, then the default messages are returned and the widget will be invalidated once the locale-specific messages have been loaded. Each widget can have its own locale by setting its `state.locale`; if no locale is set, then the default locale as set by [`dojo-i18n`](https://github.com/dojo/i18n) is assumed.
+
+```typescript
+const createI18nWidget = createWidgetBase
+	.mixin(createI18nMixin)
+	.mixin({
+		mixin: {
+			nodeAttributes: [
+				function (attributes: VNodeProperties): VNodeProperties {
+					// Load the `greetings` messages for the current locale.
+					const messages = this.localizeBundle(greetings);
+					return { title: messages.hello };
+				}
+			],
+
+			getChildrenNodes: function () {
+				// Load the "greetings" messages for the current locale. If the locale-specific
+				// messages have not been loaded yet, then the default messages are returned,
+				// and the widget will be invalidated once the locale-specific messages have
+				// loaded.
+				const messages = this.localizeBundle(greetingsBundle);
+
+				return [
+					d(createLabel, {
+						state: {
+							// Passing a message string to a child widget.
+							label: messages.purchaseItems
+						}
+					}),
+					d(createButton, {
+						state: {
+							// Passing a formatted message string to a child widget.
+							label: messages.format('itemCount', { count: 2 })
+						}
+					})
+				];
+			}
+		}
+	});
+
+const widget = createI18nWidget({
+	state: {
+		// Set the locale for the widget and all of its children. Any child can still
+		// set its own locale.
+		locale: 'fr'
+	}
 });
 ```
 
--- a/package.json
+++ b/package.json
@@ -26,8 +26,13 @@
     "dojo-compose": ">=2.0.0-beta.10",
     "dojo-core": ">=2.0.0-alpha.14",
     "dojo-has": ">=2.0.0-alpha.4",
+    "dojo-i18n": ">=2.0.0-alpha.3",
     "dojo-shim": ">=2.0.0-beta.1",
     "maquette": ">=2.3.7 <=2.4.1"
+  },
+  "dependencies": {
+    "cldr-data": "~30.0.0",
+    "globalize": "^1.1.1"
   },
   "devDependencies": {
     "@types/chai": "3.4.*",
--- a/src/mixins/createI18nMixin.ts
+++ b/src/mixins/createI18nMixin.ts
@@ -1,0 +1,150 @@
+/* tslint:disable:interface-name */
+import compose, { ComposeFactory } from 'dojo-compose/compose';
+import { assign } from 'dojo-core/lang';
+import i18n, { Bundle, formatMessage, getCachedMessages, Messages, observeLocale } from 'dojo-i18n/i18n';
+import { VNodeProperties } from 'maquette';
+import { NodeAttributeFunction, Widget, WidgetOptions, WidgetProperties, WidgetState } from '../interfaces';
+
+export interface I18nMixin<M extends Messages> {
+	/**
+	 * An array of node attribute functions which return additional attributes that should be mixed into
+	 * the final VNode during a render call. In this particular case, the node's `dir` attribute is optionally
+	 * set when `state.rtl` is a boolean, and the node's `data-locale` attribute is set when `state.locale` is
+	 * not empty.
+	 */
+	nodeAttributes: NodeAttributeFunction<I18nWidget<M, I18nProperties>>[];
+
+	/**
+	 * Return the cached messages for the specified bundle for the current locale, assuming they have already
+	 * benn loaded. If the locale-specific messages have not been loaded, they are fetched and the widget state
+	 * is updated.
+	 *
+	 * @param bundle
+	 * The required bundle object for which available locale messages should be loaded.
+	 *
+	 * @return
+	 * The localized messages, along with a `format` method for formatting ICU-formatted templates.
+	 */
+	localizeBundle(bundle: Bundle<M>): LocalizedMessages<M>;
+}
+
+export interface I18nFactory extends ComposeFactory<I18nMixin<Messages>, WidgetOptions<WidgetState, I18nProperties>> {}
+
+export interface I18nProperties extends WidgetProperties {
+	/**
+	 * The locale for the widget. Is not specified, then the root locale (as determined by `dojo-i18n`) is assumed.
+	 * If specified, the widget's node will have a `data-locale` property set to the locale, in order to facilitate
+	 * styling localized components if the use case arises.
+	 */
+	locale?: string;
+
+	/**
+	 * An optional flag indicating the widget's text direction. If `true`, then the underlying node's `dir`
+	 * property is set to "rtl". If it is `false`, then the `dir` property is set to "ltr". Otherwise, the property
+	 * is not set.
+	 */
+	rtl?: boolean;
+}
+
+/**
+ * @private
+ * An internal helper interface for defining locale and text direction attributes on widget nodes.
+ */
+interface I18nVNodeProperties extends VNodeProperties {
+	'data-locale': string | null;
+	dir: string | null;
+}
+
+export type I18nWidget<M extends Messages, P extends I18nProperties> = I18nMixin<M> & Widget<WidgetState, I18nProperties>;
+
+export type LocalizedMessages<T extends Messages> = T & {
+	/**
+	 * Formats an ICU-formatted message template for the represented bundle.
+	 *
+	 * @param key
+	 * The message key.
+	 *
+	 * @param options
+	 * The values to pass to the formatter.
+	 *
+	 * @return
+	 * The formatted string.
+	 */
+	format(key: string, options?: any): string;
+}
+
+/**
+ * @private
+ * Return the cached dictionary for the specified bundle and locale, if it exists. If the requested dictionary does not
+ * exist, then load it and update the instance's state with the appropriate messages.
+ *
+ * @param instance
+ * The widget instance.
+ *
+ * @param bundle
+ * The bundle for which to load a locale-specific dictionary.
+ *
+ * @return
+ * The locale-specific dictionary, if it has already been loaded and cached.
+ */
+function getLocaleMessages(instance: I18nWidget<Messages, I18nProperties>, bundle: Bundle<Messages>): Messages | void {
+	const { properties } = instance;
+	const locale = properties.locale || i18n.locale;
+	const localeMessages = getCachedMessages(bundle, locale);
+
+	if (localeMessages) {
+		return localeMessages;
+	}
+
+	i18n(bundle, locale).then(() => {
+		instance.invalidate();
+	});
+}
+
+const createI18nMixin: I18nFactory = compose<I18nMixin<Messages>, WidgetOptions<WidgetState, I18nProperties>>({
+	nodeAttributes: [
+		function (this: I18nWidget<Messages, I18nProperties>, attributes: VNodeProperties): VNodeProperties {
+			const vNodeProperties = {
+				'data-locale': null,
+				dir: null
+			} as I18nVNodeProperties;
+			const { locale, rtl } = this.properties;
+
+			if (typeof rtl === 'boolean') {
+				vNodeProperties['dir'] = rtl ? 'rtl' : 'ltr';
+			}
+
+			if (locale) {
+				vNodeProperties['data-locale'] = locale;
+			}
+
+			return vNodeProperties;
+		}
+	],
+
+	localizeBundle(this: I18nWidget<Messages, I18nProperties>, bundle: Bundle<Messages>): LocalizedMessages<Messages> {
+		const { locale } = this.properties;
+		const messages = getLocaleMessages(this, bundle) || bundle.messages;
+
+		return assign(Object.create({
+			format(key: string, options?: any) {
+				return formatMessage(bundle.bundlePath, key, options, locale);
+			}
+		}), messages);
+	}
+}, (instance: I18nWidget<Messages, I18nProperties>) => {
+	const subscription = observeLocale({
+		next() {
+			if (!instance.properties.locale) {
+				instance.invalidate();
+			}
+		}
+	});
+	instance.own({
+		destroy() {
+			subscription.unsubscribe();
+		}
+	});
+});
+
+export default createI18nMixin;
--- a/tests/unit/mixins/createI18nMixin.ts
+++ b/tests/unit/mixins/createI18nMixin.ts
@@ -1,0 +1,151 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+import i18n, { invalidate, switchLocale, systemLocale } from 'dojo-i18n/i18n';
+import * as sinon from 'sinon';
+import createI18nMixin from '../../../src/mixins/createI18nMixin';
+import createWidgetBase from '../../../src/createWidgetBase';
+import bundle from '../../support/nls/greetings';
+
+const createLocalized = createWidgetBase.mixin(createI18nMixin);
+let localized: any;
+
+registerSuite({
+	name: 'mixins/createI18nMixin',
+
+	afterEach() {
+		return switchLocale(systemLocale).then(() => {
+			invalidate();
+
+			if (localized) {
+				localized.destroy();
+				localized = null;
+			}
+		});
+	},
+
+	api() {
+		const localized = createLocalized();
+		assert(localized);
+		assert.isFunction(localized.localizeBundle);
+	},
+
+	'.localizeBundle()': {
+		'Returns default messages when locale bundle not loaded'() {
+			return switchLocale('fr').then(() => {
+				localized = createLocalized();
+				const messages = localized.localizeBundle(bundle);
+
+				assert.strictEqual(messages.hello, 'Hello');
+				assert.strictEqual(messages.goodbye, 'Goodbye');
+			});
+		},
+
+		'Uses `properties.locale` when available'() {
+			localized = createLocalized({
+				properties: {
+					locale: 'fr'
+				}
+			});
+			return i18n(bundle, 'fr').then(() => {
+				const messages = localized.localizeBundle(bundle);
+				assert.strictEqual(messages.hello, 'Bonjour');
+				assert.strictEqual(messages.goodbye, 'Au revoir');
+			});
+		},
+
+		'Uses default locale when no locale is set'() {
+			return switchLocale('fr').then(() => {
+				localized = createLocalized();
+				return i18n(bundle, 'fr').then(() => {
+					const messages = localized.localizeBundle(bundle);
+					assert.strictEqual(messages.hello, 'Bonjour');
+					assert.strictEqual(messages.goodbye, 'Au revoir');
+				});
+			});
+		},
+
+		'Returns an object with a `format` method'() {
+			localized = createLocalized();
+			let messages = localized.localizeBundle(bundle);
+
+			assert.isFunction(messages.format);
+			assert.strictEqual(messages.format('welcome', { name: 'Bill' }), 'Welcome, Bill!');
+
+			return switchLocale('fr').then(() => {
+				return i18n(bundle, 'fr');
+			}).then(() => {
+				messages = localized.localizeBundle(bundle);
+				assert.strictEqual(messages.format('welcome', { name: 'Jean' }), 'Bienvenue, Jean!');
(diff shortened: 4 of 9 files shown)


src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,362 @@
...
+	'data-locale': string | null;
+	dir: string | null;
+}
+
+export type I18nMixin<M extends Messages, S extends I18nState> = I18n<M> & Stateful<S>;
+
+/**
+ * A map of bundles registered to each instance.
+ */
+const bundleMap = new WeakMap<I18nMixin<Messages, I18nState>, BundleMap<Messages>>();
```
kitsonk: Ideally, since the key is the same for all these WeakMaps, it would be better to create an interface for all of them and store all this private data in the same weak map.

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,362 @@
...
+		if (labels) {
+			const bundleLabels = Object.keys(labels).reduce((bundleLabels, key) => {
+				const label = labels[key];
+
+				if (label.indexOf(`${bundle.bundlePath}:`) === 0) {
+					bundleLabels[key] = label;
+				}
+
+				return bundleLabels;
+			}, {} as I18nLabels);
```
kitsonk: Possibly use of `Object.create(null)` here?

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,362 @@
...
+	registerBundle(this: I18nMixin<Messages, I18nState>, bundle: Bundle<Messages>) {
+		const { bundlePath } = bundle;
+		const registered = bundleMap.get(this);
+
+		if (!registered[bundlePath]) {
+			registered[bundlePath] = bundle;
+		}
+	}
+}, (instance: I18nMixin<Messages, I18nState>, options?: I18nOptions<I18nState>) => {
+	bundleMap.set(instance, {} as BundleMap<Messages>);
```
kitsonk: Do you need the coercion here?

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,362 @@
...
+	const bundles = options && options.bundles || instance.bundles;
```
kitsonk: There is no merging of instance and options bundles?  Instance bundles will only be used if there are no options bundles?

src/mixins/createRenderMixin.ts
```diff
@@ -149,6 +151,7 @@ function generateID(cachedRender: RenderMixin<RenderMixinState>): string {
 const widgetClassesMap = new WeakMap<RenderMixin<RenderMixinState>, string[]>();
 
 const createRenderMixin = createStateful
+	.mixin(createI18nMixin)
```
kitsonk: I know we sort of talked about this, but this then assumes that EVERY widget will be an i18n widget.  Do we want that hard dependency, since it does feel like `RenderMixin` doesn't have any real dependency on I18N from what I can see.

mwistrand:
`dojo-i18n` has not been published to npm yet, so it currently requires a symlink.

src/mixins/createI18nMixin.ts
mwistrand: My thought here was that bundles would be either registered via instantiation or within `mixin`/`extend`, but not both. If you think it would be preferable to merge them, I don't have a problem with that.

src/mixins/createRenderMixin.ts
mwistrand: That's a tough call. On the one hand, mixing it directly into `createRenderMixin` automatically benefits other out-of-the-box widgets like `createButton`, so that users don't have to store a separate factory somewhere just to use i18n with those widgets. On the other hand, no, not every widget will need to be localized.

At the very least, I'd argue that the out-of-the-box widgets should include `createI18nMixin`. Whether they inherit that through `createRenderMixin` or mix it in themselves depends on whether the additional overhead in `createRenderMixin` does or does not outweigh the inconvenience of users needing to mix `createI18nMixin` explicitly into their own widgets.

src/mixins/createI18nMixin.ts
kitsonk: I think we should merge them, because you could see a situation where, in a complex widget, where you would have some strings that are coming from the class and some that you might want to configure on the instance, but still have the ones that were baked into the class.

kitsonk: Also, thinking if a mixin has code that is dependent upon a bundle, and then some developer wants to add their own labels, the accidentally break that mixin.  Obviously they could "overwrite" the bundle by just using the same key.

mwistrand:
The functional tests that fail here also fail on master (when tested locally).

mwistrand:
Also, dojo/i18n#19 has been resolved, so this issue PR has no outside blockers.

tests/unit/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,353 @@
...
+		},
+
+		'Uses the parent\'s `state.locale` when available'() {
+			localized = createLocalized();
+			(<any> localized).parent = <any> {
+				state: { locale: 'fr' }
+			};
+			return i18n(bundle, 'fr').then(function () {
+				const messages = localized.localizeBundle(bundle);
+				assert.strictEqual(messages['hello'], 'Bonjour');
```
agubler: Can this not be `messages.hello`?

tests/unit/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,353 @@
...
+			const dfd = this.async(300);
+			switchLocale('fr');
+			localized = createLocalized();
+			let stateUpdated = false;
+			localized.own(localized.on('statechange', () => {
+				stateUpdated = true;
+			}));
+			localized.localizeBundle(bundle);
+
+			setTimeout(dfd.callback(() => {
```
agubler: let's use `dojo-core/timing#delays` and return the promise over `this.async(...`

tests/unit/mixins/createI18nMixin.ts
mwistrand: It can be; I didn't realize TS had been updated to allow dot notation on `[key: string]` interfaces (previously the threw warnings with dot notation).

agubler:
General question, I'm not sure that we should support the ability to pass bundles via options, to me everything i18n should be statically declared within a widget and I cannot think of a reasonable use case for being able to pass in bundles and specify keys via state (get's really messy if they need to be update/configured/modified by actions too)... if you get to that point the author would probably be better to extend a widget and import their nls in the custom widget?

Also do we think we need to support per widget locales? what is the use case to have multiple widgets with differing locales on a page? I do think that there is a use case for support changing the main locale however.

Lastly when the locale changes to how does the new nls bundle get loaded and all the widgets invalidated? (assume this happens internally in the i18n module?)

Sorry for all the questions!

dylans:
Main use case is if someone wanted to build the UI for something like Google translate, or to have an intentionally bilingual UI (common scenario is apps in Canada where the country strongly encourages two main languages, and someone needs to quickly support both languages).

The main use case we've used it for is training purposes, e.g. to show how to use i18n, it helps to be able to show two languages with two instances of the same widget on a page.

agubler:
>Main use case is if someone wanted to build the UI for something like Google translate, or to have an intentionally bilingual UI (common scenario is apps in Canada where the country strongly encourages two main languages, and someone needs to quickly support both languages).

Is per widget locale required to support this? Wouldn't this being able to change the main/app locale be sufficient for something like a UI for google translate or a bilingual UI?

mwistrand:
>General question, I'm not sure that we should support the ability to pass bundles via options, to me everything i18n should be statically declared within a widget and I cannot think of a reasonable use case for being able to pass in bundles and specify keys via state (get's really messy if they need to be update/configured/modified by actions too)... if you get to that point the author would probably be better to extend a widget and import their nls in the custom widget?

If there is a generic widget that will be reused in a number of contexts (e.g., `createSelect` that renders a list of options), then forcing users to write something like:

```typescript
const createCountrySelect = createSelect
    .mixin({
        mixin: { bundles: [ countriesBundle ] }
    });
const countrySelect = createCountrySelect();
```

Seems more circuitous than it needs to be, with something like the following being preferable:

```typescript
const countrySelect = createSelect({ bundles: [ countriesBundle ] });
```

>Also do we think we need to support per widget locales? what is the use case to have multiple widgets with differing locales on a page? I do think that there is a use case for support changing the main locale however.

Most applications will not need multiple locales side-by-side, but there are use cases. For example, a language-learning application would have instructions in one locale, but language materials in a different locale. Unless the additional flexibility severely aggravates performance, is there a reason for removing it?

>Lastly when the locale changes to how does the new nls bundle get loaded and all the widgets invalidated? (assume this happens internally in the i18n module?)

Right now, `dojo-i18n/i18n` registers `Stateful` objects for update when the locale changes. However, `dojo-i18n` should arguably know nothing about `Stateful` objects, and as such once dojo/shim#59 lands, `dojo-i18n` should (imo) be updated to rely on observables.

agubler:
>If there is a generic widget that will be reused in a number of contexts (e.g., createSelect that renders a list of options), then forcing users to write something like:
>
>const createCountrySelect = createSelect
    .mixin({
        mixin: { bundles: [ countriesBundle ] }
    });
const countrySelect = createCountrySelect();
Seems more circuitous than it needs to be, with something like the following being preferable:
>
>const countrySelect = createSelect({ bundles: [ countriesBundle ] });

How would `createSelect` ever know what to do with that bundle or what the keys were without intimate knowledge of the bundle itself (i.e. which would basically need to be done author time). I don't think `createSelect` would be a generic widget like suggested unless it offered something specific outside of dealing with the `nls` bundles.

>Most applications will not need multiple locales side-by-side, but there are use cases. For example, a language-learning application would have instructions in one locale, but language materials in a different locale. Unless the additional flexibility severely aggravates performance, is there a reason for removing it?

I guess I was just worried about overcomplicating/bloating i18n before we'd got the basics in place for how to integrate into a solution with widgets before extended it to support per widget locale switching.

>Right now, dojo-i18n/i18n registers Stateful objects for update when the locale changes. However, dojo-i18n should arguably know nothing about Stateful objects, and as such once dojo/shim#59 lands, dojo-i18n should (imo) be updated to rely on observables.

Does it update the state for all the widgets that are `i18n` registered? How many renders would this cause in a complex page of internationalized widgets?

mwistrand:
>How would createSelect ever know what to do with that bundle or what the keys were without intimate knowledge of the bundle itself

`createI18nMixin` uses `state.labels` to map message bundle keys to state properties (and individual widget factories could piggy back off this as needed). Something like `createTextInput` might be a better example than `createSelect`:

```typescript
const input = createTextInput({
    bundles: [ routeSpecificBundle ],
    state: {
        locale: 'fr',
        labels: {
            placeholder: 'routeSpecificPlaceholder'
        }
    }
});
```

>Does it update the state for all the widgets that are i18n registered? How many renders would this cause in a complex page of internationalized widgets?

If the root locale changes, registered widgets are only invalidated if they don't have their own locale explicitly set. With `dojo-shim/Observable`, `createI18nMixin` would subscribe the widget, and then invalidate itself if it does not have its own locale, or if its own `state.locale` is changed. The locale switch does cascade to children, so there potentially could be a lot of renders especially for complex applications with only a single locale at the root. Then again, dynamic locale switching will happen rarely in the overwhelming majority of applications, and the [build](https://github.com/dojo/cli-build) will ensure that supported locale data are bundled efficiently.

agubler:
>createI18nMixin uses state.labels to map message bundle keys to state properties (and individual widget factories could piggy back off this as needed). Something like createTextInput might be a better example than createSelect:
>```ts
>const input = createTextInput({
>    bundles: [ routeSpecificBundle ],
>    state: {
>        locale: 'fr',
>        labels: {
>            placeholder: 'routeSpecificPlaceholder'
>        }
>    }
>});

For locale's I would envisage that I would always access them in this fashion:

```ts
	const messages = this.localizeBundle(textInput);
	return { innerHTML: messages.placeholder };
```

So if `bundles` were accepted as an option I would expect the replacement to have the same `key` as the default `bundle` within `createTextInput` and not reference a state attribute `placeholder` (which at the moment would get set by `i18n` processing the `labels` section of the `state` as I understand it).

Perhaps it would be good to get a demo to show examples of the usage in real widgets set up to go other this as it is pretty hard to completely understand from this alone? What do you think?

mwistrand:
>Perhaps it would be good to get a demo to show examples of the usage in real widgets set up to go other this as it is pretty hard to completely understand from this alone? What do you think?

It's outdated, and will be updated again with this implementation, but you can see how this is incorporated into todo-mvc [here](https://github.com/mwistrand/examples/tree/62-incorporate-i18n). As an example, `createTodoFooter` uses the `createButton` factory provided out-of-the-box by `dojo-widgets` to render the "Clear completed" button. Without i18n, the "Clear completed" text is put directly on `state.label`. With `createI18nMixin`, the same generic widget is used, but `state.labels` is used, and `state.label` is updated with the internationalize "Clear completed" text once the associated bundle has loaded.

`localizeBundle` could be used to the same effect, but then `createButton` would have to be extended just for this particular button:

```typescript
const createClearCompleted = createButton
    .mixin({
        mixin: {
            bundles: [ bundle ],
            nodeAttributes: [
                function (attributes) {
                    const messages = this.localizeBundle(bundle);
                    return { label: messages.clearCompleted };
                }
            }
        }
    });
```

I'm having a hard time seeing how that is better than the following, which seems to me to be simpler:

```typescript
const clearCompletedButton = createButton({
    bundles: [ bundle ],
    state: {
        labels: { label: 'clearCompleted' }
    }
});
```

agubler:
For me in the example of the "clear completed" button in todo-mvc that actual i18n bundle wouldn't be dealt with at the button level. It would be dealt with at the enclosing widgets level, in this case the `createTodoFooter` and then injected in as state, something like:

```ts
import { Widget, WidgetOptions, WidgetState, DNode } from 'dojo-widgets/interfaces';
import createWidgetBase from 'dojo-widgets/createWidgetBase';
import d from 'dojo-widgets/d';
import createButton from 'dojo-widgets/components/button/createButton';
import { clearCompleted } from '../actions/userActions';
import createTodoFilter from './createTodoFilter';
import todoFooterBundle from ./nls/todoFooterBundle';

export type TodoFooterState = WidgetState & {
	activeFilter?: string;
	activeCount?: number;
	completedCount?: number;
};

export type TodoFooterOptions = WidgetOptions<TodoFooterState>;

export type TodoFooter = Widget<TodoFooterState>;

const createTodoFooter = createWidgetBase
.mixin(createI18nMixin)
.mixin({
	mixin: {
		tagName: 'footer',
		classes: [ 'footer' ],
                bundle: [ todoFooterBundle ],
		getChildrenNodes: function(this: TodoFooter): (DNode | null)[] {
			const { activeCount, activeFilter, completedCount } = this.state;
                        const messages = this.localizeBundle(todoFooterBundle);
			const countLabel = activeCount === 1 ? messages.itemLeft : messages.itemsLeft;

			return [
				d('span', { 'class': 'todo-count' }, [
					d('strong', [ activeCount + ' ' ]),
					d('span', [ countLabel ])
				]),
				d(createTodoFilter, {
					state: {
						classes: [ 'filters' ],
						activeFilter
					}
				}),
				completedCount ? d(createButton, {
					listeners: {
						click: clearCompleted
					},
					state: {
						label: messages.clearCompleted,
						classes: [ 'clear-completed' ]
					}
				}) : null
			];
		}
	}
});

export default createTodoFooter;
```

It seems reasonable to assume that generic widgets such as `createButton`, `createTextNode` etc etc will actually be managed by an enclosing stateful widget that can specify the bundle required for the application.

agubler:
The same is true if we wanted to implement the `createSelect` example mentioned earlier, a widget cannot be standalone it has to be at least wrapped in a projector. For example with the example that simply displays a list of countries it could look something like this:

*createSelect*
```ts
import { ComposeFactory } from 'dojo-compose/compose';
import d from 'dojo-widgets/d';
import createWidgetBase from 'dojo-widgets/createWidgetBase';
import { DNode, Widget, WidgetState, WidgetOptions } from 'dojo-widgets/interfaces';

type SelectWidgetState = WidgetState & {
     items: string[];
};

type SelectWidget = Widget<SelectWidgetState>;

type SelectWidgetFactory = ComposeFactory<SelectWidget, WidgetOptions<SelectWidgetState>>;

const createSelect: SelectWidgetFactory = createWidgetBase.mixin({
    mixin: {
        tagName: 'ul',
        getChildrenNodes(this: SelectWidget): DNode[] {
            const { state } = this;
            
            return state.items.map((item) =>
                return d('li', [ item ]);
            });
        }
    }
});

export default createSelect;
```

*Main App Widget*

```ts
import { DNode } from 'dojo-widgets/interfaces';
import createProjector, { Projector } from 'dojo-widgets/createProjector';
import createI18nMixin from 'dojo-widgets/mixins/createI18nMixin';
import d from 'dojo-widgets/d';

import createSelect, { SelectWidget } from './createSelect';
import bundle from 'nls/appBundle';

const createApp = createProjector
.mixin(createI18nMixin)
.mixin({
    mixin: {
        bundle: [ bundle ],
        getNode(this: Projector): DNode {
            const messages = this.localizeBundle(bundle);

            const items = Object.keys(messages).map((key) => {
                return messages[key];
            });

            return d(createSelect, { state: { items } });
        }         
    }
});

export default createApp;
```

With the final usage being something like:

```ts
import createApp from './createApp';

const app = createApp();

app.append().then(() => {
    console.log('application attached');
});
```

Apologies if there are any typos or syntax errors I wrote this in the comment, I hope it kind of gives an idea of where I am coming from 😄

mwistrand:
Thanks, @agubler. That does clarify your position. This does have the benefit of separating the simplest components like `createButton` from needing to mix in `createI18nMixin`. I will simplify the mixin accordingly and push that as soon as possible.

agubler:
@mwistrand this is failing in CI for a lint error.

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,161 @@
...
+ * The widget's locale.
+ */
+function getParentLocale(instance: I18nWidget<Messages, I18nState>): string {
+	const { locale } = instance.state;
+
+	if (locale) {
+		return locale;
+	}
+
+	const parent = (<any> instance).parent;
```
agubler: we have no reference to the `parent` any more in widgets on the instance.

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @private
+ * An internal helper interface for defining locale and text direction attributes on widget nodes.
+ */
+interface I18nVNodeProperties extends VNodeProperties {
+	'data-locale': string | null;
+	dir: string | null;
+}
+
+export type I18nWidget<M extends Messages, S extends I18nState> = I18nMixin<M> & Widget<S>;
```
mwistrand: Since `Widget` only requires a single generic, and since `Messages` is a simple `[key: string]: string` interface, does is it make sense to specify `Messages` as another generic for `I18nWidget`?

README.md
```diff
@@ -343,6 +344,37 @@ app.append().then(() => {
 });
 ```
 
+### Internationalization
+
+Widgets can be internationalized by mixing in `dojo-widgets/mixins/createI18nMixin`. [Message bundles](https://github.com/dojo/i18n) are added by specifying a `bundles` array to the widget factory prototype. When manually accessing localized messages, the bundle should first be passed to `localizeBundle`. If the bundle supports the widget's current locale, but those locale-specific messages have not yet been loaded, then the default messages are returned and the widget will be invalidated once the locale-specific messages have been loaded. Each widget can have its own locale by setting its `state.locale`; if no locale is set, then the default locale as set by [`dojo-i18n`](https://github.com/dojo/i18n) is assumed.
+
+```typescript
+const createI18nWidget = createWidgetBase
```
agubler: Can we add an example where we use this in `getChildrenNodes` and pass the `i18n` values to a child widget via the state?

something like 
```ts
getChildrenNodes: function() {
    const messages = this.localizedBundle(greetings);

    return [
        d(createButton, { state: { label: messages.buttonLabel } } );
    ];
}
```

dylans:
@mwistrand This has a conflict because of the change from RxJS to ES8 Observables.

package.json
```diff
@@ -23,10 +23,13 @@
     "test": "grunt test"
   },
   "peerDependencies": {
+    "cldr-data": "~30.0.0",
     "dojo-compose": ">=2.0.0-beta.10",
     "dojo-core": ">=2.0.0-alpha.14",
     "dojo-has": ">=2.0.0-alpha.4",
+    "dojo-i18n": ">=2.0.0-alpha.3",
     "dojo-shim": ">=2.0.0-beta.1",
+    "globalize": "^1.1.1",
```
agubler: why is this a peer dependency and not a dependency of i18n?

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param bundle
+ * The bundle for which to load a locale-specific dictionary.
+ *
+ * @return
+ * The locale-specific dictionary, if it has already been loaded and cached.
+ */
+function getLocaleMessages(instance: I18nWidget<Messages, I18nProperties>, bundle: Bundle<Messages>): Messages | void {
+	const { properties } = instance;
+	const locale = properties['locale'] || i18n.locale;
```
agubler: `properties.locale`?

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,150 @@
...
+
+	i18n(bundle, locale).then(() => {
+		instance.invalidate();
+	});
+}
+
+const createI18nMixin: I18nFactory = compose<I18nMixin<Messages>, WidgetOptions<WidgetState, I18nProperties>>({
+	nodeAttributes: [
+		function (this: I18nWidget<Messages, I18nProperties>, attributes: VNodeProperties): VNodeProperties {
+			const properties = {
```
agubler: lots of `properties` going on in here, can we call the variable for `VNodeProperties` `vNodeProperties` or similar?

src/mixins/createI18nMixin.ts
```diff
@@ -0,0 +1,150 @@
...
+		return assign(Object.create({
+			format(key: string, options?: any) {
+				return formatMessage(bundle.bundlePath, key, options, locale);
+			}
+		}), messages);
+	}
+}, (instance: I18nWidget<Messages, I18nProperties>) => {
+	const subscription = observeLocale({
+		next() {
+			if (!instance.properties['locale']) {
```
agubler: `instance.properties.locale`

PR #89, opened 2016-11-04
Implement createWidgetBase and `d` proposal

### LINKED ISSUE #87: Widget Base

### LINKED ISSUE #88: Add `d` capabilities for WidgetBase

agubler:
**Type:** feature

**Description:**

The first implementation of `createWidgetBase` using the new base interfaces and introduction of the `d` proposal implementation.

Resolves #87
Resolves #88


src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,185 @@
...
+	currentChildrenMap: Map<string | Factory<Widget<WidgetState>, WidgetOptions<WidgetState>>, Widget<WidgetState>>;
+};
+
+const widgetInternalStateMap = new WeakMap<Widget<WidgetState>, WidgetInternalState>();
+
+function isWNode(child: DNode): child is WNode {
+	return child && (<WNode> child).factory !== undefined;
+}
+
+function realizeDNode(instance: Widget<WidgetState>, dNode: DNode) {
```
agubler: missing return type, also happy to bike shed on the function name.

agubler: Also @matt-gadd suggested splitting the function which we can certainly do.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,185 @@
...
+				internalState.dirty = true;
+				this.emit({
+					type: 'invalidate'
+				});
+			},
+
+			get id(this: Widget<WidgetState>): string {
+				const { id } = widgetInternalStateMap.get(this);
+
+				return id || (this.state && this.state.id) || '';
```
agubler: I think we need to utilise the auto generate id, similar to `createRenderMixin`

src/mixins/createParentListMixin.ts
```diff
@@ -50,19 +50,24 @@ export interface ParentListMixinFactory extends ComposeFactory<ParentListMixin<C
...
 		},
 
-		set children(this: ParentListMixin<Child>, value: List<Child>) {
+		set children(this: ParentListMixin<Child>, value: List<Child & Evented>) {
 			if (!value.equals(childrenMap.get(this))) {
 				value.forEach((widget) => {
 					// Workaround for https://github.com/facebook/immutable-js/pull/919
 					// istanbul ignore else
 					if (widget) {
+						widget.on('invalidate', () => {
```
agubler: Need this to ensure that the parent (specifically the projector) knew about the children invalidation.

src/mixins/interfaces.ts
```diff
@@ -11,7 +11,7 @@ export interface Child extends Renderable, Destroyable {
 	/**
 	 * A reference to the widget's parent
 	 */
-	parent: Parent | null;
+	parent?: Parent | null;
```
agubler: It is no longer mandatory as it has been removed from the new `Widget` interface

src/util/d.ts
```diff
@@ -0,0 +1,41 @@
...
+function d(tagName: string, options?: VNodeProperties, children?: (DNode | VNode)[]): HNode;
+function d(factory: ComposeFactory<Widget<WidgetState>, WidgetOptions<WidgetState>>, options: WidgetOptions<WidgetState>): WNode;
+function d(tagNameOrFactory: TagNameOrFactory, options: DOptions = {}, children?: (DNode | VNode)[]): DNode {
+	children = children ? children : [];
+
+	if (typeof tagNameOrFactory === 'string') {
+		return {
+			children: children,
+			render(this: { children: VNode[] }) {
+				return h(<string> tagNameOrFactory, <VNodeProperties> options, this.children);
```
agubler: I still had to cast these arguments, is there a way around this?

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,186 @@
...
+	const internalState = widgetInternalStateMap.get(instance);
+	let child: HNode | Widget<WidgetState>;
+	if (isWNode(dNode)) {
+		const { factory, options: { id, state } } = dNode;
+		const cachedChild = internalState.historicChildrenMap.get(id || factory);
+		if (cachedChild) {
+			child = cachedChild;
+			if (state) {
+				child.setState(state);
+				child.on('invalidate', () => {
```
matt-gadd: why do you have to attach this event every render?

agubler: noted!

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,186 @@
...
+
+			getChildrenNodes(this: Widget<WidgetState>): DNode[] {
+				let childrenWrappers: DNode[] = [];
+
+				this.childNodeRenderers.forEach((fn) => {
+					const wrappers = fn.call(this);
+					childrenWrappers = childrenWrappers.concat(wrappers);
+				});
+
+				return childrenWrappers.filter((child) => child);
```
matt-gadd: this won't filter anything other than immediate children.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,186 @@
...
+			child.own(child.on('invalidate', () => {
+				instance.invalidate();
+			}));
+			internalState.historicChildrenMap.set(id || factory, child);
+			instance.own(child);
+		}
+		if (!id && internalState.currentChildrenMap.has(factory)) {
+			console.error('must provide unique keys when using the same widget factory multiple times');
+		}
+		internalState.currentChildrenMap.set(id || factory, child);
```
matt-gadd: hoist this logical `or` to a `key` const? its used 3 times.

src/bases/createWidgetBase.ts
agubler: good point!

src/bases/createWidgetBase.ts
kitsonk: I think the return annotation would make it clearer.  It is required for exported functions, but unless it is bloody obvious what it does then even internal functions should be explicit (it also help safeguard the function returns from logic errors, especially in strict mode).  It does feel a bit complex to maintain, so yes, 👍 on splitting it.  I am not a fan of the term `realize` (even though we use it in `app`)... this actually feels like to me 🚲 🏠  `resolveVNode()` or `dNodeToVNode()` or `getVNode()`.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+import Map from 'dojo-shim/Map';
+import d from './../util/d';
+
+export interface WidgetFactory extends ComposeFactory<Widget<WidgetState>, WidgetOptions<WidgetState>> {}
+
+interface WidgetInternalState {
+	readonly id?: string;
+	dirty: boolean;
+	widgetClasses: string[];
+	cachedWidget?: VNode;
```
kitsonk: `cachedVNode`?

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+	if (isWNode(dNode)) {
+		const { factory, options: { id, state } } = dNode;
+		const childrenMapKey = id || factory;
+		const cachedChild = internalState.historicChildrenMap.get(childrenMapKey);
+		if (cachedChild) {
+			child = cachedChild;
+			if (state) {
+				child.setState(state);
+			}
+		} else {
```
kitsonk:
```
}
else {
```

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+			child = factory(dNode.options);
+			child.own(child.on('invalidate', () => {
```
kitsonk: does the `child` really *own* this?  Isn't it `instance`, so that if `instance` is destroyed, the listener is unregistered?

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+				instance.invalidate();
+			}));
+			internalState.historicChildrenMap.set(childrenMapKey, child);
+			instance.own(child);
+		}
+		if (!id && internalState.currentChildrenMap.has(factory)) {
+			console.error('must provide unique keys when using the same widget factory multiple times');
```
kitsonk: should be emitted as `instance.emit({ type: 'error', target: instance, error })`

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+		}
+		internalState.currentChildrenMap.set(childrenMapKey, child);
+	}
+	else {
+		child = dNode;
+		if (child.children) {
+			child.children = child.children.map((child: DNode) => {
+				return realizeDNode(instance, child);
```
kitsonk: could use implicit return here... `child.children.map((child) => realizeDNode(instance, child));`

Can children ever be not a `DNode`?  If this were broken out a separate function, it is likely you can tighten the types as well so that `child` could be inferred.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+			});
+		}
+	}
+	return child.render();
+}
+
+function manageDetachedChildren(instance: Widget<WidgetState>) {
```
kitsonk: This is the one thing that *scares* me, about how in practice we end up thrashing instances...  Something to potentially keep an eye on to ensure we aren't needlessly creating and destroying widgets.  Also expressing a return type of `void` would make this function more clear.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+
+				this.nodeAttributes.forEach((fn) => {
+					const newProps: VNodeProperties = fn.call(this, assign({}, props));
+					if (newProps) {
+						assign(props, newProps);
+					}
+				});
+
+				if (overrides) {
+					assign(props, overrides);
```
kitsonk: Uncovered... but I am starting to question of this, with the way we have the children nodes now.  Maybe we should drop this?

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+					childrenWrappers = childrenWrappers.concat(wrappers);
+				});
+
+				return childrenWrappers;
+			},
+
+			getNodeAttributes(this: Widget<WidgetState>, overrides?: VNodeProperties): VNodeProperties {
+				const props: VNodeProperties = {};
+
+				this.nodeAttributes.forEach((fn) => {
```
kitsonk: Also, I know this was ported over, but maybe I was being paranoid about how the node attributes functions worked... is passing in the current set in a new object and then mixin the return in an unecessary caution.

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+					const newProps: VNodeProperties = fn.call(this, assign({}, props));
```
kitsonk: I can't remember if I tried this, but potentially a good site for `Object.create(null)` instead of `{}`

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+				return props;
+			},
+
+			invalidate(this: Widget<WidgetState>): void {
+				const internalState = widgetInternalStateMap.get(this);
+				if (internalState.dirty) {
+					return;
+				}
+				internalState.dirty = true;
+				this.emit({
```
kitsonk: As per the issue with Android, would it be safer to always emit `invalidate` and shouldn't it include a property of `target`?

src/bases/createWidgetBase.ts
```diff
@@ -0,0 +1,201 @@
...
+				return internalState.cachedWidget;
+			},
+
+			tagName: 'div'
+		},
+		initialize(instance: Widget<WidgetState>, options: WidgetOptions<WidgetState> = {}) {
+			const { id, tagName } = options;
+			instance.tagName = tagName || instance.tagName;
+
+			widgetInternalStateMap.set(instance, {
```
kitsonk: In theory `widgetInternalStateMap.get(instance).id` and `instance.state.id` can be out of sync, given the following `createWidget({ id: 'foo', state: { id: 'bar' } });`  State changes to `state.id` also don't adjust the `instance.id`.  Don't know if that is desirable.  I wonder if the getter logic should only return the private ID if there is nothing in state?

src/util/d.ts
```diff
@@ -0,0 +1,43 @@
...
+import { h } from 'maquette';
+
+export type TagNameOrFactory = string | ComposeFactory<Widget<WidgetState>, WidgetOptions<WidgetState>>;
+
+export type DOptions = VNodeProperties | WidgetOptions<WidgetState>;
+
+function d(tagName: string, options?: VNodeProperties, children?: (DNode | VNode)[]): HNode;
+function d(factory: ComposeFactory<Widget<WidgetState>, WidgetOptions<WidgetState>>, options: WidgetOptions<WidgetState>): WNode;
+function d(tagNameOrFactory: TagNameOrFactory, options: DOptions = {}, children?: (DNode | VNode)[]): DNode {
+	children = children ? children : [];
```
kitsonk: `children = children || []`? or even better, a default value:

```ts
function d(tagNameOrFactory: TagNameOrFactory, options: DOptions = {}, children: (DNode | VNode)[] = [])
```

src/util/d.ts
```diff
@@ -0,0 +1,43 @@
...
+		children = children.filter((child) => child);
+
+		return {
+			children: children,
+			render(this: { children: VNode[] }) {
+				return h(<string> tagNameOrFactory, <VNodeProperties> options, this.children);
+			}
+		};
+	}
+	else if (typeof tagNameOrFactory === 'function') {
```
kitsonk: because of the return above, the `else` is unnecessary.

src/util/d.ts
```diff
@@ -0,0 +1,43 @@
...
+		return {
+			factory: tagNameOrFactory,
+			options: <WidgetOptions<WidgetState>> options
+		};
+	}
+	else {
```
kitsonk: again, because of the return, the `else` block is unnecessary.

src/util/d.ts
```diff
@@ -0,0 +1,43 @@
...
+function d(tagNameOrFactory: TagNameOrFactory, options: DOptions = {}, children?: (DNode | VNode)[]): DNode {
+	children = children ? children : [];
+
+	if (typeof tagNameOrFactory === 'string') {
+		children = children.filter((child) => child);
+
+		return {
+			children: children,
+			render(this: { children: VNode[] }) {
+				return h(<string> tagNameOrFactory, <VNodeProperties> options, this.children);
```
kitsonk: We should note here that in 2.1 the cast of `tagNameOrFactory` will be unnecessary, because arguments that aren't touched within a function scope will be treated as `const` and therefore the narrowing above will cross the function barrier.

src/bases/createWidgetBase.ts
agubler: At this point, no the child should not be anything other than a `DNode`

src/bases/createWidgetBase.ts
agubler: :+1: on `void`

As for thrashing, I think if you've removed this from the `children` return completely then it is justified to call `destroy` over leaving the instance incase it is used again. If a consumer didn't want to have the instance destroyed I'd expect the `child` to still return the `child` but toggle visibility using css.

I agree, that regardless we should monitor this to make sure that this doesn't give any cause for concern.

src/bases/createWidgetBase.ts
agubler: you'd want the `child` to own it also no? If the child was destroyed then you wouldn't want the listener hanging around? If the `instance` is destroyed then the child will implicitly be destroyed and therefore the listener also.

src/bases/createWidgetBase.ts
agubler: I've actually changed this locally already to not pass a copy of the props.

src/bases/createWidgetBase.ts
agubler: At which point I should just remove the short circuit as it doesn't add anything

src/bases/createWidgetBase.ts
agubler: Do you think we should allow construction of a widget with conflicting `id`s?

src/util/d.ts
agubler: I thought that there was something coming to assist in that... Good to know.

src/bases/createWidgetBase.ts
kitsonk: The question then becomes in what "domain"?  They would only conflict when integrated with something that cares about them being unique (like `app`).  So my feeling is at this level we shouldn't have an opinion about it, but if in something like `app` it would throw/reject situations where duplicates (which `app` already does).

src/bases/createWidgetBase.ts
kitsonk: Does it still pass props at all?  If it passes the current set, would we want to drop the return? essentially "modify these, or don't do anything".

agubler: It just returns the attributes from the mixin and then mixes them over the existing props.

src/bases/createWidgetBase.ts
```diff
@@ -68,22 +69,22 @@ function realizeDNode(instance: Widget<WidgetState>, dNode: DNode) {
 			instance.own(child);
 		}
 		if (!id && internalState.currentChildrenMap.has(factory)) {
-			console.error('must provide unique keys when using the same widget factory multiple times');
+			const errorMsg = 'must provide unique keys when using the same widget factory multiple times';
+			console.error(errorMsg);
```
kitsonk: Think about this again, it would be good to create an `Error` here, but not throw it, so downstream they get the stack trace etc and return that in the emit as well as what gets logged.  I think we also need to think about how this fits into `dojo-debug` (cc: @rishson) and if we actually want to log errors to the console or if we will go with some other sort of instrumentation that can be easier to manage.

src/bases/createWidgetBase.ts
```diff
@@ -117,26 +118,21 @@ const createWidget: WidgetFactory = createStateful
 				const props: VNodeProperties = {};
 
 				this.nodeAttributes.forEach((fn) => {
-					const newProps: VNodeProperties = fn.call(this, assign({}, props));
+					const newProps: VNodeProperties = fn.call(this);
```
kitsonk: Ah, my other comment was the other direction... I guess would we see a situation where the current set of properties would effect the return?  I could see something like that, though obviously it would create dependencies between the functions which means potentially surprising results.  This would mean though that the function would be going in "blind" and have no idea that it might be overriding a value.

src/bases/createWidgetBase.ts
```diff
@@ -117,26 +118,21 @@ const createWidget: WidgetFactory = createStateful
...
 
 			invalidate(this: Widget<WidgetState>): void {
 				const internalState = widgetInternalStateMap.get(this);
-				if (internalState.dirty) {
-					return;
-				}
 				internalState.dirty = true;
 				this.emit({
-					type: 'invalidate'
+					type: 'invalidate',
```
kitsonk: Again, thinking out loud... would there ever be a use case that we would want `invalidate` to be cancellable?  I guess if we find a use case for it we can add it.

src/util/d.ts
```diff
@@ -29,15 +28,15 @@ function d(tagNameOrFactory: TagNameOrFactory, options: DOptions = {}, children?
...
 		return {
 			factory: tagNameOrFactory,
 			options: <WidgetOptions<WidgetState>> options
 		};
 	}
-	else {
-		throw new Error('Unsupported tagName or factory type');
-	}
+
+	throw new Error('Unsupported tagName or factory type');
```
kitsonk: Again, thinking out loud... as we found out with Maquette, a throw during a render will totally halt any renders in forever at the moment.  Would an error in `d()` actually be terminal.  Also speaks to `dojo-debug` and how we instrument our code (cc/ @rishson)

src/bases/createWidgetBase.ts
agubler: Yeah agree, for now I think it's fine.

src/bases/createWidgetBase.ts
agubler: in both cases, I think it could cause surprising results because it would certainly be order dependent on mixins (which I don't like); It is ported over from the existing functionality so if we think that we need to address this some more, perhaps it would be part of a separate issue?

src/bases/createWidgetBase.ts
agubler: Roger that `new Error(errorMsg)`

PR #96, opened 2016-11-10
Convert existing widgets to widgetbase

### LINKED ISSUE #95: Convert existing widgets and mixins to widgetBase and `d`

agubler:
**Type:** feature

**Description:**

Convert existing widgets to the new `createWidgetBase` base class.

Resolves #95


agubler:
This should allow it so that the new version can just be dropped in to existing consumers

agubler:
@kitsonk can I get a review of this at some point - this should move all existing widgets to use `createWidgetBase` and other `dojo-interfaces` API.

I imagine a lot of these will require re-working when we have a container solution in place, so didn't focus on making them "better" but just to get parity for the existing widgets.

src/bases/createWidgetBase.ts
```diff
@@ -96,6 +96,13 @@ function manageDetachedChildren(instance: Widget<WidgetState>): void {
 	internalState.currentChildrenMap.clear();
 }
 
+function formatTagNameAndClasses(tagName: string, classes: string[]) {
```
agubler: Need to add unit tests for this.

src/mixins/createRenderMixin.ts
```diff
@@ -1,254 +1,9 @@
...
-
-			instance.own(instance.on('state:changed', () => instance.invalidate()));
-
-			widgetClassesMap.set(instance, []);
-		}
-	})
-	.static({
-		idBase: 'widget'
-	}) as RenderMixinFactory;
+export {
```
kitsonk: I assume the main reason we are doing this is to minimise the amount of refactoring?

src/mixins/createTabbedMixin.ts
```diff
@@ -248,18 +245,26 @@ const createTabbedMixin: TabbedMixinFactory = createRenderMixin
 					if (tabVNode.properties) {
 						(tabVNode as any).properties['data-visible'] = String(isActiveTab);
 					}
-					childrenNodes[idx] = tabVNode;
+					const childNode = {
```
kitsonk: Ideally we would be explicit what the object literal type is.  It should be `DNode`, right?

agubler: Yeah `DNode` or `HNode` I believe

PR #119, opened 2016-11-21
support 'children' as the second argument when creating HNodes

### LINKED ISSUE #114: In 'd' use second param as children if type of Array
**Enhancement** 

It is too easy to incorrectly pass the array of `children` as the 2nd param while using `d` where there are no `options` required. This is because the object type for `options` is `VNodeProperties` which is a completely optional interface and therefore TS allows `any`.

Proposal is to check if the param is an array and copy to `children`, this should reduce the number of bugs for consumers.

We should also actually type the function by `| Children[]` because otherwise if TS fix the ability to type check against completely optional interfaces it would affect consumers who have leveraged this enhancement (whether leveraged intentionally or not).

Package Version: latest

agubler:
**Type:** feature

**Description:**

Resolves #114 

Support children as the second param when using `d` to generate a ` HNode`

### DIFF
--- a/src/util/d.ts
+++ b/src/util/d.ts
@@ -14,14 +14,15 @@ export type TagNameOrFactory<S extends WidgetState, W extends Widget<S>, O exten
 
 export type DOptions<S extends WidgetState, O extends WidgetOptions<S>> = VNodeProperties | O;
 
-type Children = (DNode | VNode | null)[];
+export type Children = (DNode | VNode | null)[];
 
 function d(tagName: string, optionsOrChildren?: VNodeProperties | Children, children?: Children): HNode;
 function d<S extends WidgetState, W extends Widget<S>, O extends WidgetOptions<S>>(factory: ComposeFactory<W, O>, options: O): WNode;
 function d<S extends WidgetState, W extends Widget<S>, O extends WidgetOptions<S>>(
 	tagNameOrFactory: TagNameOrFactory<S, W, O>,
 	optionsOrChildren: DOptions<S, O> = {},
-	children: Children = []): DNode {
+	children: Children = []
+): DNode {
 
 	if (typeof tagNameOrFactory === 'string') {
 		if (Array.isArray(optionsOrChildren)) {
--- a/src/util/d.ts
+++ b/src/util/d.ts
@@ -16,7 +16,9 @@ export type DOptions<S extends WidgetState, O extends WidgetOptions<S>> = VNodeP
 
 export type Children = (DNode | VNode | null)[];
 
-function d(tagName: string, optionsOrChildren?: VNodeProperties | Children, children?: Children): HNode;
+function d(tagName: string, options: VNodeProperties, children: Children): HNode;
+function d(tagName: string, children: Children): HNode;
+function d(tagName: string): HNode;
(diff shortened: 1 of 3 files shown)


src/util/d.ts
```diff
@@ -16,28 +16,33 @@ export type DOptions<S extends WidgetState, O extends WidgetOptions<S>> = VNodeP
 
 type Children = (DNode | VNode | null)[];
 
-function d(tagName: string, options?: VNodeProperties, children?: (DNode | VNode | null)[]): HNode;
+function d(tagName: string, optionsOrChildren?: VNodeProperties | Children, children?: Children): HNode;
```
kitsonk: from a usability perspective, it might be nicer to have a better set of overloads here, it will give cleared intellisense:

```ts
function d(tagName: string, options: VNodeProperties, children?: Children): HNode;
function d(tagName: string, children: Children): HNode;
function d(tagName: string): HNode;
```

You may also need to export `Children` now so it can be "named" when creating the declaration for `d`.  I would check with a `grunt dist`.

src/util/d.ts
```diff
@@ -16,28 +16,33 @@ export type DOptions<S extends WidgetState, O extends WidgetOptions<S>> = VNodeP
...
 function d<S extends WidgetState, W extends Widget<S>, O extends WidgetOptions<S>>(factory: ComposeFactory<W, O>, options: O): WNode;
 function d<S extends WidgetState, W extends Widget<S>, O extends WidgetOptions<S>>(
 	tagNameOrFactory: TagNameOrFactory<S, W, O>,
-	options: DOptions<S, O> = {},
+	optionsOrChildren: DOptions<S, O> = {},
 	children: Children = []): DNode {
```
kitsonk: I should have caught this before, but the closing paran should be outdented so it is clear where the arguments end:

```ts
	children: Children = []
): DNode {
```

src/util/d.ts
```diff
@@ -16,28 +16,33 @@ export type DOptions<S extends WidgetState, O extends WidgetOptions<S>> = VNodeP
...
 
 	if (typeof tagNameOrFactory === 'string') {
+		if (Array.isArray(optionsOrChildren)) {
+			children = optionsOrChildren;
+			optionsOrChildren = {};
```
kitsonk: `Object.create(null)`?

src/util/d.ts
agubler: good catch, thanks.

src/util/d.ts
```diff
@@ -14,19 +14,20 @@ export type TagNameOrFactory<S extends WidgetState, W extends Widget<S>, O exten
...
 	optionsOrChildren: DOptions<S, O> = {},
-	children: Children = []): DNode {
+	children: Children = []
+): DNode {
 
 	if (typeof tagNameOrFactory === 'string') {
 		if (Array.isArray(optionsOrChildren)) {
 			children = optionsOrChildren;
-			optionsOrChildren = {};
+			optionsOrChildren = Object.create(null);
```
kitsonk: ooops... I forgot maquette Object.hasOwnPropertyDescriptor on those, which means you can't use `Object.create(null)`.

agubler: Ah I hadn't checked the CI results, should have run tests locally. doh.

agubler:
@kitsonk I didn't actually notice the overload suggestion, sorry! let me go back and have a look.

PR #146, opened 2016-12-01
Support children for widgets

### LINKED ISSUE #145: Support `widget` children
**Enhancement**

The current mechanism for supporting children is not symmetrical between creating a widget using a factory and creating `HNode`s using `d` as creating children for a widget is not currently supported.

Adding children to a fatory widget should be supported by passing an array of `DNode`s as the third argument of `d` that get set on a `children` property of the parent and used by default within the `getChildrenNode` function.

Package Version: *latest*

agubler:
**Type:** feature

**Description:**

Add support in `d` and `createWidgetBase` for passing `children` to when creating a widget using a factory.

Resolves #145

Depends on https://github.com/dojo/interfaces/pull/47

### DIFF
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -133,8 +133,7 @@ const createWidget: WidgetFactory = createStateful
 			},
 
 			getChildrenNodes(this: Widget<WidgetState>): DNode[] {
-				const internalState = widgetInternalStateMap.get(this);
-				return internalState.children;
+				return this.children;
 			},
 
 			getNodeAttributes(this: Widget<WidgetState>, overrides?: VNodeProperties): VNodeProperties {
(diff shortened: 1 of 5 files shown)


src/createWidgetBase.ts
```diff
@@ -120,8 +123,18 @@ const createWidget: WidgetFactory = createStateful
...
+			},
+
+			get children() {
+				return widgetInternalStateMap.get(this).children;
+			},
+
 			getChildrenNodes(this: Widget<WidgetState>): DNode[] {
-				return [];
+				const internalState = widgetInternalStateMap.get(this);
+				return internalState.children;
```
matt-gadd: think we can just use the getter here. `return this.children;`

PR #151, opened 2016-12-01
Add attach type for projector

### LINKED ISSUE #150: Allow `attachType` to be passed on `projector.attach`
**Enhancement**

Add `AttachOptions` that can be passed on `projector.attach({...});` containing `type` that will determine the type of projector attachment ([`merge`](http://maquettejs.org/docs/typedoc/interfaces/_maquette_.projector.html#merge)/[`append`](http://maquettejs.org/docs/typedoc/interfaces/_maquette_.projector.html#append)/[`replace`](http://maquettejs.org/docs/typedoc/interfaces/_maquette_.projector.html#replace))

Default attach type will be `append`

Package Version: latest

agubler:
**Type:** feature

**Description:**

Add attach type as an option for `projector.attach(...)`

Resolves #150

### DIFF
--- a/src/createProjector.ts
+++ b/src/createProjector.ts
@@ -16,7 +16,14 @@ export enum ProjectorState {
 	Detached
 };
 
-export type AttachType = 'append' | 'merge' | 'replace';
+/**
+ * Attach type for the projector
+ */
+export enum AttachType {
+	Append = 1,
+	Merge = 2,
+	Replace = 3
+};
 
 export interface AttachOptions {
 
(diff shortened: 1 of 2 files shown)


src/createProjector.ts
```diff
@@ -128,7 +141,19 @@ const createProjector: ProjectorFactory = createWidgetBase
 					};
 				});
 
-				projectorData.projector.append(projectorData.root, render);
+				switch (type) {
+					case 'append':
```
matt-gadd: do you think these should be symbols (something like `projector.attach({ type: projector.attachTypes.MERGE })`? or overkill?

matt-gadd: On second thought, it might be nice if we just make the types into functions, so rather than having `attach()`, we just have `append()`, `merge()` and `replace()`. That way we get intellisense and the end user can't use an incorrect attach type?

PR #181, opened 2016-12-13
Support properties to update internal state

### LINKED ISSUE #176: Widget property support
**Enhancement** 

Currently the only way to affect widget state is via actually calling `setState`. This is problematic in scenarios where some of your state is derived from a parent, as if you change anything internally on your widget state - it will be reset by the parent on the next render call.

Its questionable whether `setState` should be available outside of a widget at all, as it seems like that is an implementation detail that only the widget itself should know about.

Other libraries expose `properties` as a way of influencing state, but not directly modifying it. If we supported passing properties in the `options` object to `w` rather than passing state, we can check whether the properties have changed between renders, and fire a `propertiesChanged` function if so which can then make the needed state adjustments. This would solve the problem scenario at the top.

This will also have the nice side effect of having more public like properties that might map to multiple internal state changes.

agubler:
**Type:** feature

**Description:**

Support properties that can be passed to widgets that will be used to internally update state on the widget.

By default only new or updated properties will be used to update state automatically. The functions are public and can be customised by a widget consumer if they require different equality checks for current and previous properties or need to perform some mapping or logic to from the properties to the internal state.

Will be dependant on interfaces/compose changes.

Resolves #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,7 +394,7 @@ const projector = createProjector();
 
 projector.children = [
 	w(createTextInput, { id: 'textinput' }),
-	w(createButton, { id: 'button', label: 'Button' })
+	w(createButton, { id: 'button', properties: { label: 'Button' } })
 ];
 
 projector.append().then(() => {
@@ -416,7 +416,7 @@ const createApp = createProjector.mixin({
 		getChildrenNodes: function(this: Projector): DNode[] {
 			return [
 				w(createTextInput, { id: 'textinput' }),
-				w(createButton, { id: 'button', label: 'Button' })
+				w(createButton, { id: 'button', properties: { label: 'Button' } })
 			];
 		},
 		classes: [ 'main-app' ],
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ A subset of the `Widget` API are intended to be overridden in scenarios where th
 |---|---|---|
 |getNode|Returns the top level node of a widget|Returns a `DNode` with the widgets `tagName`, the result of `this.getNodeAttributes` and `this.children`|
 |getChildrenNodes|Returns the child node structure of a widget|Returns the widgets children `DNode` array|
-|nodeAttributes|An of function that return VNodeProperties to be applied to the top level node|Returns attributes for `data-widget-id`, `classes` and `styles` using the widget's specified `state` (`id`, `classes`, `styles`) at the time of render|
-|diffProperties|Diffs the current properties against the previous properties and returns the updated/new keys in an array|Performs a shallow comparison using `===` of previous and current properties and returns an array of the keys. Function are ignored. |
-|applyChangedProperties|Gets called to process updated/new properties|If there are updared properties they are directly set onto the widgets `state`, `widget.setState(updatedProperties)`|
+|nodeAttributes|An array of functions that return VNodeProperties to be applied to the top level node|Returns attributes for `data-widget-id`, `classes` and `styles` using the widget's specified `state` (`id`, `classes`, `styles`) at the time of render|
+|diffProperties|Diffs the current properties against the previous properties and returns the updated/new keys in an array|Performs a shallow comparison using `===` of previous and current properties and returns an array of the keys.|
+|applyChangedProperties|Gets called to apply updated/new properties|If there are updated properties they are directly set onto the widgets `state` using `setState`.|
 
 To customise the widget an optional `options` argument can be provided with the following interface.
 
@@ -100,8 +100,7 @@ To customise the widget an optional `options` argument can be provided with the 
 |properties|WidgetProperties|Props to be passed to the widget. These can be used to determine state internally|
 |nodeAttributes|Function[]|An array of functions that return VNodeProperties to be applied to the VNode|
 
-A widgets' `state` should **never** be directly set outside of the instance. In order to manipulate widget `state`, `properties` should be updated such as `widget.properties = { 'foo': 'bar' }` and then during the next render the new or updated properties are passed to `instance#processPropertiesChange`.
-
+A widgets' `state` should **never** be directly set outside of the instance. In order to manipulate widget `state`, `properties` should be updated such as `widget.properties = { 'foo': 'bar' }` and then during the next render the new or updated properties are passed to `instance#applyChangedProperties`.
 
 As a convenience, all event handlers are automatically bound to the widget instance, so state and other items on the instance can be easily accessed.
 
--- a/README.md
+++ b/README.md
@@ -255,15 +255,13 @@ registry.define('my-widget', () => {
 
 `w` is an abstraction layer for dojo-widgets that enables dojo 2's lazy instantiation, instance management and caching.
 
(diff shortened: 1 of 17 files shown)


src/createWidgetBase.ts
```diff
@@ -197,16 +216,37 @@ const createWidget: WidgetFactory = createStateful
...
 
-			get id(this: Widget<WidgetState>): string {
+			get id(this: Widget<WidgetState, WidgetProperties>): string {
 				const { id } = widgetInternalStateMap.get(this);
 
-				return id || (this.state && this.state.id) || generateID(this);
+				return id;
+			},
+
+			processPropertiesChange: function(this: Widget<WidgetState, WidgetProperties>, previousProperties: any, currentProperties): void {
```
rishson: `diffProps` and `applyProps`?

src/createWidgetBase.ts
```diff
@@ -147,18 +147,37 @@ function formatTagNameAndClasses(tagName: string, classes: string[]) {
...
 
+function generateProperties(instance: Widget<WidgetState, WidgetProperties>, previousProperties: any): any {
+	const changedPropertyKeys = instance.diffProperties(previousProperties);
+	const changedProperties: { currentProperties: any, previousProperties: any } = {
+		currentProperties: {},
+		previousProperties: {}
+	};
+
+	changedPropertyKeys.forEach((key) => {
+			changedProperties.currentProperties[key] = instance.properties[key];
```
matt-gadd: nit: formatting here

README.md
```diff
@@ -242,13 +256,15 @@ registry.define('my-widget', () => {
 
 `w` is an abstraction layer for dojo-widgets that enables dojo 2's lazy instantiation, instance management and caching.
 
-Creates a dojo-widget using the `factory` and `options`.
+Creates a dojo-widget using the `factory` and `properties`.
+
+**Note:** properties for `id` and `tagName` are automatically copied from the widget properties object and added to the widget options object used when instatiating the widget internally.
```
matt-gadd: this is no longer the case right?

src/createWidgetBase.ts
```diff
@@ -221,40 +261,53 @@ const createWidget: WidgetFactory = createStateful
...
 
 			tagName: 'div'
 		},
-		initialize(instance: Widget<WidgetState>, options: WidgetOptions<WidgetState> = {}) {
-			const { id, tagName } = options;
+		initialize(instance: Widget<WidgetState, WidgetProperties>, options: WidgetOptions<WidgetState, { id?: string }> = {}) {
+			const { tagName, properties = {} } = options;
+			const id = properties.id || options.id || generateID(instance);
+
+			if (!properties.id) {
```
matt-gadd: is this still valid?

src/d.ts
```diff
@@ -36,20 +37,20 @@ export function w<S extends WidgetState, W extends Widget<S>, O extends WidgetOp
...
+export function v(tag: string, properties: VNodeProperties, children?: DNode[]): HNode;
 export function v(tag: string, children: DNode[]): HNode;
 export function v(tag: string): HNode;
-export function v(tag: string, optionsOrChildren: VNodeProperties = {}, children: DNode[] = []): HNode {
+export function v(tag: string, propertiesOrChildren: VNodeProperties = {}, children: DNode[] = []): HNode {
 
-		if (Array.isArray(optionsOrChildren)) {
-			children = optionsOrChildren;
-			optionsOrChildren = {};
+		if (Array.isArray(propertiesOrChildren)) {
```
matt-gadd: this should still be `optionsOrChildren` right?

src/createWidgetBase.ts
agubler: I think it is, as we want to set the `state` to have the `id` from `options` / or generated if there was not an `id` supplied in `properties`

src/d.ts
agubler: I renamed the the property of `v` to `propertiesOrChildren` in line with the type `VNodeProperties`

src/createWidgetBase.ts
matt-gadd: shouldn't it always come from `options` currently or generated though?

src/createWidgetBase.ts
```diff
@@ -197,16 +216,37 @@ const createWidget: WidgetFactory = createStateful
...
 
-			get id(this: Widget<WidgetState>): string {
+			get id(this: Widget<WidgetState, WidgetProperties>): string {
 				const { id } = widgetInternalStateMap.get(this);
 
-				return id || (this.state && this.state.id) || generateID(this);
+				return id;
+			},
+
+			applyChangedProperties: function(this: Widget<WidgetState, WidgetProperties>, previousProperties: Partial<WidgetProperties>, currentProperties: Partial<WidgetProperties>): void {
```
matt-gadd: So I think with the `id` carry over issue, for now we can just apply the `id` in this block like so:
```typescript
applyChangedProperties: function(this: Widget<WidgetState, WidgetProperties>, previousProperties: Partial<WidgetProperties>, currentProperties: Partial<WidgetProperties>): void {
    if (Object.keys(currentProperties).length) {
        currentProperties.id = this.id;
        this.setState(currentProperties);
    }
}
```

src/createWidgetBase.ts
matt-gadd: Okay think I have a slightly cleaner work around for this above (to avoid the options peeling)

README.md
```diff
@@ -69,25 +69,39 @@ These are some of the **important** principles to keep in mind when developing w
...
 
+#### Widget API
+
+A subset of the `Widget` API are intended to be overridden in scenarios where the default implemented behviour does not suffice.
+
+|Function|Description|Default Behaviour|
+|---|---|---|
+|getNode|Returns the top level node of a widget|Returns a `DNode` with the widgets `tagName`, the result of `this.getNodeAttributes` and `this.children`|
+|getChildrenNodes|Returns the child node structure of a widget|Returns the widgets children `DNode` array|
+|nodeAttributes|An of function that return VNodeProperties to be applied to the top level node|Returns attributes for `data-widget-id`, `classes` and `styles` using the widget's specified `state` (`id`, `classes`, `styles`) at the time of render|
```
Tomdye: `An of function` makes no sense

README.md
```diff
@@ -242,13 +256,15 @@ registry.define('my-widget', () => {
...
+Creates a dojo-widget using the `factory` and `properties`.
+
+**Note:** properties for `id` and `tagName` are automatically copied from the widget properties object and added to the widget options object used when instatiating the widget internally.
 
 ```ts
 w(factory: string | ComposeFactory<W, O>, options: O): WNode[];
 ```
 
-Creates a dojo-widget using the `factory` and `options` and the `children`
+Creates a dojo-widget using the `factory` and `properties` and the `children`
```
Tomdye: too many `and`. 
```
Creates a dojo-widget using the `factory`, `properties` and `children``
```

README.md
Tomdye: believe this should be `An array of functions`

PR #183, opened 2016-12-16
Support props to update internal state

### LINKED ISSUE #176: Widget property support

agubler:
**Type:** feature

**This as the same as #181 but using `props` over `properties`.**

**Description:**

Support props that can be passed to widgets that will be used to internally update state on the widget.

By default only new or updated props will be used to update state automatically. The functions are public and can be customised by a widget consumer if they require different equality checks for current and previous props or need to perform some mapping or logic to from the props to the internal state.

Will be dependant on interfaces/compose changes.

Resolves #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ These are some of the **important** principles to keep in mind when developing w
 1. the widget *render* function should **never** be overridden
 2. with the exception of the top level projector you should **never** have to deal with widget instances.
 3. hyperscript should **always** be written using the dojo-widgets `v` helper function.
+4. `state` should **never** be set outside of the widget instance.
 
 ### Base Widget
 
--- a/README.md
+++ b/README.md
@@ -258,6 +258,8 @@ registry.define('my-widget', () => {
 
 Creates a dojo-widget using the `factory` and `properties`.
 
+**Note:** properties for `id` and `tagName` are automatically copied from the widget properties object and added to the widget options object used when instatiating the widget internally.
+
 ```ts
 w(factory: string | ComposeFactory<W, O>, properties: O): WNode[];
 ```
--- a/README.md
+++ b/README.md
@@ -74,20 +74,33 @@ These are some of the **important** principles to keep in mind when developing w
 
 The class `createWidgetBase` provides all base dojo-widgets functionality including caching and widget lifecycle management. It can be used directly or extended to create custom widgets.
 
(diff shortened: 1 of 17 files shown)


agubler:
Going with properties over props

README.md
```diff
@@ -69,25 +69,39 @@ These are some of the **important** principles to keep in mind when developing w
...
 
+#### Widget API
+
+A subset of the `Widget` API are intended to be overridden in scenarios where the default implemented behviour does not suffice.
+
+|Function|Description|Default Behaviour|
+|---|---|---|
+|getNode|Returns the top level node of a widget|Returns a `DNode` with the widgets `tagName`, the result of `this.getNodeAttributes` and `this.children`|
+|getChildrenNodes|Returns the child node structure of a widget|Returns the widgets children `DNode` array|
+|nodeAttributes|An of function that return VNodeProperties to be applied to the top level node|Returns attributes for `data-widget-id`, `classes` and `styles` using the widget's specified `state` (`id`, `classes`, `styles`) at the time of render|
```
matt-gadd: some english fixes needed here 😄

README.md
```diff
@@ -69,25 +69,39 @@ These are some of the **important** principles to keep in mind when developing w
...
+|diffProperties|Diffs the current properties against the previous properties and returns the updated/new keys in an array|Performs a shallow comparison using `===` of previous and current properties and returns an array of the keys. Function are ignored. |
+|processPropertiesChange|Gets called to process updated/new properties|If there are updared properties they are directly set onto the widgets `state`, `widget.state = updatedProperties`|
```
matt-gadd: is now `applyChangedProperties`

PR #195, opened 2016-12-23
Form Fields 174

### LINKED ISSUE #174: Add Form Widgets
**Enhancement** 

Related to #2. Update existing textinput and button components to be accessible, and add the following basic form widgets based on #85:

- [x] Email/number/password/etc to existing textInput
- [x] Checkbox
- [x] Radio buttons
- [x] Select box
- [x] Textarea

smhigley:
**Type:** feature

**Description:**

* Resolves #174.
* Updates createFormFieldMixin to wrap all form fields in a `<label>`, or `<div>` if no label property is provided
* Label options are set in widget properties on widget creation, and can be an object or a string
* Added basic checkbox, radio, textarea, text-like input, and select components


src/components/button/createButton.ts
```diff
@@ -2,33 +2,64 @@ import { ComposeFactory } from 'dojo-compose/compose';
...
 
 const createButton: ButtonFactory = createWidgetBase
-	.mixin(createFormFieldMixin)
 	.mixin({
 		mixin: {
 			nodeAttributes: [
 				function(this: Button): VNodeProperties {
-					return { innerHTML: this.state.label };
+					return {
+						innerHTML: this.properties['content'],
```
agubler: All of these should be able to be accessed via the dot notation like `this.properties.content`

src/mixins/createFormFieldMixin.ts
```diff
@@ -134,6 +168,64 @@ export function stringToValue(str: string): any {
 	}
 }
 
+export function getFormFieldNodeAttributes(state: any): VNodeProperties {
+	const allowedAttributes = ['value', 'type', 'checked', 'descriptionID', 'disabled', 'inputmode', 'maxlength', 'minlength', 'multiple', 'name', 'placeholder', 'readonly', 'required', 'invalid'];
+	const nodeAttributes: any = {};
+
+	for (const key in state) {
```
agubler: Also why can this not just be a `nodeAttribute` function on the instance rather than an exported function?

src/mixins/createFormFieldMixin.ts
```diff
@@ -134,6 +168,64 @@ export function stringToValue(str: string): any {
...
+			nodeAttributes[key] = '' + state[key];
+		}
+		else {
+			nodeAttributes[key] = state[key];
+		}
+	}
+	return nodeAttributes;
+}
+
+function generateWrappedChildren(this: Widget<WidgetState, WidgetProperties> & FormField<any>): DNode[] {
```
agubler: Again any reason why this isn't just the `getChildrenNodes` on the instance?

src/mixins/createFormFieldMixin.ts
```diff
@@ -134,6 +168,64 @@ export function stringToValue(str: string): any {
...
+	const { type, value, label, state } = this;
+	const inputAttributes = getFormFieldNodeAttributes(assign({}, state, <WidgetProperties> { type, value }));
+	const children = [
+		v(this.tagName,
+			inputAttributes,
+			this.getChildrenNodes()
+		)
+	];
+
+	if (label) {
```
agubler: Should we just not wrap all our form fields in a label?

src/mixins/createFormFieldMixin.ts
```diff
@@ -157,25 +249,42 @@ const createFormMixin: FormMixinFactory = createStateful
...
-			{ value, type }: FormFieldMixinOptions<any, FormFieldMixinState<any>> = {}
+			{ value, type, properties }: FormFieldMixinOptions<any, FormFieldMixinState<any>> & WidgetOptions<WidgetState, FormFieldMixinProperties> = {}
 		) {
 			if (value) {
 				instance.setState({ value });
 			}
 			if (type) {
 				instance.type = type;
 			}
+			if (properties && properties.hasOwnProperty('label')) {
```
agubler: Again we shouldn't be doing anything in the initialise really as it doesn't promote reactivity, if the consumer updates the `properties.label` on a subsequent render we should reflect that.

src/mixins/createFormFieldMixin.ts
```diff
@@ -157,25 +249,42 @@ const createFormMixin: FormMixinFactory = createStateful
...
+				const labelDefaults = {
```
agubler: seems like these could be declared for the class not every initialise of an instance.

src/mixins/createFormFieldMixin.ts
```diff
@@ -157,25 +249,42 @@ const createFormMixin: FormMixinFactory = createStateful
...
+					content: '',
+					position: 'below',
+					hidden: false
+				};
+
+				// convert string label to object
+				if (typeof properties['label'] === 'string') {
+					instance.label = assign(labelDefaults, { content: properties['label'] });
```
agubler: why do we need to put this on the instance and not just use from properties during the render cycle?

src/components/button/createButton.ts
```diff
@@ -2,33 +2,64 @@ import { ComposeFactory } from 'dojo-compose/compose';
...
-	label?: string;
+	content?: string;
+	name?: string;
+	type?: string;
+
+	/**
+	 * Accessibility attributes
+	 */
+	descriptionID?: string;
+	disabled?: boolean;
```
agubler: I would argue that disabled is not an accessibility attribute, but of course would be used to add an accessibility attribute plus other things?

src/components/button/createButton.ts
```diff
@@ -2,33 +2,64 @@ import { ComposeFactory } from 'dojo-compose/compose';
...
 			nodeAttributes: [
 				function(this: Button): VNodeProperties {
-					return { innerHTML: this.state.label };
+					return {
+						innerHTML: this.properties['content'],
+						type: this.properties['type'],
+						name: this.properties['name'],
+						role: this.properties['role'],
+						disabled: this.state.disabled,
+						'aria-pressed': this.state.pressed,
```
agubler: How do you the intend `role` state to be updated?

smhigley: There are some attributes like `name`, `role`, `type`, and `aria-describedby` that I don't think should be updated after creation. In my opinion it would be unnecessarily confusing to assistive tech, and it's hard to think of a valid use case. If you think it goes against the pattern for creating widgets, though, I can change it to access them through `state`.

src/components/button/createButton.ts
```diff
@@ -2,33 +2,64 @@ import { ComposeFactory } from 'dojo-compose/compose';
...
+						'aria-describedby': this.properties['descriptionID'],
+						'aria-haspopup': this.properties['hasPopup']
+					};
 				}
 			],
-			tagName: 'button',
-			type: 'button'
```
smhigley: There are different possible types, but I don't think it should change after creation (same reason as above). Having a type of `button` with the tagname `button` is redundant though, so I removed it as a default.

src/components/textinput/createTextInput.ts
```diff
@@ -10,7 +10,11 @@ export interface TypedTargetEvent<T extends EventTarget> extends Event {
 
 export type TextInputState = WidgetState & FormFieldMixinState<string>;
 
-export type TextInputOptions = WidgetOptions<TextInputState, WidgetProperties> & FormFieldMixinOptions<string, TextInputState>;
+export interface TextInputProperties extends WidgetProperties {
+	type?: string;
```
smhigley: I don't personally think it should. I'm open to not enforcing that, if you think it should be accessed via `state`

src/mixins/createFormFieldMixin.ts
```diff
@@ -17,14 +18,18 @@ export interface FormFieldMixinOptions<V, S extends FormFieldMixinState<V>> exte
...
+
 }
 
-export interface FormFieldMixinState<V> {
-	/**
-	 * Whether the field is currently disabled or not
-	 */
-	disabled?: boolean;
+export interface FormFieldMixinProperties extends WidgetProperties {
+	label?: string | {
```
smhigley: Content could be mandatory, that seems like a good idea. Did you want `hidden` to be a string instead of a boolean, or was that just a mistype?

src/components/button/createButton.ts
agubler: I put the wrong words in the original comment so it made very little sense!

Issue #202, opened 2017-01-05
More granular control of property assignment


pottedmeat:
**Enhancement**

The widget lifecycle currently has no way to define how properties are stored for later comparison. Currently, comparison may be customized using `diffProperties(previousProperties)` but `internalState.previousProperties` will always make a full copy of `instance.properties`.

By introducing a hook that would allow the widget to decide what properties are stored and what values they are stored with, we could significantly decrease the overhead of property serialization with `deepAssign` and give widgets increased flexibility in making decisions about property change detection.

We should decide whether this hook should allow widget authors to use reference comparison as a valid way of comparing property values. I would argue this should be allowed, though designed in such a way as to encourage a more specific method of value comparison.

**Proposal**

* Add a new method alongside `diffProperties` and `applyChangedProperties` that will take the current properties and return the keys and values that should be used for comparison.
* Change `diffProperties` to accept the previous properties and the result of this new method as its current properties. If the widget author wanted to also use the original properties, they would be available using `this.properties`.
* Create a wrapper function for storing values by reference using something like a `WeakMap` that stores symbols or otherwise unique objects keyed by these instances so that two calls to this function with the same instance would return identical values.
* Continue using something like `deepAssign` when storing to `interrnalState.previousProperties` with the exception of leaving these wrapped references intact.

matt-gadd:
@pottedmeat going to close this as I believe #212 resolves the majority of this issue. feel free to reopen if it doesn't! 😄

pottedmeat:
I was able to make some great improvements to dgrid with this but eventually ran into areas where I don't quite have enough control over the lifecycle of the properties. Reading through what I wrote above, I was unclear about the details of this.

## When `properties != previousProperties`

Let's say I have an object where every change to the object results in the increase of an internal counter stored under the `version` key. I'd still like the full object to be accessible by the rest of the code through the `properties` key but I'd like to store just the `version` value in `internalState.previousProperties` for later comparison.

With the current API, there is no way for me to leave the object intact on `properties` while saving a serialized/hashed/versioned piece of data in `internalState.previousProperties`

## When Calculating a Diffable Value is Expensive

In the case where an object I'm working with is complicated enough where the serialized/hashed value is the result of a significant amount of logic/memory/processing power, the current API would require this value be calculated twice.

1. In `diffProperties`, I would need to calculate the value to compare to `previousProperties` in order to return the keys that have changed. I would not be able to assign this value to `newProperties` as it would override the actual value.
[...]

PR #212, opened 2017-01-10
Refine properties lifecycle, add public setProperties()

### LINKED ISSUE #207: Property lifecycle improvements
**Enhancement** 
At the moment the widget property lifecycle provides diffing and application hooks. If a mixin provides certain diff capabilities, it seems like it also needs to be capable of deciding what properties actually get copied over. 

At the moment we simply deepAssign all of the properties (https://github.com/dojo/widgets/blob/master/src/createWidgetBase.ts#L268). Given the point of keeping the previous/current properties is for diffing between render cycles - it seems like you'd want control over the copying phase and that it would match the depth of whatever was done in the diffing phase. This would mean we'd be able to drop `deepAssign` by default (as its quite costly if providing reasonably complex objects).

We would change the lifecycle hooks to be something along the lines of:
```typescript
diffProperties(previousProperties: WidgetProperties, newProperties: WidgetProperties): string[]
copyProperties(previousProperties: WidgetProperties, newProperties: WidgetProperties, propertiesChanged: string[]): WidgetProperties
applyProperties(properties: WidgetProperties, propertiesChanged: string[]): void
```
Where the result returned from `copyProperties` is assigned to `this.properties`.

This should partially fix some of the concerns here: https://github.com/dojo/widgets/issues/202

matt-gadd:
**Type:** feature

**Description:**

Resolves #207, #202 

This refines the property lifecycle for widgets. The lifecycle is itself now triggered by a `setProperties` call rather than the `__render__` call.

The new lifecycle is as follows:
```ts
diffProperties(previousProperties: P, newProperties: P): string[];
assignProperties(previousProperties: P, newProperties: P, changedPropertyKeys: string[]): P;
```

`diffProperties` is responsible for diffing the previous and new properties - returning an array of property keys that have changed.
`assignProperties` returns a set of properties to be assigned to `this.properties` for usage later.

If `diffProperties` returns an array of length > 0 an event of `properties:changed` is triggered, as well as `onPropertiesChanged` being called.

These new lifecycle methods and events should give control of property diffing and assignment, as well as reacting to changes.

Notes: there are a few any hammers due to the issues with TS that will be resolved in 2.1.5.

### DIFF
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -199,7 +199,14 @@ const createWidget: WidgetFactory = createStateful
 				const internalState = widgetInternalStateMap.get(this);
 				const changedPropertyKeys = this.diffProperties(internalState.previousProperties, properties);
 				this.properties = this.assignProperties(internalState.previousProperties, properties, changedPropertyKeys);
-				this.applyProperties(this.properties, changedPropertyKeys);
+				if (changedPropertyKeys.length) {
+					this.emit({
+						type: 'properties:changed',
+						target: this,
+						properties: this.properties,
+						changedPropertyKeys
+					});
+				}
 				internalState.previousProperties = this.properties;
 			},
 
@@ -211,14 +218,12 @@ const createWidget: WidgetFactory = createStateful
 				return assign({}, newProperties, { id: this.id });
 			},
 
-			applyProperties: function(this: Widget<WidgetState, WidgetProperties>, properties: WidgetProperties, changedPropertyKeys: string[]): void {
-				if (changedPropertyKeys.length) {
-					const state = changedPropertyKeys.reduce((state, key) => {
-						(<any> state)[key] = (<any> properties)[key];
-						return state;
-					}, <WidgetProperties> {});
-					this.setState(state);
-				}
+			onPropertiesChanged: function(this: Widget<WidgetState, WidgetProperties>, properties: WidgetProperties, changedPropertyKeys: string[]): void {
+				const state = changedPropertyKeys.reduce((state, key) => {
+					(<any> state)[key] = (<any> properties)[key];
+					return state;
+				}, <WidgetProperties> {});
+				this.setState(state);
 			},
 
 			nodeAttributes: [
@@ -276,11 +281,15 @@ const createWidget: WidgetFactory = createStateful
 				children: []
 			});
 
-			instance.setProperties(properties);
+			instance.own(instance.on('properties:changed', (evt: any) => {
+				instance.onPropertiesChanged(evt.properties, evt.changedPropertyKeys);
+			}));
 
 			instance.own(instance.on('state:changed', () => {
 				instance.invalidate();
 			}));
+
+			instance.setProperties(properties);
 		}
 	})
 	.mixin(shallowPropertyComparisonMixin);
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -201,7 +201,6 @@ const createWidget: WidgetFactory = createStateful
 				this.properties = this.assignProperties(internalState.previousProperties, properties, changedPropertyKeys);
 				this.applyProperties(this.properties, changedPropertyKeys);
 				internalState.previousProperties = this.properties;
-				this.invalidate();
 			},
 
 			diffProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: WidgetProperties, newProperties: WidgetProperties): string[] {
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -213,8 +213,12 @@ const createWidget: WidgetFactory = createStateful
 			},
 
 			applyProperties: function(this: Widget<WidgetState, WidgetProperties>, properties: WidgetProperties, changedPropertyKeys: string[]): void {
-				if (Object.keys(properties).length) {
-					this.setState(properties);
+				if (changedPropertyKeys.length) {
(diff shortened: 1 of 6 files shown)


src/createWidgetBase.ts
```diff
@@ -212,15 +196,35 @@ const createWidget: WidgetFactory = createStateful
 				return this.properties.id;
 			},
 
-			applyChangedProperties: function(this: Widget<WidgetState, WidgetProperties>, previousProperties: WidgetProperties, currentProperties: WidgetProperties): void {
-				if (Object.keys(currentProperties).length) {
-					currentProperties['id'] = this.id;
-					this.setState(currentProperties);
+			setProperties(this: Widget<WidgetState, WidgetProperties>, properties: WidgetProperties) {
```
agubler: Is there any reason we didn't make this a setter? i.e. `set properties(this: Widget...)`

src/interfaces.d.ts
```diff
@@ -25,6 +25,12 @@ export interface ChildNodeFunction {
 	(this: Widget<WidgetState, WidgetProperties>): DNode[];
 }
 
+export interface PropertiesChangeEvent<T> extends EventTypedObject<'properties:changed'> {
```
agubler: I think we should look to generify the `properties` attribute and add the `on(...` signature to the `WidgetOverrides` type, this is so that a user can then get the `on` intellisense for the `properties:changed` evnet.

src/interfaces.d.ts
```diff
@@ -145,15 +151,19 @@ export interface WidgetMixin<P extends WidgetProperties> {
 	 */
 	properties: P;
 
+	setProperties(this: Widget<WidgetState, WidgetProperties>, properties: P): void;
```
agubler: doc

src/interfaces.d.ts
```diff
@@ -145,15 +151,19 @@ export interface WidgetMixin<P extends WidgetProperties> {
...
+
 	/**
 	 * Determine changed or new property keys on render.
 	 */
-	diffProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: P): string[];
+	diffProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: P, newProperties: P): string[];
+
+	assignProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: P, newProperties: P, changedPropertyKeys: string[]): P;
```
agubler: doc

src/mixins/shallowPropertyComparisonMixin.ts
```diff
@@ -1,11 +1,13 @@
 import { entries } from 'dojo-shim/object';
 import { WidgetProperties } from './../interfaces';
+import { deepAssign } from 'dojo-core/lang';
 
 /**
  * Interface for `diffProperties`
  */
 export interface ShallowPropertyComparisonMixin {
-	diffProperties<T>(previousProperties: T): string[];
+	diffProperties<T>(previousProperties: T, newProperties: T): string[];
```
agubler: Do you think we should extract these to a separate generic API that `shallowPropertyComparisonMixin` can implement and `WidgetMixin` can also extend.

If we are saying for any mixin that provides comparison functionality will need to override these function?

src/createWidgetBase.ts
matt-gadd: I think the rationale was:
* as a setter users may try to do `myWidget.properties.blah`, which wouldn't work
* as a setter although it does end up assigning to `this.properties` it runs a lifecycle before it; so isn't a pure setter
* the function matches `setState` in symmetry

src/interfaces.d.ts
matt-gadd: agreed, looks like this would be similar to the `invalidate` in terms of typing on the `WidgetOverrides`

src/interfaces.d.ts
matt-gadd: 👍

src/mixins/shallowPropertyComparisonMixin.ts
matt-gadd: Yup, I think thats a very good idea

src/createWidgetBase.ts
```diff
@@ -212,15 +196,35 @@ const createWidget: WidgetFactory = createStateful
...
+						changedPropertyKeys
+					});
 				}
+				internalState.previousProperties = this.properties;
 			},
 
-			diffProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: any): string[] {
-				return Object.keys(this.properties);
+			diffProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: WidgetProperties, newProperties: WidgetProperties): string[] {
+				return Object.keys(newProperties);
```
agubler: This cannot be covered in unit tests as a mixin (`shallowPropertyComparisonMixin `) is used by default for `createWidgetBase` that overrides this functions so there is no way to test.

src/createWidgetBase.ts
```diff
@@ -212,15 +196,35 @@ const createWidget: WidgetFactory = createStateful
...
+			},
+
+			assignProperties(this: Widget<WidgetState, WidgetProperties>, previousProperties: WidgetProperties, newProperties: WidgetProperties, changedPropertyKeys: string[]): WidgetProperties {
+				return assign({}, newProperties);
```
agubler: Same as `diffProperties` this is overridden by default so cannot be covered by unit tests.

matt-gadd:
kudos to @agubler for the typings fixes!

PR #227, opened 2017-01-12
Feature theme manager

### LINKED ISSUE #219: Theme manager
A `themeManager` is needed to hold the current theme and provide widgets with theme classes.
The themeManager will be a singleton accessed in each widget that has themeable elements and will provide intelisense for typed class names.

Should provide:

### setTheme
Setter for the loaded theme.

### clearTheme
Clears the current theme.

### getThemeClasses
Function that accepts the base theme classes from a widget along with it's override classes. Should use the base theme classes to pluck any theme classes out of the loaded theme and the overrides provided.

Tomdye:
** Feature **

**Description:**

Adds a theme manager capable of merging base and theme classes with optional override classes.

Resolves #219

### DIFF
--- a/src/theme/themeManager.ts
+++ b/src/theme/themeManager.ts
@@ -1,5 +1,3 @@
-import compose, { ComposeFactory } from 'dojo-compose/compose';
-
 export type CSSModuleClassNames = {
 	[key: string]: boolean;
 }
@@ -12,20 +10,13 @@ type Theme = {
 	[key: string]: string;
 }
 
-export interface ThemeManager {
-	setTheme(theme: {}): void;
-	getThemeClasses<T extends {}>(baseThemeClasses: T, overrideClasses?: {}): AppliedClasses<T>;
-}
-
-export interface ThemeManagerFactory extends ComposeFactory<ThemeManager, {}> { }
-
 type CacheKey = {
 	baseThemeClasses: {};
 	overrideClasses?: {};
 };
 
-const themeMap = new WeakMap<ThemeManager, Theme>();
-const themeManagerCacheMap = new WeakMap<ThemeManager, Map<CacheKey, AppliedClasses<any>>>();
+let currentTheme: Theme = {};
+let cacheMap = new Map<CacheKey, AppliedClasses<any>>();
 
 function addClassNameToMap(classMap: CSSModuleClassNames, classList: Theme, className: string) {
 	if (classList && classList.hasOwnProperty(className)) {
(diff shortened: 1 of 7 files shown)


src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+		});
+	}
+}
+
+export function setTheme(theme: {}) {
+	currentTheme = theme;
+	cacheMap = new Map<CacheKey, AppliedClasses<{}>>();
+};
+
+export function getClasses<T extends {}>(baseThemeClasses: T, overrideClasses?: {}): AppliedClasses<T> {
```
agubler: Could override classes be typed to `T` also?

src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+				themeClassSource = currentTheme;
+			}
+
+			addClassNameToMap(classMap, themeClassSource, className);
+			overrideClasses && addClassNameToMap(classMap, overrideClasses, className);
+
+			return currentAppliedClasses;
+		}, <AppliedClasses<T>> {});
+
+		cacheMap.set(cacheKey, appliedClasses);
```
agubler: why don't we return the `appliedClasses` rather then requiring a map look up for something you already have?

src/theme/themeManager.ts
agubler: Just from a readability perspective I tend to prefer

```ts
if (cacheMap.has(cacheKey)) {
    return cacheMap.get(cacheKey);
}

// ... the other stuff

return appliedClasses;
```

src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+type CacheKey = {
+	baseThemeClasses: {};
+	overrideClasses?: {};
+};
+
+let currentTheme: Theme = {};
+let cacheMap = new Map<CacheKey, AppliedClasses<any>>();
+
+function addClassNameToMap(classMap: CSSModuleClassNames, classList: Theme, className: string) {
+	if (classList && classList.hasOwnProperty(className)) {
```
agubler: do we need the check on `classList` as it is not optional?

src/theme/themeManager.ts
agubler: isn't it actually `getClasses<T extends Theme>....`?

src/theme/themeManager.ts
agubler: This can also be generic I think `addClassNameToMap<T extends Theme>(classMap: CSSModuleClassNames, classList: T, className: string)`? but perhaps that doesn't add anything though....

src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+export type AppliedClasses<T> = {
+	[P in keyof T]?: CSSModuleClassNames;
+}
+
+type Theme = {
+	[key: string]: string;
+}
+
+type CacheKey = {
+	baseThemeClasses: {};
```
agubler: Again this is `Theme`, no?

src/theme/themeManager.ts
Tomdye: I don't think the generic would add anything here, the typing for the response object is keyed off of the widget's base classes, not the overrides or theme classes.

src/theme/themeManager.ts
Tomdye: code removed

src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+	overrideClasses?: {};
+};
+
+let currentTheme: Theme = {};
+let cacheMap = new Map<CacheKey, AppliedClasses<any>>();
```
Tomdye: code removed

src/theme/themeManager.ts
```diff
@@ -0,0 +1,58 @@
...
+	if (classList && classList.hasOwnProperty(className)) {
+		// split out the classname because css-module composition combines class names with a space
+		const generatedClassNames: string[] = classList[className].split(' ');
+		generatedClassNames.forEach((generatedClassName) => {
+			classMap[generatedClassName] = true;
+		});
+	}
+}
+
+export function setTheme(theme: {}) {
```
Tomdye: I've tried that before, but then other parts start to fail, might be easier to ditch `Theme` and put `{}` for everything. Theme doesn't really give us anything.

src/theme/themeManager.ts
Tomdye: Don't think so, because the return object is only keyed off of the baseThemeClasses. user might have put a superset in overrideClasses.

Tomdye: code removed

src/theme/themeManager.ts
Tomdye: code removed

Tomdye:
@agubler comments addressed.

Tomdye:
Closing in favour of themeable mixin

PR #237, opened 2017-01-13
Allow setting a single DNode for children

### LINKED ISSUE #157: Allow setting a single DNode for children.
**Enhancement**

Extend the children setter API to allow `DNode` and `DNode[]` and set accordingly.

Package Version: latest

agubler:
**Type:** feature

**Description:**

Add `setChildren(children: DNode | DNode[])` api to replace the overridden property setter that can consume a single `DNode` as well as an array of `DNode`s

Resolves #157

### DIFF
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -105,7 +105,7 @@ function dNodeToVNode(instance: Widget<WidgetProperties>, dNode: DNode): VNode |
 			instance.emit({ type: 'error', target: instance, error: new Error(errorMsg) });
 		}
 
-		child.children = children;
+		child.setChildren(children);
 		internalState.currentChildrenMap.set(childrenMapKey, child);
 
 		return child.__render__();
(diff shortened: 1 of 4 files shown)


src/createWidgetBase.ts
```diff
@@ -151,19 +151,19 @@ const createWidget: WidgetBaseFactory = createStateful
...
 
-			set children(this: Widget<WidgetProperties>, children: DNode[]) {
+			get children(this: Widget<WidgetProperties>) {
+				return widgetInternalStateMap.get(this).children;
+			},
+
+			setChildren(this: Widget<WidgetProperties>, children: DNode | DNode[]): void {
 				const internalState = widgetInternalStateMap.get(this);
-				internalState.children = children;
+				internalState.children = Array.isArray(children) ? children : [ children ];
```
dylans: While this is of course fine, two thoughts come to mind when I see this pattern:

1. Should we have a core arrayUtil method to do things like this (probably not as it's probably more expensive, but it might be cleaner, especially if something like this is eventually added to the language.
2. Are we assuming things are arrays that should also support iterables (probably doesn't matter here in this example, but in general throughout Dojo 2, are we adequately supporting iterables or assuming arrays).

agubler:
Closing in favour of keeping the array API, will re-raise PR for the `readonly` parts

PR #251, opened 2017-01-16
Widgets README update

agubler:
**Type:** feature

Update the readme in line with the latest code changes. This is by no means fully finalised but is more accurate/up to date than the existing readme.

Easier to review using https://github.com/agubler/widgets/blob/new-readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,10 +270,10 @@ const createI18nWidget = createWidgetBase
 				const messages = this.localizeBundle(greetingsBundle);
 
 				return [
-					d(createLabel, {
+					w(createLabel, {
 						// Passing a message string to a child widget.						label: messages.purchaseItems
 					}),
-					d(createButton, {
+					w(createButton, {
 						// Passing a formatted message string to a child widget.
 						label: messages.format('itemCount', { count: 2 })
 					})
--- a/README.md
+++ b/README.md
@@ -68,8 +68,8 @@ These are some of the **important** principles to keep in mind when creating and
 1. the widget *`__render__`* function should **never** be overridden
 2. except for projectors you should **never** need to deal directory with widget instances.
 3. hyperscript should **always** be written using the @dojo/widgets `v` helper function.
-4. `state` should **never** be set outside of the widget instance.
-5. should **never** update `properties` within a widget instance.
+4. **never** set state outside of the widget instance.
+5. **never** update `properties` within a widget instance.
 
 ### Overview
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Provides Dojo2 core widget and mixin functionality for creating custom widgets. 
     	- [Introducing `v` & `w`](#v--w)
     	- [Widget Registry](#widget-registry)
     	- [Properties Lifecycle](#properties-lifecycle)
-    	- [Event Handlers](#event-handlers)
+    	- [Event Handling](#event-handling)
     	- [Internationalization](#internationalization)
     	- [Projector](#projector)
     - [Authoring Examples](#authoring-examples)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,34 +2,35 @@
 
 [![Build Status](https://travis-ci.org/dojo/widgets.svg?branch=master)](https://travis-ci.org/dojo/widgets)
 [![codecov](https://codecov.io/gh/dojo/widgets/branch/master/graph/badge.svg)](https://codecov.io/gh/dojo/widgets)
-[![npm version](https://badge.fury.io/js/@dojo/widgets.svg)](http://badge.fury.io/js/@dojo/widgets)
+[![npm version](https://badge.fury.io/js/%40dojo%2Fwidgets.svg)](https://badge.fury.io/js/%40dojo%2Fwidgets)
 
-A core widget library for Dojo.
+Provides Dojo2 core widget and mixin functionality for creating custom widgets. For Dojo2 widgets please see [@dojo/widgets](https://github.com/dojo/widgets).
```
dylans: We should be saying Dojo 2 rather than Dojo2, to avoid weirdness when we release Dojo 3. :)

README.md
```diff
@@ -39,133 +40,35 @@ npm install @dojo/has
...
-```
-
-or use the [dojo cli](https://github.com/dojo/cli-create-app) to create a complete Dojo skeleton application.
+Use the [@dojo/cli](https://github.com/dojo/cli) to create a complete Dojo skeleton application with the [@dojo/cli-create-app](https://github.com/dojo/cli-create-app) command.
 
 ## Features
 
-@dojo/widgets are based on a virtual DOM implementation named [Maquette](http://maquettejs.org/) as well as some foundational classes
-provided in [@dojo/compose](https://github.com/dojo/compose).
+Dojo 2 widgets are built using the [@dojo/compose](https://github.com/dojo/compose) composition library that promotes trait mixins over inheritence.
```
dylans: Is it really trait mixins, or traits and mixins?

README.md
```diff
@@ -39,133 +40,35 @@ npm install @dojo/has
...
 
-```html
-<div></div>
-```
-The following example demonstrates how `id`, `classes` and `styles` are applied to the generated DOM.
+### Overview
 
-```ts
-import createWidgetBase from '@dojo/widgets/createWidgetBase';
+Dojo2 widgets is designed using key reactive architecture concepts. These include unidirectional data flow, inversion of control and property passing.
```
dylans: Dojo 2

README.md
```diff
@@ -39,133 +40,35 @@ npm install @dojo/has
...
-```
-
-#### `d`
-
-`d` is the module that provides the functions and helper utilities that can be used to express widget structures within Dojo 2. This structure constructed of `DNode`s (`DNode` is the intersection type of `HNode` and `WNode`).
-
-When creating custom widgets, you use the `v` and `w` functions from the `d` module.
-
-It is imported by:
+`v` & `w` are abstractions used to express structures that will be reflected in DOM. `v` is used to create nodes that represent DOM tags for example `div`, `header` etc and allows Dojo 2 to manage lazy hyperscript creation and element caching. `w` is used to create Dojo or custom widget nodes enabling support for lazy widget instantiation, instance management and caching.
```
dylans: DOM tags, for example (add comma)
etc. instead of etc

README.md
```diff
@@ -503,37 +270,159 @@ const createI18nWidget = createWidgetBase
...
+export type Label = Widget<LabelProperties>
+
+export interface LabelFactory extends WidgetFactory<Label, LabelProperties> {}
+
+const createLabelWidget: LabelFactory = createWidgetBase.mixin({
+    mixin: {
+        tagName: 'label',
+        nodeAttributes: [
+            function(this: Label): VNodeProperties {
+                return { innerHTML: this.properties.label };
```
smhigley: I know it's nitpicky, but I'd like to disassociate `label` from `innerHTML`. Could we use `this.properties.content` or `this.properties.text` here? I'd like to use `label` to explicitly refer to labels

agubler: sure no worries (i think it's the same as the example that is there now)

smhigley: Thanks! (You're right, it is what's in there now. Just hadn't noticed it before :P)

agubler:
@dylans @smhigley I think I have covered the requested changes.

rishson:
I like the changes - makes everything more 'best practise'.

Issue #106, opened 2016-11-16
a11y Best Practices


kitsonk:
# User Story

Developers of Dojo 2 should have clear guidance on how to make their applications accessible and how best to ensure their applications are accessible.

# Solution

Most of the best practices will be for custom widget authors, which is guidance about how best to ensure their widgets will behave will when integrated into an application and that the DOM a widget generates is accessible.

dylans:
We may want to consider the [`prefers-reduced-motion` media query](https://github.com/w3c/csswg-drafts/issues/442) somewhere in our approach for a11y. Browser support is in Safari 10.1, iOS 10.3, and perhaps others. https://css-tricks.com/introduction-reduced-motion-media-query/ has a decent summary.

There are also plans to add [CSS media query settings for contrast and transparency](https://github.com/w3c/csswg-drafts/issues/443).

smhigley:
@dylans I like it! I think it would be nice to provide, with a caveat saying "we still recommend you make all your animations accessibility-friendly." Should we provide alternate styles for that and high contrast in themes based on a property that adds a class to the projector?

rishson:
High-contrast theme would be really useful (as long as it doesn't clash with a user putting their OS in high contrast mode) - is that what you were suggesting?

smhigley:
@rishson yeah, I did mean a separate high-contrast theme. Morning brain :). More contrast-y styles won't conflict with OS high contrast mode.

dylans:
Also, should we publish the aria guidelines from our Google doc as part of the meta perhaps (either as a wiki or a document), or plan to have them as part of the reference guide?

kitsonk:
Issue moved to [dojo/dojo.io #207](https://github.com/dojo/dojo.io/issues/207) via [**ZenHub**](https://www.zenhub.com/)

PR #263, opened 2017-01-18
FormLabelMixin

smhigley:
**Type:** feature

**Description:**

Relates to #174 and #106. Removes FormFieldMixin in favor of FormLabelMixin, which wraps all form fields in a `<label>` element and passes valid a11y attributes to the input node.

Also removes the Button and TextInput components, as they will soon be moved to a new repository and currently depend on FormFieldMixin


src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+	/**
+	 * Label settings for form label text, position, and visibility
+	 */
+	label?: string | FormLabelProperties;
+
+	/**
+	 * Accessibility attributes
+	 */
+	checked?: boolean;
+	descriptionID?: string;
```
dylans: we probably want to call this descriptionId for naming consistency, though that reads weird (I vs. l problem). I'd suggest seeing if we have a better name.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+	inputmode?: string;
```
dylans: we should probably be camelCase for consistency, unless there's a reason not to on these? Or all lowercase if that's more consistent with HTML attributes?

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+					}
+					else {
+						label = assign({}, labelDefaults, label);
+					}
+
+					// add label text
+					if (label.content.length > 0) {
+						children.push(v('span', {
+							innerHTML: label.content,
+							classes: { 'visually-hidden': label.hidden }
```
dylans: Does visually-* follow a particular pattern (not opposed to it, just wondering if there's a naming convention we're following).

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+						}));
+					}
+
+					// set correct order
+					if (label.position === 'above') {
```
dylans: Thinking about things like RTL and vertical languages... in Dojo 1.x, we had things like leading/trailing as aliases for above/below... not sure what the correct answer is, or if we need to add anything here for such use cases (can defer the thought until later also if we want).

tests/unit/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+			}
+		});
+
+		let vnode = <VNode> formfield.__render__();
+		let inputfield = vnode.children![0];
+
+		assert.strictEqual(inputfield.vnodeSelector, 'input');
+		assert.strictEqual(inputfield.properties!['value'], 'foo');
+		assert.strictEqual(inputfield.properties!['maxlength'], '100');
+		assert.isUndefined(inputfield.properties!['randomProp']);
```
dylans: should it be randomprop for consistency, or should property names be camelCase except when they reflect an html attribute name?

tests/unit/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+			tagName: 'input',
+			properties: {
+				value: 'foo',
+				maxlength: 100,
+				randomProp: 'qux'
+			}
+		});
+
+		let vnode = <VNode> formfield.__render__();
+		let inputfield = vnode.children![0];
```
dylans: inputfield or inputField?

tests/unit/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,109 @@
...
+
+registerSuite({
+	name: 'mixins/createFormLabelMixin',
+	construction() {
+		const formLabelMixin = createFormLabelMixin();
+
+		assert.isDefined(formLabelMixin);
+	},
+	getNode() {
+		const formfield = formLabelWidget.override({
```
dylans: formfield or formField?

src/mixins/createFormLabelMixin.ts
smhigley: You're right, camel case for "id" always trips me up. I'd definitely be up for thinking of a better name. Perhaps just "describedBy"?

src/mixins/createFormLabelMixin.ts
smhigley: I left inputmode lowercase to keep it the same as the HTML attribute. I suppose we could put 'aria-describedby' in quotes and make all of them correspond directly to HTML attributes, if you think that's more consistent.

src/mixins/createFormLabelMixin.ts
smhigley: It's a mix of a personal naming convention and a broader one. HTML5 Boilerplate uses .visuallyhidden, and I think dashes in between words makes them more readable 😄. I'd be fine switching to .visuallyhidden, or even .visuallylHidden. The "visually" part is to differentiate it from methods of hiding that use `{ display: none }` or `{ visibility: hidden }`.

src/mixins/createFormLabelMixin.ts
smhigley: Maybe "before"/"after" would be better?

src/interfaces.d.ts
```diff
@@ -301,3 +301,13 @@ export interface WidgetState extends State {
 }
 
 export interface WidgetFactory<W extends Widget<P>, P extends WidgetProperties> extends ComposeFactory<W, WidgetOptions<WidgetState, P>> {}
+
+export interface TypedTargetEvent<T extends EventTarget> extends Event {
+	target: T;
+}
+
+export interface FormLabelProperties {
```
agubler: This can just go in the FormLabelMixin.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
+import { VNodeProperties } from '@dojo/interfaces/vdom';
+import compose, { ComposeFactory } from '@dojo/compose/compose';
+import { assign } from '@dojo/core/lang';
+import { NodeAttributeFunction, DNode, Widget, WidgetOptions, WidgetState, WidgetProperties, FormLabelProperties } from './../interfaces';
```
agubler: do you think we should put line breaks in here now (it's got quite long).

```ts
import {
    This,
    That,
    Etc,
} from 'somewhere';
```

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+	 */
+	getFormFieldNodeAttributes: NodeAttributeFunction<this>;
+
+	/**
+	 * Override the default getNode function on createWidgetBase to return child input
+	 */
+	getNode(): DNode;
+}
+
+export type FormLabel = Widget<FormLabelMixinProperties> & FormLabelMixin;
```
agubler: I think we should provide doc on all exported interfaces, types, functions, etc.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+				const inputAttributes = this.getFormFieldNodeAttributes.call(this);
+				const children = [
+					v(this.tagName,
+						inputAttributes,
+						result
+					)
+				];
+
+				if (label) {
+					const labelDefaults = {
```
agubler: This can be a constant in the module I think

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+	}
+})
+.mixin({
+	aspectAdvice: {
+		after: {
+			getChildrenNodes(this: FormLabel, result: DNode[]) {
+				let { label } = this.properties;
+				const inputAttributes = this.getFormFieldNodeAttributes.call(this);
+				const children = [
+					v(this.tagName,
```
agubler: feels like this would read okay being on one line `v(this.tagName, intputAttributes, result)`

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+
+		return v(tag, this.getNodeAttributes(), this.getChildrenNodes());
+	}
+})
+.mixin({
+	aspectAdvice: {
+		after: {
+			getChildrenNodes(this: FormLabel, result: DNode[]) {
+				let { label } = this.properties;
+				const inputAttributes = this.getFormFieldNodeAttributes.call(this);
```
agubler: Why do you need to `.call` here?

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+		let tag = label ? 'label' : 'div';
+
+		if (this.classes.length) {
+			tag = `${tag}.${this.classes.join('.')}`;
+		}
+
+		return v(tag, this.getNodeAttributes(), this.getChildrenNodes());
+	}
+})
+.mixin({
```
agubler: I don't think you need the `.mixin` and can just put the `aspectAdvice` in using the `.aspect` API provided by compose.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+	aspectAdvice: {
+		after: {
+			getChildrenNodes(this: FormLabel, result: DNode[]) {
```
agubler: Can we add the return type to be clear `DNode[]`

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+const createFormLabelMixin: FormLabelMixinFactory = compose({
+	getFormFieldNodeAttributes(this: FormLabel): VNodeProperties {
+		const { properties, type } = this;
+		const attributeKeys = Object.keys(properties);
+
+		if (type) {
+			attributeKeys.push('type');
+		}
+
+		const allowedAttributes = ['checked', 'descriptionID', 'disabled', 'inputmode', 'invalid', 'maxlength', 'minlength', 'multiple', 'name', 'placeholder', 'readonly', 'required', 'type', 'value'];
```
agubler: Another module level constant perhaps?

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,164 @@
...
+		const nodeAttributes: any = {};
+
+		for (const key of allowedAttributes) {
+
+			if (attributeKeys.indexOf(key) === -1) {
+				continue;
+			}
+			else if (key === 'type') {
+				nodeAttributes['type'] = type;
```
agubler: as you have typed this as `any` you can certainly use the dot notation for accessing properties. But I think you would need to do it via the index if you typed `nodeAttributes` to `VNodeProperties` (which I believe it is).

smhigley: VNodeProperties types most of these as readonly, sadly enough. I'll change them to dot notation, though.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,220 @@
...
+	position: 'after',
+	hidden: false
+};
+
+const createFormLabelMixin: FormLabelMixinFactory = compose({
+	getFormFieldNodeAttributes(this: FormLabel): VNodeProperties {
+		const { properties, type } = this;
+		const attributeKeys = Object.keys(properties);
+
+		if (type) {
```
agubler: missing a test where `type` is available.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,220 @@
...
+
+		const allowedAttributes = ['checked', 'describedBy', 'disabled', 'invalid', 'maxLength', 'minLength', 'multiple', 'name', 'placeholder', 'readOnly', 'required', 'type', 'value'];
+		const nodeAttributes: any = {};
+
+		for (const key of allowedAttributes) {
+
+			if (attributeKeys.indexOf(key) === -1) {
+				continue;
+			}
+			else if (key === 'type') {
```
agubler: Also missing a test (probably because the above branch isn't covered)

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,220 @@
...
+				// convert string label to object
+				if (typeof label === 'string') {
+					label = assign({}, labelDefaults, { content: label });
+				}
+				else {
+					label = assign({}, labelDefaults, label);
+				}
+
+				// add label text
+				if (label.content.length > 0) {
```
agubler: the else branch isn't covered in tests.

src/mixins/createFormLabelMixin.ts
```diff
@@ -0,0 +1,220 @@
...
+			}
+			else {
+				nodeAttributes[key.toLowerCase()] = properties[key];
+			}
+		}
+
+		return nodeAttributes;
+	},
+
+	getNode(this: FormLabel): DNode {
```
agubler: I was thinking that to do the `tag` logic it might be better if we leverage the `properties:changed` event.

It would mean a small refactor to base this mixin using `createdEvented` in order to attach to the event, similar to https://github.com/dojo/widgets/blob/master/src/mixins/themeable.ts?

In the `initialize` function there would need to be something like 

```ts
instance.own(instance.on('properties:changed', (evt: PropertiesChangeEvent<FormLabelMixin, FormLabelMixinProperties>) => {
    if (includes(evt.changedPropertyKeys, 'label')) {
        instance.tag = 'label';
    }
});

// initialise for the instantiation
if (instance.properties.label) {
    instance.tag = 'label';
}
```

It would mean widgets that use the mixin would inherit from the latest changes in widgetBase automatically and reduce the overhead of keeping the two aligned.

PR #278, opened 2017-01-23
Improve `createWidgetBase` internal instance cache

### LINKED ISSUE #276: Improve `createWidgetBase` internal instance cache
**Enhancement**

The internal widget instance cache could be improved:

1. To intelligently support multiple child widgets that use the same factory without `id`s.
2. Support the `factory` changing for child that has an `id` provided.

---

Currently for scenario 1 the first widget instance will be returned for *all* children, for example:

```ts
const createSpan = createWidgetBase.mixin({
	initialize(instance) {
		console.log(instance.properties['key']);
	}
});

const createWidget1 = createWidgetBase
	.mixin(createProjectorMixin)
	.mixin({
		mixin: {
			getChildrenNodes(): DNode[] {
				return [
					w(createSpan, { key: 'key-1' }),
					v('div', {}, [
						'text',
						w(createSpan, { key: 'key-2' }, [
							w(createSpan, { key: 'key-3' })
						]),
						v('div', {}, [
							w(createSpan, { key: 'key-4' })
						])
					]),
					w(createSpan, { key: 'key-5'  })
				];
			}
		}
	});
```

Will only log out `key-1` as once the first `createSpan` is instantiated then it is cached by the factory in the parent and all the other children use the same instance based on the factory as the key for the cache lookup.

---

Currently for scenario 2, the factory is not considered in the cache look up so if the same `id` is used but an updated `factory` a cached instance will be found using the `id` and the updated `factory` will not be used.

---

To resolve both issues we could change the caching to consider multiple values as the key for the widget instance cache.

In scenario 1 an incrementing count for each `WNode` can be used along with the `factory` as the cache look up to ensure that all the children are processed. 

In scenario 2 the `factory` can be used along with the `id` to ensure that if the `factory` changes then a new widget instance is created (with the updated `factory) and the old one destroyed.

A new `MultiMap` implementation in `dojo/core` could support both these changes https://github.com/dojo/core/pull/276

agubler:
**Type:** feature

**Description:**

As per the associated issue, support `factory` changing for a child with an `id` and intelligently cache child widgets using the same `factory` without an explicit `id`.

Resolves #276

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "peerDependencies": {
     "@dojo/compose": "2.0.0-beta.21",
-    "@dojo/core": "2.0.0-alpha.20",
+    "@dojo/core": "2.0.0-alpha.21",
     "@dojo/has": "2.0.0-alpha.7",
     "@dojo/i18n": "2.0.0-alpha.5",
     "@dojo/shim": "2.0.0-beta.8",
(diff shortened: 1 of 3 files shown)


agubler:
@dylans that was on purpose, I have only written the failing tests for the scenarios currently. Once I have finished the changes (waiting on core changes) then everything should pass.

agubler:
Closing in favour of #280

PR #293, opened 2017-01-25
Fix generics for the themeable mixin and make applied classes keys non-optional

### LINKED ISSUE #291: Typings for `theme` on themeable mixin are wrong
**Bug** 

Typings for `theme` are not correctly typed and are showing as being optional.
The `themeableFactory` in the themeable tests is also incorrectly typed.

tomdye:
**Type:** bug

**Description:**
- Make applied classes keys non-optional
- Change all generics to `<T>` (for Theme) and remove / fix incorrect generics
- Change factory generic from `<{}>` to `<any>` as we do know know until it is initialized / used what the type will be.

Resolves #291

### DIFF
--- a/src/mixins/themeable.ts
+++ b/src/mixins/themeable.ts
@@ -42,15 +42,15 @@ export interface ThemeableOptions {
 /**
  * Themeable Mixin
  */
-export interface ThemeableMixin<P> extends Evented {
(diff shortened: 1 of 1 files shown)


tomdye:
Supersedes: https://github.com/dojo/widget-core/pull/292

PR #300, opened 2017-01-27
Adding web components utilities, meta-120

rorticus:
**Type:** feature

**Description:**

First step towards custom elements. This PR contains two things,

* General purpose DOM element wrapping widget (`src/util/createDomWrapper`). Effectively this widget will just append a DOM element to itself..
* Methods to decorate a custom HTMLElement and tie it to a widget. The initializer will turn the HTML element into a projector with the widget as a child (see README for details).

What this PR does *not* contain is anything to perform the actual custom element registration. That will be somewhere....else.

Resolves: https://github.com/dojo/meta/issues/120

### DIFF
--- a/src/customElements.ts
+++ b/src/customElements.ts
@@ -1,0 +1,254 @@
+import { w } from './d';
+import { WidgetProperties, WidgetFactory, Widget, DNode } from './interfaces';
+import createProjectorMixin from './mixins/createProjectorMixin';
+import createDomWrapper from './util/createDomWrapper';
+import { assign } from '@dojo/core/lang';
+import { Projector } from './mixins/createProjectorMixin';
+import { from as arrayFrom } from '@dojo/shim/array';
+/**
+ * @type CustomElementAttributeDescriptor
+ *
+ * Describes a custom element attribute
+ *
+ * @property {string}       attributeName   The name of the attribute on the DOM element
+ * @property {string?}      propertyName    The name of the property on the widget
+ * @property {Function?}    value           A function that takes a string or null value, and returns a new value. The widget's property will be set to the new value.
+ */
+export interface CustomElementAttributeDescriptor {
+	attributeName: string;
+	propertyName?: string;
+	value?: (value: string | null) => any;
+}
+
+/**
+ * @type CustomElementPropertyDescriptor
+ *
+ * Describes a widget property exposed via a custom element
+ *
+ * @property {string}       propertyName        The name of the property on the DOM element
+ * @property {string?}      widgetPropertyName  The name of the property on the widget
+ * @property {Function?}    getValue            A transformation function on the widget's property value
+ * @property {Function?}    setValue            A transformation function on the DOM elements property value
+ */
+export interface CustomElementPropertyDescriptor {
+	propertyName: string;
+	widgetPropertyName?: string;
+	getValue?: (value: any) => any;
+	setValue?: (value: any) => any;
+}
+
+/**
+ * @type CustomElementEventDescriptor
+ *
+ * Describes a custom element event
+ *
+ * @property    {string}    propertyName    The name of the property on the widget that takes a function
+ * @property    {string}    eventName       The type of the event to emit (it will be a CustomEvent object of this type)
+ */
+export interface CustomElementEventDescriptor {
+	propertyName: string;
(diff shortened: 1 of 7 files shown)


src/customElements.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	if (initialization) {
+		initialization.call(element, initialProperties);
+	}
+
+	Array.prototype.slice.call(element.children, 0).forEach((childNode: HTMLElement) => {
+		element.removeChild(childNode);
+	});
+
+	const widgetInstance = element.getWidgetFactory().mixin(createProjectorMixin)({
```
agubler: Can you tell it here that it is now a Projector? and avoid the cast to `any` below?

src/customElements.ts
```diff
@@ -0,0 +1,198 @@
...
+/**
+ * Called by HTMLElement subclass when an HTML attribute has changed.
+ *
+ * @param element
+ * @param name
+ * @param newValue
+ * @param oldValue
+ */
+export function handleAttributeChanged(element: CustomElement, name: string, newValue: string | null, oldValue: string | null) {
+	(element.getDescriptor().attributes || []).forEach((attribute) => {
```
agubler: perhaps would be cleaner if you did, but it's just a nit pick

```ts
const attributes = element.getDescriptor().attributes || [];
attributes.forEach...
```

src/util/createDomWrapper.ts
```diff
@@ -0,0 +1,72 @@
...
+					});
+
+					handleDomInsertion(this, this.properties.domNode);
+				}
+			}
+		},
+		afterUpdate(this: Widget<DomWrapperProperties>) {
+			handleDomInsertion(this, this.properties.domNode);
+		},
+		nodeAttributes: [
```
agubler: nodeAttributes will go away soon :), It will become an aspect after `render` to add the properties to the outer node

src/customElements.ts
```diff
@@ -0,0 +1,198 @@
...
+				bubbles: false,
+				detail: event
+			}));
+		};
+	});
+
+	// find children
+	let children: DNode[] = [];
+
+	Array.prototype.slice.call(element.children, 0).forEach((childNode: HTMLElement, index: number) => {
```
agubler: could this be replaced with a spread i.e. `[...element.children]`?

rorticus: I get a TS error, 'HTMLCollection is not an array type'

agubler: Ah nice (worked in the dev tools) but obviously typed as not an array.

src/util/createDomWrapper.ts
```diff
@@ -0,0 +1,72 @@
...
+		data.vNode.domNode = notNullNode;
+	}
+}
+
+const createDomWrapper: DomWrapperFactory = createWidget.mixin({
+	mixin: {
+		afterCreate(this: Widget<DomWrapperProperties>) {
+			if (this.properties.domNode) {
+				// render has already been called at this point, so our vnode should be cached
+				const vNode = this.__render__();
```
matt-gadd: would it be nicer to aspect after the `__render__` rather than calling it again to get the vnode? it would work fine at the moment, but i worry in the future we might add something to `__render__` that might cause side effects by calling it again out of the render cycle

src/customElements.ts
agubler: Can you use `Array.from`? which is shim'd in `@dojo/shim/array`?

```ts
arrayFrom(element.children).forEach(...);
```

src/customElements.ts
rorticus: Works! :)

src/customElements.ts
rorticus: Separated!

src/util/createDomWrapper.ts
rorticus: Yeah... when your PR is merged I will update this

src/customElements.ts
rorticus: Converted to use `@dojo/shim/array`'s `from`

src/util/createDomWrapper.ts
rorticus: Good idea. That cleaned up the code a bit.

PR #303, opened 2017-01-28
Generic DNode decorator util

### LINKED ISSUE #302: Provide generic decorator function for DNodes
**Enhancement**

In order to support functionally decorating `DNode`s using a modifier based on a predicate, provide a generic decorate function that can be used leveraged by mixins primarily in an after aspect of `getNode` / `getChildrenNodes` or soon `render`

agubler:
**Type:** feature

**Description:**

Generic decorator for DNodes.

Resolves #302

### DIFF
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -120,7 +120,7 @@ function dNodeToVNode(instance: Widget<WidgetProperties>, dNode: DNode): VNode |
 		return child.__render__();
 	}
 
-	dNode.vnodes = dNode.children
+	dNode.vNodes = dNode.children
 		.filter((child) => child !== null)
 		.map((child: DNode) => {
 			return dNodeToVNode(instance, child);
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -120,7 +120,7 @@ function dNodeToVNode(instance: Widget<WidgetProperties>, dNode: DNode): VNode |
 		return child.__render__();
 	}
 
-	dNode.children = dNode.children
+	dNode.vnodes = dNode.children
 		.filter((child) => child !== null)
 		.map((child: DNode) => {
 			return dNodeToVNode(instance, child);
--- a/src/d.ts
+++ b/src/d.ts
@@ -37,7 +37,7 @@ export function isHNode(child: DNode): child is HNode {
 }
 
 /**
- * Generic decorate function for DNodes, the modified in place based on the provided predicate
+ * Generic decorate function for DNodes, the nodes are modified in place based on the provided predicate
  * and modifier functions.
  *
  * If no predicate is supplied then the modifier will be executed on all nodes.
--- a/src/d.ts
+++ b/src/d.ts
@@ -37,8 +37,10 @@ export function isHNode(child: DNode): child is HNode {
 }
 
 /**
- * Generic decorate function for DNodes, the nodes are modified in place based on the provided predicate
+ * Generic decorate function for DNodes. The nodes are modified in place based on the provided predicate
  * and modifier functions.
+ *
+ * The children of each node are flattened and added to the array for decoration.
  *
  * If no predicate is supplied then the modifier will be executed on all nodes.
  */
--- a/src/d.ts
+++ b/src/d.ts
@@ -86,8 +86,8 @@ export function v(tag: string, propertiesOrChildren: VNodeProperties = {}, child
 		return {
 			children,
 			properties,
-			render<T>(this: { vnodes: VNode[], properties: VNodeProperties }, options: { bind?: T } = { }) {
-				return h(tag, assign(options, this.properties), this.vnodes);
+			render<T>(this: { vNodes: VNode[], properties: VNodeProperties }, options: { bind?: T } = { }) {
+				return h(tag, assign(options, this.properties), this.vNodes);
 			},
 			type: HNODE
 		};
--- a/src/d.ts
+++ b/src/d.ts
@@ -36,6 +36,28 @@ export function isHNode(child: DNode): child is HNode {
 	return Boolean(child && (typeof child !== 'string') && child.type === HNODE);
 }
 
+/**
+ * Generic decorate function for DNodes, the modified in place based on the provided predicate
+ * and modifier functions.
+ *
+ * If no predicate is supplied then the modifier will be executed on all nodes.
+ */
+export function decorate(dNodes: DNode | DNode[], modifier: (dNode: DNode) => void, predicate?: (dNode: DNode) => void): DNode | DNode[] {
+	let nodes = Array.isArray(dNodes) ? [ ...dNodes ] : [ dNodes ];
+	while (nodes.length) {
+		const node = nodes.pop();
+		if (node) {
+			if (!predicate || predicate(node)) {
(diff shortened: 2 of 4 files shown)


src/d.ts
```diff
@@ -36,6 +36,28 @@ export function isHNode(child: DNode): child is HNode {
 	return Boolean(child && (typeof child !== 'string') && child.type === HNODE);
 }
 
+/**
+ * Generic decorate function for DNodes, the modified in place based on the provided predicate
```
dylans: grammar reads a bit oddly here.

src/d.ts
```diff
@@ -36,6 +36,28 @@ export function isHNode(child: DNode): child is HNode {
...
+ * and modifier functions.
+ *
+ * If no predicate is supplied then the modifier will be executed on all nodes.
+ */
+export function decorate(dNodes: DNode | DNode[], modifier: (dNode: DNode) => void, predicate?: (dNode: DNode) => void): DNode | DNode[] {
+	let nodes = Array.isArray(dNodes) ? [ ...dNodes ] : [ dNodes ];
```
dylans: As I ask every time I see an isArray check, should this API also support an iterable?

src/d.ts
```diff
@@ -36,6 +36,28 @@ export function isHNode(child: DNode): child is HNode {
...
+	while (nodes.length) {
+		const node = nodes.pop();
+		if (node) {
+			if (!predicate || predicate(node)) {
```
dylans: would probably help to add a comment explaining this if statement and then next one so we remember in a year what they do.

src/interfaces.d.ts
```diff
@@ -81,17 +81,27 @@ export interface FactoryRegistryInterface {
...
-	children: (DNode | (VNode | string))[];
+	children: (DNode | string)[];
 
 	/**
 	 * render function that wraps returns VNode
 	 */
 	render<T>(options?: { bind?: T }): VNode;
 
 	/**
+	 * The properties used to create the vnode
```
dylans: VNode (for consistency)

tests/unit/d.ts
```diff
@@ -92,7 +108,102 @@ registerSuite({
...
+				assert.isUndefined(children![1].properties['decorated']);
+				assert.isUndefined(children![2].properties['decorated']);
+				assert.isUndefined(children![3].properties['decorated']);
+				assert.strictEqual(children![4], 'my text');
+				assert.isNull(children![5]);
+				assert.isUndefined(children![6].properties['decorated']);
+				assert.isUndefined(children![7].properties['decorated']);
+			}
+		},
+		'applies modifier to all row when no predicate supplied'() {
```
dylans: all row(s) ?

src/d.ts
agubler: I don't this so as the nodes will primarily be used straight from `getNode`, `getChildrenNodes` or `render` all of which return an array or single node.

src/d.ts
agubler: What do you mean? If there is no predicate or the predicate returns true for the node then the modifier runs for the node.

I'd rather clean the code to make it so that it is clear than add an inline comment, do you have any suggestions? What makes the `if` unclear?

src/d.ts
dylans: ok, thanks!

src/d.ts
dylans: It just really wasn't obvious to me at all what this is doing or why. I mean, I understand the words decorate, predicate, and modifier, but I guess I'm lacking context for how this is used. Also it wasn't clear to me at a first glance, if something is either not a predicate or a predicate call with node as a parameter, why these are the conditions to call modifier on the node. And for the statement below, basically if something is a WNode or HNode and it has children, then a list of nodes is an array that is spread from a list of nodes and children? So basically you're flattening the loop of nodes to iterate over? And you're modifying the list of nodes within the while loop? I find it confusing to change the length of a while loop while iterating over the while loop, but that may just be me (there's no side effects as you're not adding to the list of nodes before the one you're currently iterating over obviously). :) Anyway, I'm probably just being too difficult on this particular block of code, but those were my thoughts when reviewing it.

matt-gadd: The actual loop code originates from my original pr: https://github.com/dojo/widget-core/pull/270. The reason it's a while loop is I unrolled it from it being recursive. For the rest of the code I think the jsdoc comment is clear on how the functionality behaves:

```
* Generic decorator function for DNodes, the nodes are modified in place based on the provided predicate
 * and modifier functions.
 *
 * If no predicate is supplied then the modifier will be executed on all nodes.
```

agubler: Yes flattening the nodes in order to loop through them checking the predicate and if required running the modifier.

I think the code itself is actually pretty clear especially used in context of the jsdoc function description.

Perhaps should add something to the functions jsdoc describing the behaviour with children, that the decorator will traverse the tree if one is provided?

matt-gadd: Yeah perhaps indicating that it will decorate a tree of DNodes in the jsdoc would be useful

tests/unit/d.ts
```diff
@@ -92,7 +108,102 @@ registerSuite({
...
+				assert.isUndefined(children![1].properties['decorated']);
+				assert.isUndefined(children![2].properties['decorated']);
+				assert.isUndefined(children![3].properties['decorated']);
+				assert.strictEqual(children![4], 'my text');
+				assert.isNull(children![5]);
+				assert.isUndefined(children![6].properties['decorated']);
+				assert.isUndefined(children![7].properties['decorated']);
+			}
+		},
+		'applies modifier to all rows when no predicate supplied'() {
```
matt-gadd: shouldn't this be `nodes` not `rows`?

src/createWidgetBase.ts
```diff
@@ -120,7 +120,7 @@ function dNodeToVNode(instance: Widget<WidgetProperties>, dNode: DNode): VNode |
 		return child.__render__();
 	}
 
-	dNode.children = dNode.children
+	dNode.vnodes = dNode.children
```
matt-gadd: do you think we should camelcase this? `vNodes`?

src/d.ts
dylans: "The nodes are modified in place" is a bit clearer that "the modified in place". I guess I just didn't know what the intent of the code is (decorators in the context of higher order functions I assume?)

So basically the intent to to traverse a tree of nodes, flatten them, and apply a modifier and/or predicate to each node?

agubler: The general concept is that the decorator accepts a single or array of `DNode`s, a `modifier` function and optionally a `predicate`.

It loops through each of the DNodes, checking the `predicate` by passing the current node and evaluating the return value (`true`/`false`) and if `true`, runs the node through the `modifier`. If there is no `predicate` provided then the `modifier` is run for all nodes.

As the decorator iterates through the nodes it spreads the children of each onto the node array in order to process them through the decorator, essentially flatten the tree structure of DNodes for decoration.

The original DNode argument (single or array) is returned (having had the modifications applied).

dylans: Ok. I guess the name predicate was throwing me off... in Dojo 1.x and in other APIs I'm familiar with, it would have been called filter probably. Either way, this is fine.

PR #307, opened 2017-01-30
Remove getNodeAttributes/nodeAttributes API

### LINKED ISSUE #306: Remove support for `getNodeAttributes` reducer
**Enhancement**

Given the new [DNode decorator utility](https://github.com/dojo/widget-core/pull/303) and the move to a [single `render` function](https://github.com/dojo/widget-core/pull/305) for authoring widgets, the util should be used for decorating nodes with all additional properties (including adding/modifying VNodeAttributes) - This can be done either by providing a mixin that explicitly just decorates nodes for a given requirement or used within more complex mixins to add the desired modifications needed to the widget's nodes to support the mixins functionality.

The decorator is not limited to the top level node of a widget and depending on the predicate can target any/all of the nodes within the widgets structure.

This change is **breaking**, consumers will need to change existing `nodeAttribute` functions to use the decorator util in an after aspect of the `render` function.

Lastly, the default `classes`, `styles`, and widget `id` nodeAttributes functionality will also be removed from `createWidgetBase` and replaced by a mixin that will need to be explicitly mixed in when the functionality is required.

agubler:
**Type:** feature

**Description:**

As per associated issue.

Resolves #306


src/mixins/createFormLabelMixin.ts
```diff
@@ -125,92 +115,84 @@ const labelDefaults = {
...
-	},
-
-	initialize(instance: any) {
-		instance.own(instance.on('properties:changed', (evt: PropertiesChangeEvent<FormLabelMixin, FormLabelMixinProperties>) => {
-			instance.tagName = evt.properties.label ? 'label' : 'div';
-		}));
-
-		if (instance.properties && instance.properties.label) {
-			instance.tagName = 'label';
+		else if (key === 'type') {
```
tomdye: could this be a `switch` ?

agubler: Potentially it could be but as you observed, I am not changing the functionality of the mixin. Only updating inline with the latest widget changes.

agubler: follow up, I am not sure a `switch` would be correct given the logic in the `if else` conditions.

src/createWidgetBase.ts
```diff
@@ -160,11 +151,8 @@ const createWidget: WidgetBaseFactory = createEvented
 				return properties;
 			},
 
-			classes: [],
-
 			getNode(this: Widget<WidgetProperties>): DNode {
-				const tag = formatTagNameAndClasses(this.tagName, this.classes);
-				return v(tag, this.getNodeAttributes(), this.getChildrenNodes());
+				return v('div', {}, this.getChildrenNodes());
```
dylans: By switching to a default of `div`, don't we essentially make this unusable for dataviz where we may want to create widgets as SVG? Or would a div always wrap an svg tag and then an svg dom?

agubler: My working assumption is that the dataviz mixin would just wrap any dataviz component with an `svg` node (essentially doing the same as what the mixin does now by setting the tagName to `svg`)

agubler:
@dylans what new assumptions are we making as a result of these changes? The recommended pattern will be to implement `render` now rather than `getNode` or `getChildrenNode` (they are still default for backwards compatibility) and in the render the author as full control on the node structure of the widget.

PR #312, opened 2017-01-31
Implement bind for widget properties

matt-gadd:
**Type:** feature

**Description:**

Resolves https://github.com/dojo/widget-core/issues/308

### DIFF
--- a/src/createWidgetBase.ts
+++ b/src/createWidgetBase.ts
@@ -42,6 +42,8 @@ interface WidgetInternalState {
  */
 const widgetInternalStateMap = new WeakMap<Widget<WidgetProperties>, WidgetInternalState>();
 
+const bindFunctionPropertyMap = new WeakMap<(...args: any[]) => any, { boundFunc: (...args: any[]) => any, scope: any }>();
+
 const propertyFunctionNameRegex = /^diffProperty(.*)/;
 
 function getFromRegistry(instance: Widget<WidgetProperties>, factoryLabel: string): FactoryRegistryItem | null {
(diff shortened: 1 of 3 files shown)


src/createWidgetBase.ts
```diff
@@ -143,6 +149,24 @@ function manageDetachedChildren(instance: Widget<WidgetProperties>): void {
 	});
 }
 
+function bindFunctionProperties(instance: Widget<WidgetProperties>, properties: WidgetProperties) {
+	Object.keys(properties).forEach((propertyKey) => {
+		const func = properties[propertyKey];
```
dylans: func probably violates our naming conventions, but I'll look the other way a I don't see a better option here.

src/createWidgetBase.ts
```diff
@@ -143,6 +149,24 @@ function manageDetachedChildren(instance: Widget<WidgetProperties>): void {
...
+		const bind = properties.bind;
+
+		if (typeof func === 'function') {
```
dylans: I guess I feel like some of this belongs in dojo/core probably, so that we're not reimplementing binding throughout Dojo 2, though I realize this is pretty specific.

matt-gadd: What am I re-implementing? it's one line of code thats part of the language?

src/createWidgetBase.ts
matt-gadd: happy to take a suggestion on a more appropriate name.

src/createWidgetBase.ts
matt-gadd: or are you talking of the block? if so its specific to properties in `widget-core`, so wouldn't be applicable in `dojo/core`

dylans: I guess the logic of (if it's a function do this, if it's a string do that)... that said, I realize it's pretty specific to what you're doing here, so just ignore me. :)

src/createWidgetBase.ts
dylans: function is the better name, but it's taken/reserved. :)

PR #313, opened 2017-01-31
Adding utility to register a custom element

rorticus:
**Type:** feature

**Description:**

This is another chunk of custom elements code. The code in this PR will actually register a custom element with the browser. The intent is that the actual method call will come from generated code that the tooling.

There are no tests for this as I don't really know how to test this since it's doing things specific to DOM element creation...


src/registerCustomElement.ts
```diff
@@ -0,0 +1,115 @@
...
+		}
+	});
+}
+
+/**
+ * Register a custom element using the v0 spec of custom elements.
+ *
+ * @param descriptorFactory
+ */
+export function registerCustomElementV0(descriptorFactory: CustomElementDescriptorFactory) {
```
matt-gadd: I think we should ditch the V0 support, as far as i'm aware only chrome implemented it and so it doesn't give us any more support.

rorticus: Can do. technically it works in IE 10 while the V1 support doesn't. But if we're not interested in that, it's outta here!

rorticus: Crushed v0 support with an iron fist.

src/registerCustomElement.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Register a custom element using the v1 spec of custom elements. Note that
+ * this is the default export, and, expects the proposal to work in the browser.
+ * This will likely require the polyfill and native shim.
+ *
+ * @param descriptorFactory
+ */
+export function registerCustomElementV1(descriptorFactory: CustomElementDescriptorFactory) {
```
agubler: Do we need the `V1` suffix?

rorticus: Good point! I guess we don't need that anymore...

src/registerCustomElement.ts
```diff
@@ -0,0 +1,59 @@
...
+		getWidgetFactory(): WidgetFactory<any, any> {
+			return this.getDescriptor().widgetFactory;
+		}
+
+		getDescriptor(): CustomElementDescriptor {
+			return descriptor;
+		}
+
+		static get observedAttributes(): string[] {
+			return (descriptor.attributes || []).map(attribute => attribute.attributeName);
```
agubler: there is a missing test branch here.

src/customElements.ts
```diff
@@ -129,6 +131,23 @@ function getWidgetPropertyFromAttribute(attributeName: string, attributeValue: s
 	return [ propertyName, value ];
 }
 
+let customEventClass = global.CustomEvent;
+
+if (typeof customEventClass !== 'function') {
+	const customEvent = function (event: string, params: any) {
+		params = params || { bubbles: false, cancelable: false, detail: undefined };
+		let evt = document.createEvent('CustomEvent');
```
agubler: can this be a `const`

src/registerCustomElement.ts
rorticus: picky picky :)

src/customElements.ts
rorticus: yup

src/registerCustomElement.ts
agubler: That's the difference between codecov checks passing and failing 😄

rorticus: Haha! Zing!

PR #316, opened 2017-02-02
Changed BaseClassses shape and key to _key

tomdye:
**Type:** bug

**Description:**

BaseClasses is now a flat object again such that we can use the `typed-css-modules` cli and don't have to use `baseTheme.classes.className` and instead can use `baseTheme.className`.

### DIFF
--- a/src/mixins/themeable.ts
+++ b/src/mixins/themeable.ts
@@ -74,7 +74,7 @@ export interface ThemeableMixin extends Evented {
  * Themeable
  */
 export interface Themeable extends ThemeableMixin {
-	baseClasses: BaseClasses;
+	baseClasses: {};
 	properties: ThemeableProperties;
 }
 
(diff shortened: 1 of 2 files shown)


src/mixins/themeable.ts
```diff
@@ -74,25 +74,16 @@ export interface ThemeableMixin extends Evented {
  * Themeable
  */
 export interface Themeable extends ThemeableMixin {
-	baseClasses: BaseClasses;
+	baseClasses: {};
```
agubler: is this not `BaseClasses`?

tomdye: Nope, if I make it `BaseClasses` you lose the typings on the css file you import because you have to declare it as `BaseClasses` which is basically just a string index object used within this file to make TS happy.

agubler: Okay so I just tried this with your latest changes and using BaseClasses as the type and it worked, with intellisense ¯\_(ツ)_/¯

tomdye: It didn't work when I used it in my test code. Did you have to case your required `css` as `BaseClasses` ?

agubler: Nope I just did `import * as css from 'my/css/file';`

tomdye: It does not work for me when i create a custom widget and try to set it's imported CSS as `baseClasses`. I get the message:

PR #319, opened 2017-02-03
FormLabelMixin: pass classes and label-specific attributes to the <label> node

### LINKED ISSUE #317: FormLabelMixin needs to set some properties on <label>
**Bug **

If a label property is set, the root node of the form widget (the <label> element) cannot receive any node attributes.

**Expected behavior:**
In this scenario, `value` and `name` should be set on the form field, while `classes` and `form` should go on the `<label>` node:

```
formField.setProperties({
	value: 'bar',
	name: 'baz',
	classes: classesObj,
	form: 'formID'
});
```

smhigley:
**Type:** bug

Resolves #317

### DIFF
--- a/src/mixins/createFormLabelMixin.ts
+++ b/src/mixins/createFormLabelMixin.ts
@@ -58,6 +58,11 @@ export interface FormLabelMixinProperties {
 	describedBy?: string;
 
 	/**
+	 * ID of a form element associated with the form field
+	 */
+	form?: string;
+
+	/**
 	 * Indicates the value entered in the form field is invalid
 	 */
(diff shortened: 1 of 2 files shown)


src/mixins/createFormLabelMixin.ts
```diff
@@ -58,6 +58,11 @@ export interface FormLabelMixinProperties {
 	describedBy?: string;
 
 	/**
+	 * ID of a form element associated with the form field
+	 */
+	form?: string;
```
agubler: this naming confused me, even more so once I read the jsdoc - should probably include `id` in the name.

PR #322, opened 2017-02-03
Moving properties into initialization

rorticus:
**Type:** bug

**Description:**

Moving properties into widget construction of custom element, instead of afterwards.


rorticus:
Really, codecov. 0.01%?

PR #324, opened 2017-02-06
Allowing afterCreate to be called on the projector, issue #323

### LINKED ISSUE #323: createProjectorMixin should not overwrite the afterCreate property
**Bug**

**Code**
```typescript
function afterCreate() {
	// never called
}
const createWidget = createWidgetBase
	.mixin(createProjectorMixin)
	.override({
		render() {
			return v('span', {
				innerHTML: 'hello world',
				afterCreate
			});
		}
	});
```

**Expected behavior:**

My `afterCreate` function should be called.

**Actual behavior:**

My `afterCreate` function is not called.

rorticus:
**Type:** bug

**Description:**

This will allow you to override `afterCreate` on the projector and have both your method, and the projector's method, called.  This is sort of brute-forced in there, as I couldn't think of a clever way to use aspects for something like this, since the method itself is dynamic.

Resolves #323

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -1,8 +1,9 @@
+import { EventTargettedObject, Handle } from '@dojo/interfaces/core';
 import Promise from '@dojo/shim/Promise';
 import { createProjector as createMaquetteProjector, Projector as MaquetteProjector } from 'maquette';
 import { EventTargettedObject, Handle } from '@dojo/interfaces/core';
+import { WidgetProperties } from './../WidgetBase';
 import { Constructor, WidgetProperties } from './../interfaces';
-import { WidgetBase } from './../WidgetBase';
 import cssTransitions from '../animations/cssTransitions';
 
 /**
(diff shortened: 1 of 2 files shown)


src/mixins/createProjectorMixin.ts
```diff
@@ -142,6 +143,10 @@ function attach(instance: Projector, { type }: AttachOptions) {
 
 	projectorData.attachPromise = new Promise((resolve, reject) => {
 		projectorData.afterCreate = () => {
+			if (projectorData.originalAfterCreate) {
+				projectorData.originalAfterCreate.call(instance);
```
matt-gadd: does this not need the arguments?

rorticus: No you're right, I should pass those in just in case the original call needs them. Good catch!

matt-gadd: also should the scope be on the instance? the scope should be whatever is bound right? so you should be able to just call it and pass the args as a rest arg?

matt-gadd: in fact, it should perhaps be called on the scope of whatever the `bind` property passed to maquette is I think?

PR #326, opened 2017-02-07
Convert to use TS2.2 Class Mixins over dojo/compose

### LINKED ISSUE #327: Convert from dojo/compose to TS2.2 class mixins
**Enhancement**

With the impending support for [class mixins](https://github.com/Microsoft/TypeScript/pull/13743) in TS2.2 (currently @ `rc`), we should be able to leverage these native class mixins to replace `dojo/compose`.

This should be more familiar for consumers and provide a lower barrier of entry, enables widget-core to evolve with the latest/next enhancements in Typescript.

A (nice) side affect is that widgets should actually become cleaner and easier to type than they currently are using `dojo/compose`.

**Code**

example of a class mixin:

```ts
export function RegistryMixin<T extends WidgetConstructor>(base: T): T {
	return class extends base {
		properties: RegistryMixinProperties;

		constructor(...args: any[]) {
			super(...args);
			this.own(this.on('properties:changed', (evt: PropertiesChangeEvent<this, RegistryMixinProperties>) => {
				if (includes(evt.changedPropertyKeys, 'registry')) {
					this.registry = evt.properties.registry;
				}
			}));
			const { properties: { registry } } = this;
			if (registry) {
				this.registry = registry;
			}
		}

		public diffPropertyRegistry(previousValue: FactoryRegistry, value: FactoryRegistry): PropertyChangeRecord {
			return {
				changed: previousValue !== value,
				value: value
			};
		}
	};
}
```

example or widget authoring with mixin usage

```ts
export class Button extends I18nMixin(StatefulMixin(Themeable(WidgetBase))) {

    properties: ButtonProperties;

    constructor(options: any) {
        options.baseClasses = css;
        super(options);
    }

    onClick() {
        let { state: { counter = 0 } }  = this;
        counter++;
        this.setState({ counter });
    }

    render() {
        const { state: { counter = 0 } }  = this;
        const messages = this.localizeBundle(bundle);

        return v('div', [
            v('button', { type: 'button', onclick: this.onClick, innerHTML: messages.click, classes: this.classes(css.myClass).get() }),
            v('span', { classes: this.classes(css.myClass).get() }, [ String(counter) ])
        ]);
    }
}
```

agubler:
**Type:** feature

**Description:**

Convert widget-core to use TS2.2 class mixins instead of `dojo/compose` 

There are a couple of interesting changes included

1) The `baseTheme` property for the Themeable mixin is set via a class decorator (`@theme` exported from Themeable)
2) The “result” from using the class Mixins requires a generic to define the Widget’s properties `class MyWidget extends ThemeableMixin(WidgetBase)<MyWidgetProperties>` and excluding this does actually provide a pretty obscure error `No base constructor has the specified number of type arguments.` along with a few others.
3) I’ve removed the default export for all base, mixin and main classes in place of named exports (This is personal preference, so we can debate this potentially)
4) The constructor argument for any widget is now *just properties*, the WidgetOptions class/interface has been removed.
5) Allowed pascal casing for variables, mainly because the result can be a class in cases like `const Projector = ProjectorMixin(Button);`

Items to follow up on:

1) A generic solution for decorating the prototype with attributes, eg. provide an object bag from widget base that attributes get set to and then copied from for usage.
2) Decorators to add functions to a list used by widget base for "property specific diffs" and "render decorators". This will enable us to move away from strict regular expressions.
3) Need to check the order that "decorator" functions get applied, should work back up the inheritance chain.
4) Type doc corrected/added where appropriate

Resolves #327


mwistrand:
The i18n-specific changes look good. 👍

rorticus:
Custom element stuff looks good 👍

src/bases/Destroyable.ts
```diff
@@ -0,0 +1,63 @@
...
+	 * @constructor
+	 */
+	constructor() {
+		this.handles = [];
+	}
+
+	/**
+	 * Register handles for the instance that will be destroyed when `this.destroy` is called
+	 *
+	 * @param {Handle} handle The handle to add for the instance
```
rishson: Nit: remove types from comments as per [guidelines](https://github.com/dojo/meta/blob/master/STYLE.md#inline-documentation)

src/mixins/ProjectorMixin.ts
```diff
@@ -0,0 +1,207 @@
...
+	 */
+	root: Element;
+
+	/**
+	 * The status of the projector
+	 */
+	readonly projectorState: ProjectorState;
+}
+
+export function ProjectorMixin<T extends WidgetConstructor>(base: T): T & Constructor<Projector> {
```
rorticus: Any reason this isn't the default export? Same goes for WidgetBase I think.

src/bases/Destroyable.ts
agubler: Add a note that we need to follow up on type doc throughout the repo.

src/mixins/ProjectorMixin.ts
agubler: I have done this for all classes, it's a topic for discussion i guess but I you'll rarely just use the default.

src/WidgetBase.ts
```diff
@@ -0,0 +1,525 @@
...
+ */
+export interface PropertiesChangeRecord<P extends WidgetProperties> {
+	changedKeys: string[];
+	properties: P;
+}
+
+/**
+ * WidgetBase constructor type
+ */
+export type WidgetConstructor = new (...args: any[]) => WidgetBase<WidgetProperties>;
```
maier49: Would it make sense to write this as
```typescript
export type WidgetConstructor = Constructor<WidgetBase<WidgetProperties>>;
```
?

agubler: I think that absolutely makes sense @maier49, do you mind if we fix the up in a follow up PR?

PR #329, opened 2017-02-08
Adding custom CSS transition module, issue #328

### LINKED ISSUE #328: Support custom transition active/exit classes
**Enhancement** 

Maquette is putting a `-active` suffix on our class names during CSS transitions. Because CSS modules obfuscate our class names, we are unable to actually target those classes.

See the great write-up by @bitpshr in the original issue, https://github.com/AFASSoftware/maquette/issues/87 , for details.

We need our own transition module to support CSS transitions with our CSS modules.

rorticus:
**Type:** feature

**Description:**

Our own CSS transition module. This allows us to specify `enterAnimationActive` and `exitAnimationActive` class names on nodes, to work with CSS modules.

Resolves #328


src/animations/cssTransitions.ts
```diff
@@ -0,0 +1,83 @@
...
+	enterAnimationActive?: string;
+	exitAnimationActive?: string;
+}
+
+function determineBrowserStyleNames(element: HTMLElement) {
+	if ('WebkitTransition' in element.style) {
+		browserSpecificTransitionEndEventName = 'webkitTransitionEnd';
+		browserSpecificAnimationEndEventName = 'webkitAnimationEnd';
+	}
+	else if ('transition' in element.style) {
```
dylans: any reason to not make this be `else if ('transition' in element.style || 'MozTransition' in element.style)` given that the body of each of these if blocks is the same? If the reason is to eventually remove the mozilla specific version, then perhaps at least the if ('transition' in block should come first since that's the standard, or does WebKit evaluate to true for both but need the webkit version?

rorticus: That makes sense 👍 This was from the original maquette code and I guess I didn't optimize it. Fixed!

src/animations/cssTransitions.ts
```diff
@@ -0,0 +1,79 @@
...
+		}
+	};
+
+	startAnimation();
+
+	element.addEventListener(browserSpecificAnimationEndEventName, transitionEnd);
+	element.addEventListener(browserSpecificTransitionEndEventName, transitionEnd);
+}
+
+export namespace cssTransitions {
```
tomdye: Is there a preference to using `namespace` in this instance over 

```
function exit() ...
function enter()  ...

export default {
   enter,
   exit
}
```

tomdye: As discussed, I believe this should now be changed to export an object.

rorticus: Fixed 👍

tests/functional/cssTranstions.ts
```diff
@@ -0,0 +1,62 @@
+import * as registerSuite from 'intern!object';
+import * as assert from 'intern/chai!assert';
+import pollUntil = require('intern/dojo/node!leadfoot/helpers/pollUntil');
+
+registerSuite({
+	name: 'cssTransitions',
+
+	'-active gets appended to enter/exit animations by default'(this: any) {
```
matt-gadd: do you think these could just be unit tests?

rorticus: Maybe? Can we unit test changes to DOM elements? I see we have JSDOM loaded in widget-core, but I'm not really sure what I can do with that.

PR #334, opened 2017-02-09
Return of the default exports

agubler:
**Type:** discussion

**Description:**

Add module default exports where sensible.

### DIFF
--- a/src/mixins/Themeable.ts
+++ b/src/mixins/Themeable.ts
@@ -239,3 +239,5 @@ export function ThemeableMixin<T extends Constructor<WidgetBase<WidgetProperties
 		}
 	};
 }
+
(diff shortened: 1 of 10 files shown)


src/mixins/Themeable.ts
```diff
@@ -237,3 +237,5 @@ export function ThemeableMixin<T extends WidgetConstructor>(base: T): Constructo
 		}
 	};
 }
+
+export default ThemeablMixin;
```
matt-gadd: spelling here?

agubler: Ha that's a crappy (a bug in the existing code that I have spelt the interface incorrectly so it let me export that instead of the function!)

PR #350, opened 2017-02-12
Fix up mixin interfaces

agubler:
**Type:** bug

**Description:**

Interfaces can be the same name as the function for mixins provided they are both exported from the same module. Also the Base class needs to be specified first for the return intersection type.

### DIFF
--- a/src/mixins/Stateful.ts
+++ b/src/mixins/Stateful.ts
@@ -12,7 +12,7 @@ export interface State {
 /**
  * Stateful interface
  */
-export interface Statful {
+export interface StatfulMixin {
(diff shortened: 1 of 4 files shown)


src/mixins/Stateful.ts
```diff
@@ -12,7 +12,7 @@ export interface State {
 /**
  * Stateful interface
  */
-export interface Statful {
```
dylans: Stateful not Statful, see below too

agubler: Thanks, had always been wrong.

PR #355, opened 2017-02-13
Using Evented and Destroyable from dojo/core, core #286

rorticus:
**Type:** feature

**Description:**

After https://github.com/dojo/core/pull/287 is merged, `Evented` and `Destroyable` will be available in dojo/core, so this PR uses them!

Resolves core 286


agubler:
Do you need to bump the core version?

rorticus:
@agubler Bumped!  It still isn't going to build yet because I bumped the i18n version too and that isn't released yet.

PR #357, opened 2017-02-14
Decouple initial set properties from the widget base constructor

### LINKED ISSUE #356: Initial 'properties:changed' event fires too late
**Enhancement** 

In order to provide support for decorator functionality in non decorator environments and prevent the need to manually call "set up" functions from constructors to deal with the initial properties, we need to address the point in the lifecycle where properties initially get set.

Currently this is in the constructor of widget base which means that the event has already fired before the sub classes' functions have been registered.

Proposal is to stop setting the properties during construction and call `setProperties` as a second action, in most cases this will still be internal to widget base in `dNodeToVNode` with the exception of when dealing with a projector widget.

agubler:
**Type:** feature

**Description:**

As described by the issue, with the initial setProperties occurring in WidgetBase constructor it means that the first event is missed by sub classes and inadvertently promotes a pattern of putting logic in the constructor. This can be a problem for non decorator alternatives to decorators provided functionality, such as `@theme` as there is no reasonable way to ensure that `this.baseClasses` is available in in the mixins constructor.

This will become a core principle of Dojo 2 widgets, that only variable initialisation and listening to events (`.on`) should happen in the constructor.

Resolves #356

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -103,9 +103,8 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 
 	/**
 	 * @constructor
-	 * @param options widget options for construction
 	 */
-	constructor(properties: P) {
+	constructor() {
 		super({});
 
 		this._children = [];
@@ -134,15 +133,13 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 		this.own(this.on('properties:changed', (evt: PropertiesChangeEvent<WidgetBase<WidgetProperties>, WidgetProperties>) => {
 			this.invalidate();
 		}));
-
-		this.setProperties(properties);
 	}
 
 	public get properties(): Readonly<P> {
 		return this._properties;
 	}
 
-	public setProperties(properties: P & { [index: string]: any }): void {
+	public setProperties(properties: P): void {
 		const diffPropertyResults: { [index: string]: PropertyChangeRecord } = {};
 		const diffPropertyChangedKeys: string[] = [];
 
@@ -150,7 +147,7 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 
 		this.diffPropertyFunctionMap.forEach((property: string, diffFunctionName: string) => {
 			const previousProperty = this.previousProperties[property];
-			const newProperty = properties[property];
+			const newProperty = (<any> properties)[property];
 			const self: { [index: string]: any } = this;
 			const result: PropertyChangeRecord = self[diffFunctionName](previousProperty, newProperty);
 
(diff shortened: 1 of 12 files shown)


src/mixins/Themeable.ts
```diff
@@ -273,7 +278,18 @@ export function ThemeableMixin<T extends Constructor<WidgetBase<WidgetProperties
...
 
 			if (themeChanged || overrideClassesChanged) {
-				this.generatedClassNames = this.generateThemeClasses(this.baseClasses, theme, overrideClasses);
+				this.initialised ?
+					this.generateThemeClasses(this.baseClasses, theme, overrideClasses) :
+					this.initiliaseClasses();
+			}
+		}
+
+		private initiliaseClasses() {
```
agubler: Some doc and update to US spelling.

src/mixins/Themeable.ts
```diff
@@ -159,15 +159,19 @@ export function ThemeableMixin<T extends Constructor<WidgetBase<WidgetProperties
 		private generatedClassNames: ClassNameFlagsMap;
 
 		/**
+		 * Indicates if the base classes have been initialised.
+		 */
+		private initialised: boolean;
```
agubler: US spelling please.

src/WidgetBase.ts
```diff
@@ -134,23 +133,21 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
+	public setProperties(properties: P): void {
 		const diffPropertyResults: { [index: string]: PropertyChangeRecord } = {};
 		const diffPropertyChangedKeys: string[] = [];
 
 		this.bindFunctionProperties(properties);
 
 		this.diffPropertyFunctionMap.forEach((property: string, diffFunctionName: string) => {
 			const previousProperty = this.previousProperties[property];
-			const newProperty = properties[property];
+			const newProperty = (<any> properties)[property];
```
agubler: Does anyone know a better way to do this? I don't want an index type on the object (WidgetProperties) as would like it to correctly report unknown properties...?

src/mixins/Themeable.ts
dylans: that's neither US nor UK spelling there ;)

agubler: haha indeed.

src/mixins/Themeable.ts
```diff
@@ -273,7 +278,21 @@ export function ThemeableMixin<T extends Constructor<WidgetBase<WidgetProperties
...
 			if (themeChanged || overrideClassesChanged) {
-				this.generatedClassNames = this.generateThemeClasses(this.baseClasses, theme, overrideClasses);
+				this.initialized ?
+					this.generateThemeClasses(this.baseClasses, theme, overrideClasses) :
+					this.initializeClasses();
+			}
+		}
+
+		/**
+		 * Initialize the reverse loop up map and the generatedThemeClasses.
```
matt-gadd: reverse loop?

agubler: I don't know what you are referring to 😉

Issue #352, opened 2017-02-13
Decorator functionality in non-decorator environments


matt-gadd:
For users who do not want to use `experimentalDecorators` in TypeScript, and consumers who use `widget-core` as JS , we are going to need to provide equivalent functionality without Decorators.

agubler:
resolved by #358

PR #358, opened 2017-02-15
Widget Base decorator support including `diffProperty` and `afterRender`

### LINKED ISSUE #330: Widget base support for decorating properties
**Enhancement** 

With the adoption of classes comes the possibility to use decorators, to enrich (decorate if you will) classes. As a result we should provide a known object bag from `WidgetBase` for all decorators to use to add properties into that can be used by mixins.

This should be a best practice to help reduce the pollution of the prototype and provide a generic place for consumers to retrieve decorated properties.

agubler:
**Type:** feature

**Description:**

Add built in widget base support for decorators and support base functionality in non decorator environments. Add `@afterRender` and `@diffProperty` core decorator used like 

```ts
class MyWidget extends WidgetBase{
    
    @diffProperty('foo')
    diffFoo(previousProperty: any, newProperty: any) {
        // do diff here
        return {
            changed: true,
            value: 'myFoo'
        }
    }

    @afterRender
    addThingsToRenderResult(result: DNode) {
        // do things here
        return result;
    }
}
```

With the equivalent functionality in a non decorator environment using `addDecorator` in the constructor.

```ts
class MyWidget extends WidgetBase{
    
    constructor() {
        super();
        this.addDecorator('diffProperty', { propertyName: 'foo', diffFunction: this.diffFoo });
        this.addDecorator('afterRender', this.addThingsToRenderResult);
    }

    diffFoo(previousProperty: any, newProperty: any) {
        // do diff here
        return {
            changed: true,
            value: 'myFoo'
        }
    }

    addThingsToRenderResult(result: DNode) {
        // do things here
        return result;
    }
}
```

Resolves #330 & #352

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -27,14 +27,30 @@ interface WidgetCacheWrapper {
 }
 
 /**
- * Regular expression to find specific property diff functions
+ * Diff property configuration
  */
-const propertyFunctionNameRegex = /^diffProperty(.*)/;
+interface DiffPropertyConfig {
+	propertyName: string;
+	diffFunction: Function;
+}
 
 /**
- * Regular expression to find render decorator functions
+ * Decorator that can be used to register a function to run as an aspect to `render`
  */
-const decoratorFunctionNameRegex = /^renderDecorator.*/;
+export function afterRender(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+	target.addDecorator('afterRender', target[propertyKey]);
+}
+
+/**
+ * Decorator that can be used to register a function as specific property diff
+ *
+ * @param propertyName the name of the property that the diff function is for
+ */
+export function diffProperty(propertyName: string) {
+	return function (target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+		target.addDecorator('diffProperty', { propertyName, diffFunction: target[propertyKey] });
+	};
+}
 
 /**
  * Main widget base for all widgets to extend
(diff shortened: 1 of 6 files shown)


src/WidgetBase.ts
```diff
@@ -27,14 +27,30 @@ interface WidgetCacheWrapper {
...
- * Regular express to find render decorator functions
+ * Decorator that can be used to register a function to run as an aspect to `render`
  */
-const decoratorFunctionNameRegex = /^renderDecorator.*/;
+export function afterRender(target: any, propertyKey: string, descriptor: PropertyDescriptor) {
+	target.addDecorator('afterRender', target[propertyKey]);
+}
+
+/**
+ * Decorator that can be used to register a function as specific property diff
```
dylans: as a specific property diff

src/WidgetBase.ts
```diff
@@ -27,14 +27,30 @@ interface WidgetCacheWrapper {
...
+ *
+ * @param propertyName the name of the property that the diff function is for
```
dylans: Perhaps "The name of the property of which the diff function is applied"?

src/WidgetBase.ts
```diff
@@ -97,6 +113,11 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	private bindFunctionPropertyMap: WeakMap<(...args: any[]) => any, { boundFunc: (...args: any[]) => any, scope: any }>;
 
 	/**
+	 * A generic bag for decorators
```
dylans: property bag?

PR #359, opened 2017-02-16
Remove dependency on Maquette's Projector.

matt-gadd:
**Type:** feature

**Description:**
This removes our reliance on the Maquette projector, so we can now have more control over things such as event handling.

* Events now are listeners rather than handlers, this should mean no longer needing an `afterCreate` in widgets for things such as `transitionend`. As well as working with polyfills such as PEPjs.

* At the moment, maquette gives me no way to actually tear down event listeners on removal of a domnode - these should be garbage collected but we will have to see. 

* Event listeners are added on a per node basis, nothing clever. This may be problematic with a large number of listeners. I have code here: https://github.com/matt-gadd/widget-core/blob/projector-replace/src/mixins/Projector.ts#L150-L173. Which switches to delegating all the events to the root node of the projector, so there is only ever 1 listener per event for everything. There are issues with this approach for events that don't bubble such as `onfocus` and `onblur` though. In the future we might want to switch to a synthetic event system similar to Reacts.

* We should consider making all of our events `passive` by default, given the way things are going. Being able to configure this would be useful also.

* We no longer schedule a render when an event is triggered which is what Maquette does by default. This is simply because it doesn't make sense to do this with our invalidation system.

* In the future we might want to consider inverting this, making the eventHandlerInterceptor a no-op and passing an addEventHandler function in the `projectorOptions`, which nodes receive in the `afterCreate` and `afterUpdate`. This would be nicer as the registration would happen from the widget itself.

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -1,8 +1,6 @@
 import { Handle } from '@dojo/interfaces/core';
 import global from '@dojo/core/global';
 import Promise from '@dojo/shim/Promise';
-import WeakMap from '@dojo/shim/WeakMap';
-import Map from '@dojo/shim/Map';
 import { dom, Projection, ProjectionOptions, VNodeProperties } from 'maquette';
 import { WidgetBase } from './../WidgetBase';
 import { Constructor, WidgetProperties } from './../interfaces';
@@ -82,11 +80,6 @@ export interface ProjectorMixin {
 	readonly projectorState: ProjectorAttachState;
 }
 
-export interface EventIntercepterItem {
-	handler: Function;
-	properties: VNodeProperties;
-}
-
 export function ProjectorMixin<T extends Constructor<WidgetBase<WidgetProperties>>>(base: T): T & Constructor<ProjectorMixin> {
 	return class extends base {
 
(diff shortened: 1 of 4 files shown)


src/mixins/Projector.ts
```diff
@@ -54,40 +57,71 @@ export interface ProjectorMixin {
...
 	root: Element;
 
 	/**
 	 * The status of the projector
 	 */
-	readonly projectorState: ProjectorState;
+	readonly projectorState: ProjectorAttachState;
+}
+
+export interface EventIntercepterItem {
```
agubler: do we need to export this interface?

src/mixins/Projector.ts
```diff
@@ -54,40 +57,71 @@ export interface ProjectorMixin {
...
+		private boundDoRender: FrameRequestCallback;
+		private boundRender: Function;
+		private eventIntercepterNodeMap: WeakMap<Node, Map<string, EventIntercepterItem>>;
 
 		constructor(...args: any[]) {
 			super(...args);
-			const maquetteProjectorOptions = {
-				transitions: cssTransitions
+
+			this.eventIntercepterNodeMap = new WeakMap<Node, Map<string, EventIntercepterItem>>();
```
agubler: Can we do this initialization in outside the constructor?

src/mixins/Projector.ts
```diff
@@ -145,34 +202,50 @@ export function ProjectorMixin<T extends Constructor<WidgetBase<WidgetProperties
...
 		}
 
-		private scheduleRender(event: EventTargettedObject<this>) {
-			const { target: projector } = event;
-			if (this.projectorState === ProjectorState.Attached) {
-				projector.emit({
-					type: 'render:scheduled',
-					target: projector
+		private eventHandlerInterceptor(propertyName: string, callback: Function, domNode: Node, properties: VNodeProperties) {
+			const eventName = propertyName.substr(2);
```
agubler: What's `2`, is it to remove `on`?

matt-gadd: yes

matt-gadd:
After some thinking, i've decided for now that this PR should simply move us away from the Maquette projector - and we can further think on how we can better handle events. This means I've reverted the `eventHandlerInterceptor` code to be more like Maquette's currently is, just without the `scheduleRender`

PR #372, opened 2017-02-20
`WidgetBase` and `Themeable` Array Optimisations

agubler:
**Type:** feature

**Description:**

Convert uses of array functions `forEach`, `reduce`, `map`, `filter` to simple `for` loops.

Whilst this may seem trivial the `render`, `this.classes` and `setProperties` pipeline can get executed thousands of times each lifecycle.

*Benchmark (20 Runs) - Adding 1000 Todos, Completing Todos (Individually) and Removing Todos (Individually)*

Chrome 56

todomvc_benchmark

Safari 10.0.3

todomvc_benchmark

Firefox 51

todomvc_benchmark

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -212,13 +212,14 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	}
 
 	public diffProperties(previousProperties: P & { [index: string]: any }, newProperties: P & { [index: string]: any }): PropertiesChangeRecord<P> {
-		const changedKeys = Object.keys(newProperties).reduce((changedPropertyKeys: string[], propertyKey: string): string[] => {
-			if (previousProperties[propertyKey] !== newProperties[propertyKey]) {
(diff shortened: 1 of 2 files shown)


agubler:
@dylans I didn't run on any other browser for those comparisons, but certainly can do. Both myself and Matt G mentioned it would be great if it could be done in the transpilation stage however that's not possible. Whilst I agree in some case, I actually think the the change for `this.classes` are much better in terms of code clarity and don't think the others are actually too much of a step back at all. Obviously that is subjective though.

As for a comment, I think that we are going to normalise all array code blocks to avoid using the helper functions across the entire `widget-core` code base as it certainly seems like WidgetBase and Themeable are going to be especially hot paths. We should put something in the readme if that is the case.

agubler:
@dylans I have updated the description to show the comparison graphs between the before and after across the 3 browsers I have installed (chrome, firefox, safari).

dylans:
Ok, fair enough.

PR #375, opened 2017-02-22
Revert "`WidgetBase` and `Themeable` Array Optimisations (#372)"

agubler:
This reverts commit 661088e5313d4e596c4babefe6565103962e774a.

**Type:** bug

**Description:**

This change reverts the updates through `WidgetBase` and `Themeable` to use `for` loops. The changes were bugged, as `break;` was used where it should have been `continue;` (all unit tests passed 😞). Once these errors had been corrected the benchmark difference was negligible and therefore it is preferable to keep the newer array constructs.


matt-gadd:
Given the big perf gains in `this.classes` by the erroneous change, maybe we should raise an issue to look at performance specifically for that?

agubler:
@matt-gadd Okay.

PR #378, opened 2017-02-24
Remove symbol-shim typings

agubler:
**Type:** feature

**Description:**

Remove dependency on symbol shim typings, requires https://github.com/dojo/core/pull/294 and https://github.com/dojo/i18n/pull/62 to be released.

Related to https://github.com/dojo/meta/issues/113

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,6 +4,13 @@
 		"declaration": false,
 		"emitDecoratorMetadata": true,
 		"experimentalDecorators": true,
+		"lib": [
+			"dom",
+			"es5",
+			"es2015.iterable",
+			"es2015.symbol",
+			"es2015.symbol.wellknown"
+		],
 		"module": "umd",
 		"moduleResolution": "node",
 		"noImplicitAny": true,
--- a/package.json
+++ b/package.json
@@ -23,10 +23,10 @@
     "test": "grunt test"
   },
   "peerDependencies": {
-    "@dojo/core": "2.0.0-alpha.24",
-    "@dojo/has": "2.0.0-alpha.7",
-    "@dojo/i18n": "2.0.0-alpha.6",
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -4,6 +4,13 @@
 		"declaration": false,
 		"emitDecoratorMetadata": true,
 		"experimentalDecorators": true,
+		"lib": [
+			"dom",
+			"es5",
+			"es2015.iterable",
+			"es2015.symbol",
+			"es2015.symbol.wellknown"
+		],
```
dylans: Don't you need `"types": [ "intern" ]` somewhere in here?

agubler: Nope actually we don't for our internal packages.

agubler: I could add it so that when we need it in the future it is there.

dylans: It was in all of the other similar PRs which is why it caught my attention, e.g. https://github.com/dojo/shim/pull/76/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R23 , so mostly I noticed that this and i18n were inconsistent with the earlier PRs.

agubler: It wasn't/isn't needed in the `shim` one, the `core` PR also did not include the `intern` types in the `tsconfig.json`.

agubler: Once we land https://github.com/dojo/grunt-dojo2/pull/99 and release we can actually remove the typings file completely (at which point we do need the `"types": [ "intern" ]` - perhaps we should include that in these changes.

dylans: ok, I was just comparing against the first PR raised and didn't notice the issue... it sounds like adding `"types": [ "intern" ]` could save us another round of meta updates after this lands, but it's your call. :)

agubler: @dylans just to let you know this has already been done as part of the changes for the reporter.

PR #382, opened 2017-02-24
Add dom node lifecycle callbacks to WidgetBase

### LINKED ISSUE #361: Make dom nodes created from `v` accessible in a widget
At the moment the only mechanism we have for accessing a dom node is via maquette's `afterCreate` and `afterUpdate` methods which are per `v` call which is a bit clunky. We perhaps should provide a a mechanism in WidgetBase that allows you to access all dom nodes in a widget by `key`, as well as a lifecycle function on WidgetBase where you know these will be available.

edhager:
**Type:** feature

I added `onElementCreated` and `onElementUpdated` methods to `WidgetBase` so widget authors can have easy access to the dom elements that are created for the widget from the vdom.

This update registers vdom event handlers for `afterCreate` and `afterUpdate` which in turn call  `onElementCreated` and `onElementUpdated`.  The vdom event handlers are static methods to abide by the vdom rule that property values that are functions in the vdom node properties object can't change between updates.  Those vdom event handlers need references to the widgets so `onElementCreated` and `onElementUpdated` can be called and they need references to any vdom event handlers that were set in the `v` calls.  That info is being stored in the nodes' property objects.   I originally tried using a map but it is difficult to determine a key for that map.  vdom nodes can have the same `key` property if they have different parent nodes.

Resolves #361


src/WidgetBase.ts
```diff
@@ -160,6 +181,62 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
+				}
+			}, isHNode);
+			return node;
+		});
+	}
+
+	private setLifecycleProperties(properties: VirtualDomProperties, callbackKey: string) {
+		let lsProps = properties._lifecycleProperties;
+		if (!lsProps) {
+			properties._lifecycleProperties = lsProps = {
```
matt-gadd: I'm not sure if this is a good idea, isn't the widget reference going to end up on a dom node property? Which seems a bit leaky, and potentially cyclical? If we go down this route we'll now also be squatting on a property that we'd need to document - I don't think it'd be possible to actually stop it via interfaces which is a shame (due to the index signature).

src/WidgetBase.ts
```diff
@@ -160,6 +181,62 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
+		// Do nothing by default.
+	}
+
+	/**
+	 * Widget lifecycle method that is called whenever a vdom node is updated.
+	 * Override this method to access the dom nodes that were updated in the dom.
+	 * @param element The dom node represented by the vdom node.
+	 * @param key The vdom node's key.
+	 */
+	protected onElementUpdated(element: Element, key: string): void {
```
matt-gadd: I'm starting to be skeptical on how useful maquette's `afterUpdate` function actually would be to the end user. It seems like its called regardless of whether the specific dom node is updated or not, which kind of makes me think if we decide to keep this lifecycle hook we should perhaps only call it once per update (and not per node).

matt-gadd:
Given the new lifecycle hooks do you think we should narrow our `VirtualDomProperties` interface to no longer take an `afterCreate` or `afterUpdate` function? It seems like that would no longer be the recommended way to do things.

matt-gadd:
> vdom nodes can have the same key property if they have different parent nodes.
I'm not sure if we should care about supporting this for widgets. It will be difficult for an end user to actually uniquely identify a node which has the same key anyway (if we're not passing `VNodeProperties` through in the lifecycle hooks).

Given maquette will actually do no duplicate `key` checking regardless (even on sibling nodes). Perhaps we would better only supporting unique keys within a widget, and logging a warn when duplicate keys are used.

edhager:
@matt-gadd Take a look at this prototype.  https://github.com/edhager/widget-core/commit/aab68781a86473b6bf566c14dd0ffe19f20382f5

If we remove the requirement that the `afterCreate` and `afterUpdate` callbacks set in the widget implementations has to be maintained by the lifecycle code, then the code in WidgetBase is greatly simplified.  

The `afterCreate` and `afterUpdate` callbacks created in WidgetBase only need references to the widget.  That is handled by binding the callback functions to the widget outside of the render functions.  That way each time render is called on the vnodes, the functions don't change.  This approach doesn't care whether the keys are unique or not because they are not being used to identify the vnodes.  The code is only making sure a vnode has a key before it sets the callback functions.

If you think this is the way to go, I can make the changes in my branch that is tied to this PR.

edhager:
@matt-gadd I merged the changes into my widgetbase-node-lifecycle branch.

src/WidgetBase.ts
```diff
@@ -160,6 +180,44 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
+		// and onElementUpdated
+		this.addDecorator('afterRender', (node: DNode) => {
+			// Using callback functions that were bound outside of this render call to make sure
+			// the functions don't change.
+			decorate(node, (node: HNode) => {
+				node.properties.afterCreate = afterCreateCallback;
+			}, isHNodeWithKey);
+
+			decorate(node, (node: HNode) => {
+				node.properties.afterUpdate = afterUpdateCallback;
```
matt-gadd: could the after update hook be added in the decorator call above? saves having to iterate the nodes twice?

tests/unit/lifecycle.ts
```diff
@@ -0,0 +1,261 @@
...
+
+	public render(): HNode {
+		let vnode = this.vnode;
+
+		if (vnode) {
+			if (this.modify) {
+				vnode.children.push(v('div', {
+					key: 'addition',
+					id: 'addition',
+					afterCreate: incrementCreateCounter.bind(this),
```
matt-gadd: i'm a little confused by these tests. I thought we were going to remove the `afterUpdate` and `afterCreate` hooks from `v` completely?

edhager: v() can be used outside of widget development right?

src/WidgetBase.ts
edhager: Yeah. I'll make that change.

tests/unit/lifecycle.ts
matt-gadd: the only thing that can prcoess `v` is WidgetBase

edhager: I wasn't planning on removing them from the `v` interfaces.  I can if you want.  

We will need a subset of VirtualDomProperties to use with `v` because we still need `afterUpdate` and `afterCreate` in VirtualDomProperties for use in the `HNode` interface.

As these tests show, if the vnode does not have a key, then the `afterCreate` and `afterUpdate` callbacks passed to `v` will be called.  The lifecycle methods are attached to only vnodes with key properties.

matt-gadd: Yeah I see. It would be nice if we could narrow the interface, but the way we currently have it makes that kind of tricky (because you still need it when you attach the afterCreate and afterUpdate). Happy to leave this for now

matt-gadd:
@edhager really sorry for the late reply to your comments. I think we can take this if you resolve the conflicts.

PR #390, opened 2017-03-08
Include Pointer Event Polyfill with Projector

matt-gadd:
**Type:** feature

**Description:**

This is to discuss, as its one of the few polyfills we actually have that are external and not from `dojo/shim`. I think we should include it by default in the projector, rather than it being opt-in. We've done similar with css-transitions and the projector, and pepjs is only 6kb min & gzipped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,5 +48,8 @@
     "sinon": "^1.17.6",
     "tslint": "^3.11.0",
     "typescript": "~2.2.1"
+  },
+  "dependencies": {
+    "pepjs": "^0.4.2"
   }
 }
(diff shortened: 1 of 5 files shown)


dylans:
I agree that we should include it by default as this is not something that someone should need to think about.

matt-gadd:
this will need changes downstream to all the intern configs before it can be taken

agubler:
@matt-gadd is that examples (via the `cli-test-intern` command) and widgets as it stands?

matt-gadd:
@agubler I don't believe `cli-test-intern` needs updating (due to it using webpack), just widgets. Which I think are the following: 
https://github.com/dojo/widgets/blob/master/src/common/tests/intern.ts
https://github.com/dojo/widgets/blob/master/src/dialog/example/index.html
https://github.com/dojo/widgets/blob/master/src/label/example/index.html
https://github.com/dojo/widgets/blob/master/src/slidepane/example/index.html

Which probably further validates the need for https://github.com/dojo/widgets/issues/27

agubler:
@matt-gadd Raised https://github.com/dojo/widgets/pull/62

agubler:
This should be good to go now.

PR #392, opened 2017-03-08
Update all references to factory

### LINKED ISSUE #391: Update all references to `factory`
**Enhancement** 

The are lots of legacy references to `factory` that remain in the code base that need to be updated.

agubler:
**Type:** bug

**Description:**

Remove legacy references to `factory`.

Outstanding is the packge name of the `WidgetRegistry` which is still called `FactoryRegistry.ts`.

Resolves #391


matt-gadd:
I presume this is a breaking change due to the `customElement` api rename. If so, shall we do the rename of `FactoryRegistry.ts` in this PR too?

PR #394, opened 2017-03-09
Private variables and visibility clean up

### LINKED ISSUE #388: Review function visibility (public/protected/private)
**Bug** 

Review the visibility of all functions across the project.

agubler:
**Type:** feature

**Description:**

Update private variables prefixed with `_` and update visibility for class method/properties.

Resolves #388

### DIFF
--- a/src/interfaces.d.ts
+++ b/src/interfaces.d.ts
@@ -371,13 +371,6 @@ export interface WidgetBaseInterface<P extends WidgetProperties> extends Evented
 	diffProperties(previousProperties: P & { [index: string]: any }, newProperties: P & { [index: string]: any }): PropertiesChangeRecord<P>;
 
 	/**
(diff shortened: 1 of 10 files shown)


src/interfaces.d.ts
```diff
@@ -371,13 +371,6 @@ export interface WidgetBaseInterface<P extends WidgetProperties> extends Evented
 	diffProperties(previousProperties: P & { [index: string]: any }, newProperties: P & { [index: string]: any }): PropertiesChangeRecord<P>;
 
 	/**
-	 * Default render, returns a `div` with widget's children
```
dylans: My only question is, why does this go away.

agubler: Because it is now `protected` so cannot be on the interface.

PR #396, opened 2017-03-09
Changing the way decorators are stored

### LINKED ISSUE #395: Confusing behavior with class decorators
**Bug**

Base classes are getting decorators intended for subclasses.

**Code**

```typescript
function decorator(value: string) {
    return function (constructor: Function) {
        constructor.prototype.addDecorator(value);
    };
}

class Base {
    protected _decorators: any;

    protected addDecorator(value: string) {
        if (!this._decorators) {
            this._decorators = [];
        }

        this._decorators.push(value);
    }

    getDecorators() {
        return this._decorators;
    }
}

@decorator('Test1')
class Test1 extends Base {
}

@decorator('Test2')
class Test2 extends Test1 {
}

const test1 = new Test1();
const test2 = new Test2();

console.log("test1", test1.getDecorators());
console.log("test2", test2.getDecorators());
```

The logs reveal that `Test1` actually has two decorators now, instead of only one. The way this snippet works is similar to how `WidgetBase` adds decorators, which has the same problem.

**Expected behavior:**

`Test1` should have one decorator, `"Test1"`

**Actual behavior:**

`Test` has two decorators, `"Test2"`

rorticus:
**Type:** bug / feature

**Description:**

Fix for confusing class decorators

Resolves #395

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -1,13 +1,13 @@
 import { Evented, BaseEventedEvents } from '@dojo/core/Evented';
 import { assign } from '@dojo/core/lang';
 import { EventedListenerOrArray } from '@dojo/interfaces/bases';
+import { Handle } from '@dojo/interfaces/core';
 import { VNode } from '@dojo/interfaces/vdom';
 import Map from '@dojo/shim/Map';
 import Promise from '@dojo/shim/Promise';
 import Set from '@dojo/shim/Set';
 import WeakMap from '@dojo/shim/WeakMap';
 import { v, registry, isWNode } from './d';
-import WidgetRegistry, { WIDGET_BASE_TYPE } from './WidgetRegistry';
 import {
 	DNode,
 	WidgetConstructor,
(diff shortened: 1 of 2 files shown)


src/WidgetBase.ts
```diff
@@ -292,25 +292,68 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	 */
 	protected addDecorator(decoratorKey: string, value: any): void {
 		value = Array.isArray(value) ? value : [ value ];
-		if (!this._decorators) {
-			this._decorators = new Map<string, any[]>();
+
+		let decoratorList = decoratorMap.get(this.constructor);
```
agubler: nit: it's a Map not a list

src/WidgetBase.ts
```diff
@@ -292,25 +292,68 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
+	 * @return An array of decorator values
+	 * @private
+	 */
+	private _buildDecoratorList(decoratorKey: string): any[] {
+		const allDecorators = [];
+
+		let constructor = this.constructor;
+
+		while (constructor) {
+			const decoratorList = decoratorMap.get(constructor);
```
agubler: nit: same as above, it's a map.

src/WidgetBase.ts
rorticus: Updated!

src/WidgetBase.ts
rorticus: Updated!

PR #398, opened 2017-03-09
Adding an option to propagate properties to child widgets, issue #393

### LINKED ISSUE #393: Add decorator for automatic property passing
**Enhancement** 

We'd like a decorator to automatically propagate properties from parent widgets to child widgets. This would make it easier to pass properties like `theme` from the application all the way down to the child widgets.

Intended usage:

```typescript
@passProperty('theme')
    // theme is automatically passed to child widgets
}
```

rorticus:
**Type:** feature

**Description:**

Adding a decorator that will propagate a single property from a parent widget to child widgets,

```typescript
interface MyProperties extends WidgetProperties {
    myProperty: string;
}

@propagateProperty('myProperty')
class MyWidget extends WidgetBase<MyProperties> {
    render() {
        return w(AnotherWidget, {});
    }
}
```

In this example, at runtime, the`AnotherWidget` wnode is constructed with the `myProperty` value from `MyWidget`.

There are some rules on this:

* If the propagated property does not exist in the widget's properties, it is not passed along.
* If the propagated property is already explicitly passed, the propagated value is ignored.

The `ThemeableMixin` was modified to propagate the `theme` property, and the `I18nMixin` was modified to propagate the `locale` and `rtl` properties.

Resolves #393


rorticus:
Added `registry`!

src/WidgetBase.ts
```diff
@@ -385,6 +391,13 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 			const { children, properties = {} } = dNode;
 			const { key } = properties;
 
+			const propagatedPropertyNames = this.getDecorator('propagateProperty');
```
agubler: Do you think this would be better as an `@afterRender` that utilises the decorator from `d.ts`? so runs after each render rather than in dNodeToVNode?

agubler:
Can we update the README, we might even need a section on decorators.

src/WidgetBase.ts
matt-gadd: was about to comment on this. I was indeed hoping we'd be able to do this down the standard decoration mechanism in `@afterRender` rather than adding additional code to WidgetBase

rorticus:
README updated!

src/WidgetBase.ts
rorticus: I talked with @matt-gadd a bit about this this morning. There doesn't seem to be a clever way to actually use `afterRender` here. We can't add it from the `propagateProperty` decorator function because then you can't add the decorator programatically. Also, running `propagateProperty` 5 times would add 5 `afterRender` hooks, each looping over the resulting node, which is less efficient than looping over once.

This may need a good think 😄

agubler: I'm out at lunch (not that you need to know that), but I'll have a think :)

agubler: I was thinking more that there would be be a an method that would take care of the propagation.

```ts
@afterRender
propagateProperties(result: DNode) {
    const propagatedPropertyNames = this.getDecorator('propagateProperty');
    decorate(result, (dnode: DNode) => { /* add properties */ }, isWNode);
}
```

And looking at this makes me think that it could really be a separate mixin that added this `afterRender`.

Perhaps I'm way off base here.

rorticus: Yeah we could add a `PropagatePropertiesMixin`, but using the decorator wouldn't be so convenient as you'd have to make sure you used this mixin (and know that something else you're using, like `ThemeableMixin` is already using it).

agubler: I thought maybe we could mix that in automatically for mixins that require it, again might be that it's not possible.

```ts
export function FakeMixin<T extends Constructor<WidgetBase<FakeMixinProperties>>>(base: T): T {
	class FakeMixin extends PropagatePropertyMixin(base) {

        @afterRender
        protected propagateProperties(result: DNode) {
            decorate(result, (result: DNode) => {}, isWNode);
            return result;
        }
    }
    return FakeMixin;
}

export default FakeMixin;
```

agubler:
@rorticus I don't think that this will be required once the injector/container pattern has landed, closing for now. We can re-open if we think that the functionality is still valuable.

PR #408, opened 2017-03-15
Update @dojo/core to alpha.26

mwistrand:
**Type:** enhancement

**Description:**

Updates the `@dojo/core` peer dependency to alpha.26.

Resolves #???


agubler:
@mwistrand You need to optimistically update the about to be released versions of other dependencies too (like `i18n` for widget-core)

mwistrand:
i18n alpha.8 is released, so all peer deps use the same version of core.

PR #411, opened 2017-03-15
Add attribute to sniff and fork addDecorator code depending caller

### LINKED ISSUE #406: Can't use `addDecorator` on an instance anymore
**Bug** 

The latest changes to store decorators based on class type, have made it so that you can no longer call `this.addDecorator` to add a decorator that applies only to a single instance of a class.  Calling `addDecorator` creates a decorator that applies to _every_ instance of the class.

**Code**

The following test illustrates the problem,

```typescript
'decorators can be applied at instance level'() {
	let renderCallCount = 0;

	class TestWidget extends WidgetBase<any> {
		constructor() {
			super();

			this.addDecorator('afterRender', (node: DNode) => {
				renderCallCount++;
				return node;
			});
		}

		render() {
			return v('div');
		}
	}

	new TestWidget();
	const widget2 = new TestWidget();
	widget2.__render__();

	assert.equal(renderCallCount, 1);
}
```

**Expected behavior:**

`renderCallCount` should be 1

**Actual behavior:**

`renderCallCount` is 2 because a new `afterRender` is added every time a new `TestWidget` is instantiated.

agubler:
**Type:** bug

**Description:**

Fork behaviour for adding decorators functionality depending whether added programmatically or via typescript decorators.

Resolves #406

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -85,8 +85,6 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	 */
 	static _type: symbol = WIDGET_BASE_TYPE;
 
-	private _constructed: boolean = true;
-
 	/**
 	 * on for the events defined for widget base
 	 */
@@ -337,7 +335,11 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	 */
 	protected addDecorator(decoratorKey: string, value: any): void {
 		value = Array.isArray(value) ? value : [ value ];
-		if (!this._constructed) {
+		if (this._properties) {
+			const decorators = this._decoratorCache.get(decoratorKey) || [];
+			this._decoratorCache.set(decoratorKey, [ ...decorators, ...value ]);
+		}
+		else {
 			let decoratorList = decoratorMap.get(this.constructor);
 			if (!decoratorList) {
 				decoratorList = new Map<string, any[]>();
@@ -350,10 +352,6 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 				decoratorList.set(decoratorKey, specificDecoratorList);
 			}
 			specificDecoratorList.push(...value);
-		}
-		else {
-			const decorators = this._decoratorCache.get(decoratorKey) || [];
-			this._decoratorCache.set(decoratorKey, [ ...decorators, ...value ]);
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/WidgetBase.ts
```diff
@@ -335,20 +335,24 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
...
 	protected addDecorator(decoratorKey: string, value: any): void {
 		value = Array.isArray(value) ? value : [ value ];
-
-		let decoratorList = decoratorMap.get(this.constructor);
-		if (!decoratorList) {
-			decoratorList = new Map<string, any[]>();
-			decoratorMap.set(this.constructor, decoratorList);
+		if (this._properties) {
+			const decorators = this._decoratorCache.get(decoratorKey) || [];
+			this._decoratorCache.set(decoratorKey, [ ...decorators, ...value ]);
```
rorticus: Since `_decoratorCache` is populated on the first call to `getDecorator`, I think maybe that calling `addDecorator` and then `getDecorator` will _only_ get the one you just put in the cache, and not any of the others that are on the class level (because the cache is set so it won't build the decorator list). Am I crazy?

agubler: You cannot use decorators and the programmatic API together, it's one or the other. In the get call it just returns the entry from `_decoratorCache` if it exists so adding them directly into there when dealing with an instance should be fine.

There is a sentence in the README indicating that the two methods should not be mixed.

I could however move the population of the `_decoratorCache` to the WidgetBase constructor now that using the API adds them directly into the instance's cache...

rorticus: Ah! OK then. If they are only being applied to the instance than life should be good.

src/WidgetBase.ts
```diff
@@ -335,20 +335,24 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	 */
 	protected addDecorator(decoratorKey: string, value: any): void {
 		value = Array.isArray(value) ? value : [ value ];
-
-		let decoratorList = decoratorMap.get(this.constructor);
-		if (!decoratorList) {
-			decoratorList = new Map<string, any[]>();
-			decoratorMap.set(this.constructor, decoratorList);
+		if (this._properties) {
```
matt-gadd: up to you if you want to keep this or not, but I think we can sniff if we're the instance or not with the following:
```ts
this.hasOwnProperty('constructor');
```

PR #414, opened 2017-03-16
Add granular DiffTypes to the diffProperty decorator

matt-gadd:
**Type:** bug / feature

**Description:**
This resolves partially https://github.com/dojo/widget-core/issues/404. It doesn't change any of the fallback diff behaviour, but now adds diffTypes that can be applied per property at a class level, as well as supporting the existing `diffProperty` behaviour. The new usage of `diffProperty` at a class level looks like:

```ts
@diffProperty('foo', DiffType.SHALLOW)
@diffProperty('bar', DiffType.REFERENCE)
class MyWidget extends WidgetBase<any> {
}
```

Note that the decorators are stackable and take a DiffType of the following:

```
CUSTOM,
ALWAYS,
IGNORE,
REFERENCE,
SHALLOW,
AUTO
```

The `AUTO` type will select a diff behaviour based on the following:
* If its a function, ignore diffing completely
* If its an object or array, diff the values shallowly by reference
* otherwise diff by reference


src/diff.ts
```diff
@@ -0,0 +1,103 @@
...
+			}
+			else if (isObjectOrArray(newProperty)) {
+				result = shallow(previousProperty, newProperty);
+			}
+			else {
+				result = reference(previousProperty, newProperty);
+			}
+		break;
+		default:
+			result = {
```
agubler: Perhaps we should emit or log a warning if a user accidentally gets here.

src/WidgetBase.ts
```diff
@@ -55,9 +59,14 @@ export function afterRender(target: any, propertyKey: string, descriptor: Proper
...
  */
-export function diffProperty(propertyName: string) {
-	return function (target: any, propertyKey: string, descriptor: PropertyDescriptor) {
-		target.addDecorator('diffProperty', { propertyName, diffFunction: target[propertyKey] });
+export function diffProperty(propertyName: string, diffType = DiffType.CUSTOM, diffFunction?: Function) {
+	return function (target: any, propertyKey?: string, descriptor?: PropertyDescriptor) {
+		if (diffType === DiffType.CUSTOM && propertyKey && descriptor) {
+			target.addDecorator('diffProperty', { propertyName, diffType, diffFunction: target[propertyKey] });
+		}
+		else if (diffType && !propertyKey && !descriptor) {
```
agubler: We don't cover the `else` branch (i'm not sure it is ever a realistic outcome)

src/diff.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function ignore(previousProperty: any, newProperty: any): PropertyChangeRecord {
+	return {
+		changed: false,
+		value: newProperty
+	};
+}
+
+function custom(previousProperty: any, newProperty: any, meta: any): PropertyChangeRecord {
```
matt-gadd: the `meta` is meant to be a generic value bag. we only have one usage of it at the moment which does have `diffFunction` and `scope`, but thats extremely coupled to the existing functionality of `diffProperty`. I think we'll revisit the decorator functions in https://github.com/dojo/widget-core/issues/402, so I think we should leave as is for now (given its internal anyway).

src/WidgetBase.ts
```diff
@@ -34,7 +37,8 @@ interface WidgetCacheWrapper {
  */
 interface DiffPropertyConfig {
 	propertyName: string;
-	diffFunction: Function;
+	diffType: DiffType;
+	diffFunction?: Function;
```
agubler: Sorry I meant `diffFunction` not `diffProperty`.

src/diff.ts
agubler: Understood

PR #421, opened 2017-03-21
Enhance property diffing system (refs #404)

### LINKED ISSUE #404: Enhancements to the existing Property diffing system
**Enhancement**

At the moment the Property Diffing system has the following mechanisms:
* per property control via the `@diffProperty($propertyName)` decorator. This allows the user to specify a custom diff function which returns a PropertyChangeRecord. This takes precedence over the catch all function.
* `diffProperties` in `WidgetBase`. This is the default catch all diff function for properties that do not have a per property control. Simple reference equality is used to determine whether a property has changed.

Because we have default behaviour for diffing, end user authors tend to presume that the diffing mechanism is sophisticated. Which so far has resulted in a few gotcha's:
* when passing functions (quite common for event handlers), if they are declared inline (or bound inline) in the `render` function, a new function will be created each render resulting in reference equality failing.
* when passing objects in a property inline in the the `render` function, once again a new object will be created per render so the reference equality will fail.
* when passing objects as a property and the same reference is used, and a value inside the object changes, reference equality is not adequate to pick up that change.

To remedy this we should consider whether we should have a default catch all behaviour at all and what that should be if it still exists. If not how can we still make it easy for the user to deal with property diffing while consciously opting-in to said behaviour.

rorticus:
**Type:** feature

**Description:**

- Changed automatic property diffing to use `DiffType.AUTO`.
- Automatic diff logic is no longer overridable.
- Multiple `@diffProperty` decorators supported on a single property. Decorators run until a changed property is found.

Resolves #404

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,7 +274,20 @@ If a property has a custom diff function then that property is excluded from the
 ##### The 'properties:changed' event
 
 When `diffProperties` has completed, the results are used to update the properties on the widget instance.
-If any properties were changed, then the `properties:changed` event is emitted.
+If any properties were changed, then the `properties:changed` event is emitted. If the new properties do **not** contain keys from the previous properties, the properties are marked as changed.
+
+```typescript
+// set the initial properties
+$widget->setProperties({
+	foo: true,
+	bar: true
+});
+
+// properties:changed will include the "bar" property
+$widget->setProperties({
+	foo: true
+});
+```
 
 Attaching a listener to the event is exposed via a decorator `@onPropertiesChanged`.
 
--- a/README.md
+++ b/README.md
@@ -200,36 +200,61 @@ Properties passed to the `w` function represent the public API for a widget.
 
 The properties lifecycle starts when properties are passed to the widget.
 The properties lifecycle is performed in the widgets `setProperties` function.
-This function uses the widget instance's `diffProperties` function to determine whether any of the properties have changed since the last render cycle.
-By default `diffProperties` provides a shallow comparison of the previous properties and new properties.
+Properties are differenced using `DiffType.AUTO`.
 
-The `diffProperties` function is also responsible for creating a copy (the default implementation uses`Object.assign({}, newProperties)` of all changed properties.
-The depth of the returned diff is equal to the depth used during the equality comparison.
-
-<!-- add example of 'depth' -->
-
-**Note:** If a widget's properties contain complex data structures that you need to diff, then the `diffProperties` function will need to be overridden.
+**Note:** If a widget's properties contain complex data structures that you need to diff, then individual control is required using the `diffProperty` decorator.
 
 ##### Custom property diff control
 
-Included in `WidgetBase` is functionality to support targeting a specific property with a custom comparison function using a decorator `diffProperty`. For non-decorator environments (Either JavaScript/ES6 or a TypeScript project that does not have the experimental decorators configuration set to true in the `tsconfig`), the decorator functions can be called directly from the constructor.
+You can control individual property differencing by using the `@diffProperty` decorator. Properties with a `diffProperty` decorator **will be excluded from automatic differencing**.
 
-e.g. for a property `foo` you would add a function to the widget class and either use the decorator function or register the decorator in the `constructor`.
+###### At the class level
 
-*using the `diffProperty` decorator*
+`@diffProperty(propertyName, diffType)` can be applied at the class level if you want to use a pre-defined diff function.
 
-```ts
+```typescript
+@diffProperty('title', DiffType.REFERENCE)
+class MyWidget extends WidgetBase<MyProperties> {
+}
+```
+
+The following diff functions are provided:
+
+| Type                 | Description                                                                       |
+| -------------------- | ----------------------------------------------------------------------------------|
+| `DiffType.ALWAYS`    | Always report a property as changed.                                              |
+| `DiffType.AUTO`      | Ignore functions, shallow compare objects, and reference compare all other values.|
+| `DiffType.CUSTOM`    | Provide a custom diffing function.                                                |
+| `DiffType.IGNORE`    | Never report a property as changed.                                               |
+| `DiffType.REFERENCE` | Compare values by reference (`old === new`)                                       |
+| `DiffType.SHALLOW`   | Treat the values as objects and compare their immediate values by reference.      |
+
+`DiffType.CUSTOM` is unique in that it takes a third parameter, the diff function. This function has the following signature:
+
+```
+(previousValue: any, newValue: any) => {
+  changed: boolean;
+  value: any;
+}
+```
+
+###### At the method level
+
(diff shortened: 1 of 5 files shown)


rorticus:
CI failures are from TypeDoc updates... please ignore :)

tests/unit/customElements.ts
```diff
@@ -38,6 +38,14 @@ function createFakeElement(attributes: any, descriptor: CustomElementDescriptor)
 	};
 }
 
+@diffProperties
+class PropertyDiffingWidgetBase<P> extends WidgetBase<P> {
+	getDecorators() {
+		console.log('The decorators are');
```
matt-gadd: is this debugging info needed?

rorticus: Good catch! Removed.

README.md
```diff
@@ -200,17 +200,40 @@ Properties passed to the `w` function represent the public API for a widget.
 
 The properties lifecycle starts when properties are passed to the widget.
 The properties lifecycle is performed in the widgets `setProperties` function.
-This function uses the widget instance's `diffProperties` function to determine whether any of the properties have changed since the last render cycle.
-By default `diffProperties` provides a shallow comparison of the previous properties and new properties.
 
-The `diffProperties` function is also responsible for creating a copy (the default implementation uses`Object.assign({}, newProperties)` of all changed properties.
-The depth of the returned diff is equal to the depth used during the equality comparison.
+This function uses the widget's `diffProperty` and `diffProperties` decorators to determine whether any of the properites have changed since the last render cycle.
```
matt-gadd: first off, thanks for being awesome and updating the readme 😄. do you think somewhere we should state in bold lettering that by default - changing a property will not cause a change/re-render, without adding either of those decorators?

src/WidgetBase.ts
```diff
@@ -242,15 +257,14 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 	}
 
 	public setProperties(properties: P): void {
-		const diffPropertyResults: { [index: string]: PropertyChangeRecord } = {};
+		const diffPropertyResults: { [index: string]: any } = {};
```
matt-gadd: is this no longer a `PropertyChangeRecord`?

README.md
rorticus: Yeah good idea.

src/WidgetBase.ts
rorticus: That's the weird thing. I don't think it was ever actually a `PropertyChangedRecord` 😄 (https://github.com/dojo/widget-core/blob/master/src/WidgetBase.ts#L263). It looks like, at some point, instead of storing property diff results in a single variable it was separated out into `diffPropertyResults` for the values, and `diffPropertyChangedKeys` for the keys.

matt-gadd: ha weird!

README.md
```diff
@@ -200,17 +200,40 @@ Properties passed to the `w` function represent the public API for a widget.
...
+The `@diffProperties` decorator will provide a generic diff implementation for _all_ the properties on a widget. By default `diffProperties` uses `DiffType.AUTO` to compare properties:
 
-##### Custom property diff control
+```typescript
+@diffProperties // <-- uses DiffType.AUTO
+class MyWidget extends WidgetBase<WidgetProperties> {
+}
+```
+
+But this can be changed to any `DiffType`:
```
agubler: Do you think we should document the available `DiffType`s?

src/WidgetBase.ts
```diff
@@ -71,6 +69,23 @@ export function diffProperty(propertyName: string, diffType = DiffType.CUSTOM, d
 }
 
 /**
+ * Decorator that can be used to enable property diffing of all properties
+ */
+export function diffProperties(constructor: Function): void;
```
agubler: Do we need this? Is it unreasonable to have the consumers *always* call `@diffProperties()`?

src/WidgetBase.ts
```diff
@@ -271,17 +285,42 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 			delete this._previousProperties[propertyName];
 		});
 
-		const diffPropertiesResult = this.diffProperties(this._previousProperties, properties);
-		this._properties = assign(diffPropertiesResult.properties, diffPropertyResults);
+		const diffPropertiesDecorators = this.getDecorator('diffProperties');
+		if (diffPropertiesDecorators.length) {
+			const diffDecorator = diffPropertiesDecorators[diffPropertiesDecorators.length - 1];
```
agubler: Is this to deal with `@diffProperties` being added twice? That interesting re extended widgets that the behaviour for diffProperties added by the super class can be "overridden".

matt-gadd: hmmm in my mind the base class `diffProperties` should always take precedence?

README.md
rorticus: Yes! I assumed it was documented already, but that might have been a poor assumption 😄

matt-gadd: thats my bad, I didn't document them in my PR.

src/WidgetBase.ts
rorticus: I think it is unreasonable. It makes for some strange looking code, but it feels like strange to have users to use `()` in this one case, just because we made the DiffType optional. `@diffProperties` feels a lot more natural to me than `@diffProperties()`

matt-gadd: Either way we should probably just decide a style, I know this is overloaded but with no params I'd imagine users would expect it to be like `@afterRender` and `onPropertiesChanged@`, which don't take parenthesis.

agubler: However with the changes for enhanced decorators we are going to change all decorators to accept an optional function - https://github.com/dojo/widget-core/issues/402

src/WidgetBase.ts
rorticus: Right, I wanted to take the "last" diffProperties, so it can be overridden. My thinking was if you had extended a widget, that already had a `@diffProperties`, but you added a new property that had special logic, you could just give your own `@diffProperties` as well. Now that I type that out though it doesn't make sense because you could just use a single `@diffProperty` ...

rorticus: You want me to always take the one in the base class?

agubler: Yes I think that makes the most sense, in the extended widget if you need a different behaviour for a property then as you say you can use `diffProperty`.

It does raise the question about diffing for a specific property, can they be overridden? Are they overridden? or will both diffs run?

matt-gadd: in my head it seems the most logical at least. when i write a widget i'm essentially stamping the diff behaviour with that decorator. If someone writes a mixin with additive behaviour they can still use the granular `@diffProperty`. It does bring up a valid other point in relation to my previous PR though. What should we do in the case of a base class using `@diffProperty` and a class extending it overriding it? im not sure of the behaviour.

matt-gadd: @agubler beat me to it.

PR #429, opened 2017-03-29
Adding a render:rendered event, issue #409

### LINKED ISSUE #409: Add a `render:rendered` event?
**Enhancement**

I have a use case where I want to *observe* what a widget actually renders.  Currently widgets emit an `render:scheduled` event, but there is no "formal" way of observing what actually gets rendered when the `__render__()` is called.

It would be good if we had something like:

```typescript
export interface RenderEventObject<W extends WidgetBase> extends EventTargettedObject<W> {
	result: VNode | string | undefined;
}
```

I also question the need of the `render:scheduled` at the moment.  I know I had originally put it in because it was something that we were struggling with getting working, but now the way the lifecycle works, it seems unnecessary.

rorticus:
**Type:** feature

**Description:**

Adding a `render:rendered` event that is emitted with the final result of the `__render__` method.

```typescript
widget.on('render:rendered', event => {
    console.log('rendered ', event.result);
});
```

Note that this will get emitted whether or not the result was cached. Calling `__render__` will **always** emit this event.

Resolves #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,16 @@ w('my-widget', properties, children);
 The example above that uses a string for the `widgetConstructor `, is taking advantage of our [widget registry](#widget-registry) functionality.
 The widget registry allows for the lazy instantiation of widgets.
 
+### Responding to Render Events
+
+Sometimes you may want to be notified when a widget is rendered. Widgets will emit a `render:rendered` event with the result of the render operation.
+
+```typescript
+widget.on('render:rendered', event => {
+    // handle render event
+});
+```
+
 ### Writing Custom Widgets
 
 The `WidgetBase` class provides the functionality needed to create Custom Widgets.
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -17,7 +17,8 @@ import {
 	PropertyChangeRecord,
 	PropertiesChangeRecord,
 	PropertiesChangeEvent,
-	HNode
+	HNode,
+	RenderRenderedEvent
 } from './interfaces';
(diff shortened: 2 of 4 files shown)


src/WidgetBase.ts
```diff
@@ -350,8 +352,21 @@ export class WidgetBase<P extends WidgetProperties> extends Evented implements W
 			if (widget) {
 				this._cachedVNode = widget;
 			}
+
+			this.emit({
```
kitsonk: This highlights that this block could likely be written so there is only a single return statement (and a single emit).

rorticus: Good call. Cleared up to a single emit now!

README.md
```diff
@@ -172,6 +172,16 @@ w('my-widget', properties, children);
 The example above that uses a string for the `widgetConstructor `, is taking advantage of our [widget registry](#widget-registry) functionality.
 The widget registry allows for the lazy instantiation of widgets.
 
+### Responding to Render Events
+
+Sometimes you may want to be notified when a widget is rendered. Widgets will emit a `render:rendered` event with the result of the render operation.
```
dylans: This is a bit wishy-washy... we should either say why you might want to be notified, or just state something like "Widgets will emit a `render:rendered` event with the result of the render operation. Listen to this event to be notified."

agubler:
I'm struggling to understand the use case for this other than perhaps testing and I don't think we should ever be exposing `VNode`s.

kitsonk:
I am afraid I let us jump the gun on this one.  We should deliver #430 and any change along this line we should address later.

PR #431, opened 2017-03-30
Removing render:scheduled event

### LINKED ISSUE #430: Remove `render:scheduled` from Projector
**Enhancement**

From #409:

>  I also question the need of the`render:scheduled` at the moment. I know I had originally put it in because it was something that we were struggling with getting working, but now the way the lifecycle works, it seems unnecessary.

Further discussion with @matt-gadd and @agubler we agreed it should be removed.

rorticus:
**Type:** feature

**Description:**

Removing `render:scheduled` event from the projector. Since it was only used for testing, we can get away with reproducing the functionality with subclasses.

Resolves #430

### DIFF
--- a/tests/unit/mixins/Projector.ts
+++ b/tests/unit/mixins/Projector.ts
@@ -148,26 +148,13 @@ registerSuite({
 		assert.equal(projector.root, root);
 	},
 	'pause'() {
-
-		let called = false;
-
-		class RenderCapturingWidget extends TestWidget {
-			public scheduleRender() {
-				super.scheduleRender();
-
-				if ((<any> this)._scheduled) {
-					called = true;
-				}
-			}
-		}
-
-		const projector = new RenderCapturingWidget();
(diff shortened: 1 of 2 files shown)


tests/unit/mixins/Projector.ts
```diff
@@ -167,11 +167,20 @@ registerSuite({
...
 		let called = false;
-		projector.on('render:scheduled', () => {
-			called = true;
-		});
+
+		class RenderCapturingWidget extends TestWidget {
+			public scheduleRender() {
+				super.scheduleRender();
+
+				if ((<any> this)._scheduled) {
```
agubler: Could you spy on `global.requestAnimationFrame()` being called rather than accessing a private?

agubler:
@rorticus there are conflicts in this PR now

rorticus:
@agubler conflicts resolved!

PR #453, opened 2017-04-18
RegistryHandler abstraction

matt-gadd:
**Type:** feature

**Description:**
This PR makes management and usage of Registries a little easier. The PR adds a few things:

* Support for arbitrary amount of registries, which are abstracted on usage (before we only supported a global and local registry, but then allowed overriding the local via a property using the `RegistryMixin`).
* `Registry.get()` no longer ever returns a promise, it either returns the widget or null.
* We no longer manage waiting on a registry item in `WidgetBase`, it's all dealt with by the `RegistryHandler`.
* You can now use before defining when using a widget from the `Registry` and the containing widget will re-render with it once it is loaded. This aligns more with Custom Elements.

With the above changes we now have a pattern for dealing with Widgets that haven't been loaded yet (over the default of rendering nothing until it's ready). For example:

```typescript
render() {
    const FooOrLoading = this.registries.get('foo') || LoadingSpinner;
    return w(FooOrLoading, {});
}
```

Remaining questions pertaining to this PR are:
* Are multiple registries (over the existing global and local) useful?
* Should we provide the user to be able to get from a specific registry? ie: at the moment the registries are resolved where local wins. It seems in quite a lot of scenario's where we'd take a registry as a property that we'd want that to win?

### DIFF
--- a/src/mixins/Registry.ts
+++ b/src/mixins/Registry.ts
@@ -1,9 +1,7 @@
-import { includes } from '@dojo/shim/array';
 import WidgetRegistry from '../WidgetRegistry';
-import { WidgetBase, onPropertiesChanged, diffProperty } from './../WidgetBase';
+import { WidgetBase, diffProperty } from './../WidgetBase';
 import {
 	PropertyChangeRecord,
-	PropertiesChangeEvent,
 	Constructor,
 	WidgetProperties
 } from '../interfaces';
@@ -17,17 +15,19 @@ export function RegistryMixin<T extends Constructor<WidgetBase<RegistryMixinProp
 
 		@diffProperty('registry')
 		public diffPropertyRegistry(previousValue: WidgetRegistry, value: WidgetRegistry): PropertyChangeRecord {
+			let changed = false;
+			if (!previousValue) {
+				this.registries.add(value);
+				changed = true;
+			}
+			else if (previousValue !== value) {
+				this.registries.replace(previousValue, value);
+				changed = true;
+			}
 			return {
-				changed: previousValue !== value,
+				changed,
 				value: value
 			};
-		}
-
-		@onPropertiesChanged()
-		protected onPropertiesChanged(evt: PropertiesChangeEvent<this, RegistryMixinProperties>) {
-			if (includes(evt.changedPropertyKeys, 'registry')) {
-				this.registry = evt.properties.registry;
-			}
 		}
 	};
 	return Registry;
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,0 +1,57 @@
+import { Evented } from '@dojo/core/Evented';
+import { WidgetConstructor } from './interfaces';
+import WidgetRegistry from './WidgetRegistry';
+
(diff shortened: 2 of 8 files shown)


agubler:
I have removed the `WIP` tag as I think these changes can go in separately to the any additional changes to address the concerns documented in the description. I will open an issue for these.

src/mixins/Registry.ts
```diff
@@ -17,18 +15,20 @@ export function RegistryMixin<T extends Constructor<WidgetBase<RegistryMixinProp
...
 		public diffPropertyRegistry(previousValue: WidgetRegistry, value: WidgetRegistry): PropertyChangeRecord {
+			let changed = false;
+			if (!previousValue) {
+				this.registries.add(value);
+				changed = true;
+			}
+			else if (previousValue !== value) {
+				this.registries.replace(previousValue, value);
+				changed = true;
+			}
```
agubler: Is there a branch here for comparing the length of registries where the previous and new that are the same instance?

dylans:
This probably needs an update to the readme as well.

agubler:
@dylan I'll add that into my outstanding readme update.

src/mixins/Registry.ts
agubler: A reply from @matt-gadd mentions that it's not required to compare lengths anymore as if a new item is added then it will invalidate any widget waiting for it. So no need to perform the check here.

agubler:
@dylans There is not much that needs to change in the readme, as the API for defining a widget has not changed. However we probably need a section on how to work with the registry in a widget and the registry mixin.

agubler:
Issue for readme updates: https://github.com/dojo/widget-core/issues/459

pottedmeat:
In dgrid, I've been using the concept of having a grid-scoped registry, and user-defined registry that can override the grid-scoped registry, or define widgets that will be used by configurable end-points like editable cells.

This new class suits what I'm doing really well, I'm not sure the best way to use it in the grid with the way it's integrated in `WidgetBase`:

* As far as I can tell, the `registries` property is typed `any` - what is the intended external interface
* `_registries` is created automatically in the `WidgetBase` constructor - so I'm assuming I shouldn't create my own
* `_registries` is private - how should I go about adding both the grid registry and the registry passed in my properties to this `RegistryHandler`?
* `_registries` is always created - even though all of the child widgets in the grid were/are able to use an identical instance of the registry

matt-gadd:
@pottedmeat I have fixed the typings in https://github.com/dojo/widget-core/pull/471, thanks for the heads up. You should be able to take a user defined registry via a property using the `Registry` mixin the same as before (it was updated in this PR)

pottedmeat:
Yeah, now that I see the typing it makes more sense.

Is there any way to prevent the RegistryHandler from being continuously re-wrapped? It should be identical in all grid child widgets.

matt-gadd:
@pottedmeat Not currently no, they are meant to be per widget. You can add additional registries locally so sharing the `RegistryHandler` would bleed beyond the boundary of the widget itself. Only the handler is created on construction (not the registries themselves).

PR #455, opened 2017-04-19
Readme updates

agubler:
**Type:** feature

**Description:**

README.md updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ We also provide a suite of pre-built widgets to use in your applications: [(@doj
         - [Projector](#projector)
         - [Event Handling](#event-handling)
         - [Widget Registry](#widget-registry)
+        - [Injecting State](#injecting-state)
         - [Theming](#theming)
         - [Internationalization](#internationalization-i18n)
         - [Web Components](#web-components)
@@ -49,7 +50,6 @@ npm install @dojo/widget-core
 npm install @dojo/has
 npm install @dojo/shim
 npm install @dojo/core
-npm install @dojo/compose
 npm install @dojo/i18n
 npm install maquette
 ```
@@ -64,12 +64,12 @@ The smallest `@dojo/widget-core` example looks like this:
 ```ts
 class MyWidget extends WidgetBase<WidgetProperties> {
 	render() {
-       return v('h1', { innerHTML: 'Hello, Dojo!' });
+       return v('h1', [ 'Hello, Dojo!' ]);
 	}
 }
 
 const Projector = ProjectorMixin(MyWidget);
-const projector = new Projector({});
+const projector = new Projector();
 
 projector.append(root);
 ```
@@ -117,36 +117,27 @@ The following code creates an element with the specified `tag`
 v(tag: string): HNode[];
 ```
 
-where `tag` is in the form: element.className(s)#id, e.g.
-
-```
-h2                  (produces <h2></h2>)
-h2.foo              (produces <h2 class="foo"></h2>)
-h2.foo.bar          (produces <h2 class="foo bar"></h2>)
-h2.foo.bar#baz      (produces <h2 class="foo bar" id="baz"></h2>)
-h2#baz              (produces <h2 id="baz"></h2>)
-```
-
-`classNames` should be delimited by a period (`.`).
-
-**Please note**, both the `classes` and `id` portions of the `tag` are optional.
-
-
 The following code renders an element with the `tag` and `children`.
 
 ```ts
-v(tag: string, children: (DNode | null)[]): HNode[];
+v(tag: string, children: DNode[]): HNode[];
 ```
 
 The following code renders an element with the `tag`, `properties` and `children`.
 
 ```ts
-v(tag: string, properties: VirtualDomProperties, children?: (DNode | null)[]): HNode[];
+v(tag: string, properties: VirtualDomProperties, children?: DNode[]): HNode[];
 ```
 
 As well as interacting with the VDOM by passing it HyperScript, you can also pass it Dojo 2 Widgets or Custom Widgets using the `w` function.
 
 #### `w`
+
+The following code creates a widget using the `widgetConstructor`.
+
+```ts
+w<P extends WidgetProperties>(widgetConstructor: string | WidgetBaseConstructor<P>): WNode[];
+```
 
 The following code creates a widget using the `widgetConstructor` and `properties`.
 
@@ -154,23 +145,33 @@ The following code creates a widget using the `widgetConstructor` and `propertie
 w<P extends WidgetProperties>(widgetConstructor: string | WidgetBaseConstructor<P>, properties: P): WNode[];
 ```
 
+The following code creates a widget using the `widgetConstructor` and `children`.
+
+```ts
+w<P extends WidgetProperties>(widgetConstructor: string | WidgetBaseConstructor<P>, children: DNode[]): WNode[];
+```
+
 The following code creates a widget using the `widgetConstructor`, `properties` and `children`
 
 ```ts
-w<P extends WidgetProperties>(widgetConstructor: string | WidgetBaseConstructor<P>, properties: P, children: (DNode | null)[]): WNode[];
+w<P extends WidgetProperties>(widgetConstructor: string | WidgetBaseConstructor<P>, properties: P, children: DNode[]): WNode[];
 ```
 Example `w` constructs:
 
 ```ts
+w(WidgetClass);
 w(WidgetClass, properties);
+w(WidgetClass, children);
 w(WidgetClass, properties, children);
 
+w('my-widget');
 w('my-widget', properties);
+w('my-widget', children);
 w('my-widget', properties, children);
 ```
 
-The example above that uses a string for the `widgetConstructor `, is taking advantage of our [widget registry](#widget-registry) functionality.
-The widget registry allows for the lazy instantiation of widgets.
+The example above that uses a string for the `widgetConstructor `, is taking advantage of the [widget registry](#widget-registry) functionality.
+The widget registry allows for the lazy loading of widgets.
 
 ### Writing Custom Widgets
 
@@ -184,13 +185,6 @@ import { WidgetBase } from '@dojo/widget-core/WidgetBase';
 ```
 
 **All** widgets should extend from this class.
-
-#### Public API
-
-|Function|Description|Default Behaviour|
-|---|---|---|
-|render|Returns the DNode structure of the widget|Returns a `div` HNode with children of `this.children`.
-|diffProperties|Diffs the current properties against the previous properties and returns an object with the changed keys and new properties|Performs a shallow comparison between the previous and current properties, copies the properties using `Object.assign` and returns the resulting `PropertiesChangeRecord`.|
 
 #### The 'properties' lifecycle
 
@@ -291,8 +285,6 @@ $widget->setProperties({
 
 Attaching a listener to the event is exposed via a decorator `@onPropertiesChanged`.
 
-*using the `onPropertiesChanged ` decorator*
-
 ```ts
 class MyWidget extends WidgetBase<WidgetProperties> {
 
@@ -303,7 +295,7 @@ class MyWidget extends WidgetBase<WidgetProperties> {
 }
 ```
 
-*registering the `onPropertiesChanged ` function in the constructor*
+For non decorator enviroments the listener can be registered using the `onPropertiesChanged ` function in the constructor.
 
 ```ts
 class MyWidget extends WidgetBase<WidgetProperties> {
@@ -319,7 +311,7 @@ class MyWidget extends WidgetBase<WidgetProperties> {
 }
 ```
 
-*Example event payload*
+Example event payload
 
 ```ts
 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,60 +117,61 @@ The following code creates an element with the specified `tag`
 v(tag: string): HNode[];
 ```
 
-where `tag` is in the form: element.className(s)#id, e.g.
```
rishson: Is this no longer supported?

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
 });
 ```
 
+#### Injecting State
+
+Working with larger widget structures, it can quickly become tiresome and complex to pass all the required properties down the tree. Additionally having to do so often means that widgets are forced to be aware of properties that they they don't need, but are required to know about in order to ensure they are propagated to the children.
```
rishson: I think we need a section upfront that states all widgets are linked in a tree. I know its obvious, but it should still be stated.

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+
+Dojo 2 provides a mechanism to inject state directly to parts of the widget structure; this is done by defining an `Injector` in the `registry` and passing a context object that will source the state that is to be injected.
```
rishson: 'widget structure' -> widget tree, tree of widgets. Don't mind, but needs to be consistent with section above.

README.md
agubler: We don't use it in any of our widgets and it would not be considered a "best practice" to. So I think it's best to not remove it from the readme.

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+import { Injector, BaseInjector } from '@dojo/widget-core/Injector';
+
+const myStateContext = {
+	theme: 'solid'
+};
+
+registry.define('state', Injector(BaseInjector, myStateContext));
+```
+
+To use the injected state, create a `beforeRender` method that returns a render function which creates a `w` reference to the registry item like any other widget and pass the properties required by the `InjectorProperties` interface.
```
rishson: Maybe add an explanation of the above code, e.g. Here you can see we created a context called `myStateContext` and added a single state property `theme` to it, a gave this property the value of `solid`? or too tutorially?

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+			getChildren(context: any, children: DNode[]): DNode[] {
+				return [];
+			},
+			children
+		});
+	};
+}
+```
+
+This will inject the values of `myState` as properties to the widget, as the returned object from `getProperties` is mixed over the widgets' existing properties.
```
rishson: I don't see how the above code will inject state. Can we say how, even if its hidden and automatic?

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+			},
+			children
+		});
+	};
+}
+```
+
+This will inject the values of `myState` as properties to the widget, as the returned object from `getProperties` is mixed over the widgets' existing properties. 
+
+For convenience, Dojo 2 provides mixin called `Container` that will decorate a widget with the above `beforeRender` implementation. Using the `Container` mixin enables any view widget to have state injected without coupling the widget to always have state injected. This means the widget can also be used as a normal widget with properties being passed from it's parent.
```
rishson: view widget is a term that needs defining, e.g. aren't all widgets 'view widgets' but some can also be container widgets?

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+import { Container } from '@dojo/widget-core/Container';
+
+function getProperties(context: any, properties: any): any {
+	return context;
+}
+
+const MyViewWidgetContainer = Container(MyViewWidget, 'state', { getProperties });
+```
+
+**Note:** that both the `getProperties` and `getChildren` functions do not need to be provided, if the functions are not defined the default mappers will be used that return an empty object and an empty array respectively.
```
rishson: Should we have a section 'Default container implementation' that doesn't have the `getProperties` function, and then a section on 'Advanced container implementation' that is like the code below?

README.md
agubler: Perhaps, but `injecting state` is not the place for that (I don't know where that would actually go in the current structure).

README.md
agubler: The state will get injected as properties, as the returned object from `getProperties` is mixed over the existing `properties`, is that not enough to explain how?

README.md
agubler: Well view widgets is a pretty common term for a component that only has presentational concerns; I was trying to emphasise that if a consumer implements the `beforeRender` manually then they have coupled the widget to having state injected.

The `Container` is a way that a consumer can connect the injected state to an existing widgets without having to modify or couple the widget to having state injected.

For example we could take `Checkbox` from `@dojo/widgets` and "connect" that to some state but also use the `Checkbox` elsewhere as a "normal" Dojo 2 widget.

README.md
agubler: I don't think so, the default implementation is effectively a `noop`. Just wanted to indicate that you don't need to provide both mappers, if you only want to inject properties.

README.md
agubler: Perhaps a bit tutorially, happy to add something if there is a consensus.

README.md
rishson: 👍 If its still _possible_ though, then maybe add to the reference guide (when written).

README.md
rishson: Maybe in the `w` section? Something like:

```ts
[in Baz.ts]

protected render: DNode {
  return v(div, {}, [
    w(Foo, {}, [Qux]);
    w(Bar, {});
  ]);
}
```
So now you have created instances of `Foo` and `Bar` widgets from inside the `render` call of `Baz` widget. Because you passed a child `Qux` to the `w` call of `Foo`, you have created a tree of widgets like so:

```
div (Baz)
  |
  - Foo
  |   |
  |   - Qux
  |
  - Bar
```
Note that in the DOM, you will see the result of the `render` call for `Foo, Bar, and Qux`;

agubler: Do we really think we need to explain this in the technical readme?

rishson: No - OK. This should be added to the beta2 tutorial 'State containers and advanced property passing'.

agubler: I would imagine the fact that the widgets are a tree would be in one of the the starter tutorials?

README.md
```diff
@@ -303,7 +295,7 @@ class MyWidget extends WidgetBase<WidgetProperties> {
 }
 ```
 
-*registering the `onPropertiesChanged ` function in the constructor*
+For non decorator enviroments the listener can be registered using the `onPropertiesChanged ` function in the constructor.
```
dylans: enviroments -> environments

README.md
```diff
@@ -336,11 +328,51 @@ Finally once all the attached events have been processed, the properties lifecyc
 
 <!-- TODO: render lifecycle goes here -->
 
-##### AfterRender
+Occassionally, in a mixin or base widget class, it my be required to provide logic that needs to be executed before or after a widgets' `render` call. These lifecycle hooks are supported in `WidgetBase` and operate as before and after aspects.
```
dylans: widgets' -> widget's (unless you meant widgets to be both plural and possessive)

dylans: Is it clear in this context what before and after aspects are? Should we link to something about AOP style advice or ?

README.md
```diff
@@ -336,11 +328,51 @@ Finally once all the attached events have been processed, the properties lifecyc
...
-##### AfterRender
+Occassionally, in a mixin or base widget class, it my be required to provide logic that needs to be executed before or after a widgets' `render` call. These lifecycle hooks are supported in `WidgetBase` and operate as before and after aspects. 
+
+The functionality is provided by the `beforeRender` and `afterRender` decorators.
+
+***Note:*** Both the `beforeRender` and `afterRender` functions are executed in the order that they are specified from the super class up to the final class.
+
+##### BeforeRender
+
+The `beforeRender` call receives the widgets' `render` function, `properties` and `children` and is expected to return a function that satifies the `render` API. The `properties` and `children` are passed to enable them to be manipulated or decorated prior to the `render` being called.
```
dylans: widgets' -> widget's

README.md
```diff
@@ -336,11 +328,51 @@ Finally once all the attached events have been processed, the properties lifecyc
...
+
+The functionality is provided by the `beforeRender` and `afterRender` decorators.
+
+***Note:*** Both the `beforeRender` and `afterRender` functions are executed in the order that they are specified from the super class up to the final class.
+
+##### BeforeRender
+
+The `beforeRender` call receives the widgets' `render` function, `properties` and `children` and is expected to return a function that satifies the `render` API. The `properties` and `children` are passed to enable them to be manipulated or decorated prior to the `render` being called. 
+
+This is the only time in the widget lifecycle that exposes either of these attributes to be manipluted outside of the property system.
```
dylans: Maybe add a sentence or two on why, and why this is a good thing?

README.md
```diff
@@ -336,11 +328,51 @@ Finally once all the attached events have been processed, the properties lifecyc
...
+		properties.extraAttribute = 'foo';
+		// Return or replace the `render` function
+		return () => {
+			return v('my-replaced-attribute');
+		};
+	}
+}
+```
+
+And using the beforeRender function for non decorator environments:
```
dylans: `beforeRender`

README.md
```diff
@@ -336,11 +328,51 @@ Finally once all the attached events have been processed, the properties lifecyc
...
+		return renderFunc;
+	}
+}
+```
 
-Occassionally, in a mixin or base widget class, it my be required to provide logic that needs to be executed using the result of a widget's `render`. `WidgetBase` supports registering functions that opererate as an after aspect to the `render` function using a provided decorator `afterRender` or in non-decorator environments by calling the decorator method directly.
+##### AfterRender
 
-*Using the `afterRender` decorator*
+The `afterRender` call receives the returned `DNode`s from a widgets' `render` call, so that the nodes can decorated, manipulated or even swapped.
```
dylans: widgets' -> widget's

README.md
```diff
@@ -352,7 +384,7 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
 }
 ```
 
-*Without using decorators*
+And using the afterRender function for non decorator environments:
```
dylans: `afterRender`

README.md
dylans: "Additionally having to do so often means that widgets are forced to be aware of properties that they they don't need, but are required to know about in order to ensure they are propagated to the children." is a bit hard to follow, perhaps something like "Needing to frequently pass all required properties also requires widgets to be aware of properties that are only needed for ensuring their propagation to child widgets."

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+			getChildren(context: any, children: DNode[]): DNode[] {
+				return [];
+			},
+			children
+		});
+	};
+}
+```
+
+This will inject the values of `myState` as properties to the widget, as the returned object from `getProperties` is mixed over the widgets' existing properties.
```
dylans: widgets' -> widget's

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+			},
+			children
+		});
+	};
+}
+```
+
+This will inject the values of `myState` as properties to the widget, as the returned object from `getProperties` is mixed over the widgets' existing properties. 
+
+For convenience, Dojo 2 provides mixin called `Container` that will decorate a widget with the above `beforeRender` implementation. Using the `Container` mixin enables any view widget to have state injected without coupling the widget to always have state injected. This means the widget can also be used as a normal widget with properties being passed from it's parent.
```
dylans: "Dojo 2 provides mixin called `Container`" -> "Dojo 2 provides a mixin called `Container`"

dylans: it's -> its

README.md
```diff
@@ -479,6 +509,89 @@ registry.define('my-widget', () => {
...
+function getProperties(context: any, properties: any): any {
+	return context;
+}
+
+const MyViewWidgetContainer = Container(MyViewWidget, 'state', { getProperties });
+```
+
+**Note:** that both the `getProperties` and `getChildren` functions do not need to be provided, if the functions are not defined the default mappers will be used that return an empty object and an empty array respectively.
+
+There may be times when the default `BaseInjector` doesn't meet the all your needs. For example if the context contains a reference to an eventable instance, you may want to add an event listener in the Injector to perform some logic, perhaps invalidate the widget.
```
dylans: "doesn't meet the all your needs." -> doesn't fully meet your needs.""

dylans: Injector -> `Injector` ?

README.md
```diff
@@ -852,15 +965,15 @@ to the widget instance.
 
 These are some of the **important** principles to keep in mind when creating and using widgets:
 
-1. The widget *`__render__`* function should **never** be overridden
+1. The widgets' *`__render__`*, *`__setProperties__`*, *`__setChildren__`* functions should **never** be called or overridden
```
dylans: widgets' -> widget's

agubler:
@rishson @dylans I believe i've addressed or responded to all the feedback :)

PR #456, opened 2017-04-19
Lazily calculate class object when `get` function is called.

### LINKED ISSUE #454: Convert `this.classes` to lazily generate classes
**Enhancement**

Calling `this.classes` should only register the intent to use a class and not execute all the memorisation logic until the `.get()` function has been called. This will help to the exact classes applied to each node regardless of where the `this.classes()` call has been made in the render function. 

It will also help remove complexity in unit tests when using `@dojo/test-extras/harness` that currently requires the classes to called in the exact order as they are in the `render` rather than the just in the structure of the Nodes.

agubler:
**Type:** feature

**Description:**

The classes object will be calculated at the point that the `get` function is called not when `this.classes` is called.

```ts
render() {
    const preCalculatedClasses = this.classes(css.classTwo);

    return v('div', { classes: this.classes(css.classOne) }, [
        v('span', { classes: preCalculatedClasses })
    ]);
}
```

Currently this would result in the `div` HNode including the negating class for the `preCalculatedClasses` like `{ classTwo: false, classOne: true }`. This is because `preCalculatedClasses` was created with `this.classes` first, which is confusing given the order of the classes in the render structure.

With this change the calculation is as lazy as possible and the classes will be memorised in an expected order; meaning that the `div` will only contain `{ classOne: true }` and the `span` will contain the negating class that was used in the `div`, `{ classOne: false, classTwo: true }`. 

Resolves #454

### DIFF
--- a/src/mixins/Themeable.ts
+++ b/src/mixins/Themeable.ts
@@ -33,9 +33,13 @@ export interface ThemeableProperties extends WidgetProperties {
 export interface ClassesFunctionChain {
 	(): ClassNameFlags;
 	/**
-	 * The classes to be returned when get() is called
+	 * The theme classes to be returned when get() is called
 	 */
-	classes: ClassNameFlags;
+	classes: string[];
+	/**
+	 * The fixed classes to be returned when get() is called
+	 */
+	fixedClasses: string[];
 	/**
 	 * Function to pass fixed class names that bypass the theming
 	 * process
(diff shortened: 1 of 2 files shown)


src/mixins/Themeable.ts
```diff
@@ -175,7 +179,31 @@ export function ThemeableMixin<T extends Constructor<WidgetBase<ThemeablePropert
 				this.recalculateThemeClasses();
 			}
 
-			const appliedClasses = classNames
+			const themeable = this;
+			function classesGetter(this: ClassesFunctionChain) {
```
tomdye: can this be defined outside the `classes` function?

agubler: I don't think so, it's got a different scope and also needs to reference the themeable scope.

tomdye: agreed, diregard

PR #472, opened 2017-05-03
Support typing of widget children

### LINKED ISSUE #465: Allow generic typing for Children.

agubler:
**Type:** feature

**Description:**

Enhance the existing typings to enable the expected children of a widget to be typed; This includes the widget and the properties for the widget.

`WidgetBase` supports a second generic type for `children` and defaults to `DNode` but can be restricted to `WNode` (any widget or properties) or more specifically `WNode<MyChildWidget>`.

This change includes adding a default of `WidgetProperties` for the 1st generic argument, `P` in both `WidgetBase` and `WidgetBaseInterface`.

<p><details>
  <summary><b>Examples</b></summary>
  <div>

```ts
import { WidgetBase } from './WidgetBase';
import { WidgetProperties, WNode } from './interfaces';
import { w, v } from './d';

interface MyWidgetProperties extends WidgetProperties {
	foo: string;
}
interface MyOtherWidgetProperties extends WidgetProperties {
	baz: boolean;
}
class MyWidget extends WidgetBase<MyWidgetProperties> {}
class MyOtherWidget extends WidgetBase<MyOtherWidgetProperties> {}
class MyCustomisedWidget extends WidgetBase<WidgetProperties> {
	doSomthingWidgety() { }
}
class AnotherCustomisedWidget extends WidgetBase<WidgetProperties> {
	doSomthingWidgety() { }
}
class MyContainer extends WidgetBase {}
class MyWidgetContainer extends WidgetBase<WidgetProperties, WNode<MyWidget>> {}
class MyCustomisedWidgetContainer extends WidgetBase<WidgetProperties, WNode<MyCustomisedWidget>> {}

// Correctly reports the the properties for `MyWidget` have not been fulfilled.
w(MyContainer, {}, [
	v('div'),
	w(MyWidget, { })
]);

// The errors are now resolved due after providing the requried properties for `MyWidget`
w(MyContainer, {}, [
	w(MyWidget, { foo: '' }),
	v('div')
]);

// Correct reports errors as `MyOtherWidget` does not duck type to `MyWidget`
w(MyWidgetContainer, {}, [
	w(MyOtherWidget, { baz: true }),
	w(MyWidget, { foo: '' })
]);

// Errors resolved when all children match the correct duck type
w(MyWidgetContainer, {}, [
	w(MyWidget, { foo: '' })
]);

// Correct reports errors as `WidgetBase` does not duck type to `MyCustomisedWidget`, missing `doSomthingWidgety`.
w(MyCustomisedWidgetContainer, {}, [
	w(WidgetBase, { }),
	w(MyCustomisedWidget, { })
]);

// Errors resolved when all children match the correct duck type
w(MyCustomisedWidgetContainer, {}, [
	w(MyCustomisedWidget, { key: '' }),
	w(AnotherCustomisedWidget, {}),
	w<MyCustomisedWidget>('customised-widget', {}) // when using registry string, need to pass widget type.
]);
```
</div></details></p>

**IDE errors:**

typingexamples_ts_-_widget-core

Resolves #465

### DIFF
--- a/tests/unit/d.ts
+++ b/tests/unit/d.ts
@@ -1,7 +1,7 @@
 import * as registerSuite from 'intern!object';
 import * as assert from 'intern/chai!assert';
 import { assign } from '@dojo/core/lang';
-import { DNode, HNode, WidgetProperties } from '../../src/interfaces';
+import { DNode, HNode, WNode, WidgetProperties } from '../../src/interfaces';
 import { WidgetBase } from '../../src/WidgetBase';
 import { v, w, decorate, registry, WNODE, HNODE, isWNode, isHNode } from '../../src/d';
 import WidgetRegistry from './../../src/WidgetRegistry';
@@ -12,11 +12,21 @@ class TestFactoryRegistry extends WidgetRegistry {
 	}
 }
 
-interface TestProperties extends WidgetProperties {
-	mand: boolean;
+interface ChildProperties extends WidgetProperties {
+	myChildProperty: string;
 }
 
-class TestWidget extends WidgetBase<TestProperties> {
+interface TestProperties extends WidgetProperties {
+	required: boolean;
+}
+
+class TestChildWidget extends WidgetBase<ChildProperties> {
+	render() {
+		return v('div');
+	}
+}
+
+class TestWidget extends WidgetBase<TestProperties, WNode<TestChildWidget>> {
 	render() {
 		return v('outernode', { type: 'mytype' }, [
 			v('child-one'),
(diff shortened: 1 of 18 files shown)


tests/unit/d.ts
```diff
@@ -57,24 +67,42 @@ registerSuite({
...
+			assert.deepEqual(dNode.widgetConstructor, WidgetBase);
+			assert.deepEqual(dNode.properties, { id: 'id', classes: [ 'world' ]});
 			assert.lengthOf(dNode.children, 1);
 			assert.isTrue(isWNode(dNode));
 			assert.isFalse(isHNode(dNode));
 		},
 		'create WNode wrapper using label with properties and children'() {
-			const properties = { mand: false };
-			const dNode = w<TestProperties>('my-widget', properties, [ w(WidgetBase, properties) ]);
+			const properties: any = { id: 'id', classes: [ 'world' ] };
```
kitsonk: I assume `w()` no longer likes the inferred type?  Is this going to become frustrating/challenging, or do you think this won't cause too much `any` ugliness in real life?

tests/unit/mixins/Registry.ts
```diff
@@ -57,7 +57,7 @@ registerSuite({
 			class IntegrationTest extends TestWithRegistry {
 				render() {
 					return v('div', [
-						w('test', { id: `${Math.random()}` })
+						w<any>('test', { id: `${Math.random()}` })
```
kitsonk: When using the registry now, are we going to have to supply some sort of generic argument?

agubler: Yes if it's any more than the default (which is `WidgetProperties`), this was recommended (undocumented most likely) anyway previously as because otherwise you didn't get the type guards for the properties.

tests/unit/d.ts
agubler: The type was never inferred correctly for registry items, it just defaulted to `any`. To get the correct guarding the consumer will need to pass the interface or type for the widget.

tests/unit/mixins/Registry.ts
kitsonk: Have we run this against `todo-mvc` or `todo-mvc-kitchensink` to see if we get any compile errors (wanted or unwanted)?  Because an alternative is to have an overload with a 'string' that would not require a supplied generic like this.  I guess my feeling is that `<any>` seems like an odd generic to have to supply.  It is like telling TypeScript to ignore something it can reasonably ignore anyways and lends end developers to just use the `<any>` hammer everywhere to just keep TypeScript quiet, whether they are using the registry or not.

agubler: @kitsonk I always upgrade the downstream application(s), which will include the examples, as well as `widgets`. We wouldn't want to provide an overload that removed the generic entirely, because as i mentioned it recommended to actually tell the function the typing when referencing a registry item.

I am not recommending using `any`, it was really just a lazy change for the test. It should be the shape that provides the typings for the registry widget, so that you fulfil the property API and also conform to the new children API.

agubler: For reference these are the changes I needed to make for the examples: https://github.com/dojo/examples/pull/210

PR #476, opened 2017-05-05
Support for injecting a theme into Themeable widgets

### LINKED ISSUE #474: Theme Injection
**Enhancement** 

The need to pass the `theme` property to all of the widgets in the tree regardless of whether a widget actually requires the property (i.e. has had `ThemeableMixin` applied) is confusing, cumbersome and potentially impossible when dealing with external components.

To alleviate this pain it would be beneficial if a `theme` could be registered externally and injected into the themeable widgets before the `render` if the widget hasn't been explicitly passed a `theme` property.

The theme should be able to be updated invalidating all themeable widgets and the new theme classes reflected after a re-render.

agubler:
**Type:** feature

**Description:**

Adds support for globally defining the `theme` using a `Registry` and `ThemeInjector`. If the `theme` has been defined in the registry, it will inject the theme property into all themeable widgets that have not explicitly received a `theme` property.

When the `theme` is updated using the `.set` API of the theme context, all themeable widgets are invalidated.

Resolves #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -740,6 +740,38 @@ class MyThemeableWidget extends ThemeableMixin(WidgetBase)<MyThemeableWidgetProp
 
 The theme can be applied to individual widgets or to a project and property passed down to its children.
 
+##### Injecting a theme
+
+The theming system supports injecting a theme configured outside of the usual property passing down the widget tree. 
+
+This is configured using a `ThemeInjectorContext` that is passed to the `Injector` mixin along with the `ThemeInjector` class. Once the theme injector is defined in the registry the `theme` can be changed by called the `ThemeInjectorContext#set(theme: any)` API on the instance of the injector context.
+
+```ts
+// Create the singleton injector context
+const themeInjectorContext = new ThemeInjectorContext(myTheme);
+
+// Create the base ThemeInjector using the singleton context and the Injector mixin
+const ThemeInjectorBase = Injector<ThemeInjectorContext, Constructor<ThemeInjector>>(ThemeInjector, themeInjectorContext);
+
+// Define the created ThemeInjector against the static key exported from `Themeable`
+registry.define(INJECTED_THEME_KEY, ThemeInjectorBase);
+```
+
+Once this is defined any themeable widgets without an explicit `theme` property will be controlled via the theme set within the `themeInjectorContext`. To change the theme simple call `themeInjectorContext.set(myNewTheme)` and all widgets that are using the global theme will be updated to the new theme.
+
+Finally `Themeable` exports a helper function wraps the behaviour defined above and returns the context, with optional parameters for the `theme` and the `registry` to defined the injector in. If no `registry` is provided then the global registry is used.
+
+```ts
+// uses global registry
+const context = registerThemeInjector(myTheme);
+
+// setting the theme
+context.set(myNewTheme);
+
+// uses the user defined registry
+const context = registryThemeInjector(myTheme, myRegistry);
+```
+
 ##### Overriding Theme Classes
 
 As we are using `css-modules` to scope widget css classes, the generated class names cannot be used to target specific nodes and apply custom styling to them. Instead you must use the `extraClasses` property to pass your generated classes to the widget. This will only effect one instance of a widget and will be applied on top of, rather than instead of, theme classes.
--- a/README.md
+++ b/README.md
@@ -742,9 +742,9 @@ The theme can be applied to individual widgets or to a project and property pass
 
 ##### Injecting a theme
 
-The theming system supports injecting a theme that is configured outside of the usual property passing down the widget tree. 
+The theming system supports injecting a theme that is configured externally to the usual mechanism of passing properties down the widget tree.
 
-This is done using a `ThemeInjectorContext` instance that is passed to the `Injector` mixin along with the `ThemeInjector` class. Once the theme injector is defined in the registry the `theme` can be changed by called the `ThemeInjectorContext#set(theme: any)` API on the instance of the injector context.
+This is done using a `ThemeInjectorContext` instance that is passed to the `Injector` mixin along with the `ThemeInjector` class. Once the theme injector is defined in the registry, the `theme` can be changed by calling the `ThemeInjectorContext#set(theme: any)` API on the instance of the injector context.
 
 ```ts
 // Create the singleton injector context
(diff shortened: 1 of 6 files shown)


agubler:
Still needs a helper util to create and define the ThemeInjector in the global registry and an update to the widget readme.

src/mixins/Themeable.ts
```diff
@@ -1,8 +1,11 @@
 import { assign } from '@dojo/core/lang';
+import { Evented } from '@dojo/core/Evented';
 import { includes, find } from '@dojo/shim/array';
 import Map from '@dojo/shim/Map';
-import { Constructor, WidgetProperties, PropertiesChangeEvent } from './../interfaces';
-import { WidgetBase, onPropertiesChanged, handleDecorator } from './../WidgetBase';
+import { Constructor, DNode, WidgetProperties, PropertiesChangeEvent } from './../interfaces';
```
agubler: I generally do `./` because it seems to be a imprinted in my keyboard muscle memory. I don't mind standardising if it's something that our linter can detect.

src/Injector.ts
```diff
@@ -10,11 +10,31 @@ export interface GetChildren {
...
+	getProperties: GetProperties;
+	getChildren: GetChildren;
+}
+
+/**
+ * Default noop Mappers for the container.
+ */
+export const defaultMappers: Mappers = {
+	getProperties(inject: any, properties: any): any {
+		return {};
```
agubler: Do we have a standard on this? and it one really "better" than the other?

src/Injector.ts
```diff
@@ -48,9 +68,9 @@ export function Injector<C, T extends Constructor<BaseInjector<C>>>(Base: T, con
 			const {
 				render,
 				properties,
-				getProperties,
+				getProperties = defaultMappers.getProperties,
```
agubler: Roger that, kinda of a change outside the scope of this PR so missed it. The change is needed tho, this is the first real usage of the injector pattern (so found some small fixes were required.)

src/Injector.ts
kitsonk: `Object.create(null)` is better in situations where the object is truly only being used as a property map and there is no need to access the larger `Object` functionality, like `.hasOwnProperty()` or related.  It has an appreciable performance improvement in cases where the object has a lot of property lookups into it, as there is no chain to ascend when looking for a key.

As far as a standard, no, I don't think we have ever codified though every once in a while, when I see an empty object literal I wonder if `Object.create(null)` could/should be used.  If we were to codify it, it would be as above, when it is purely being used a strict map/hash.

README.md
```diff
@@ -694,6 +694,38 @@ class MyThemeableWidget extends ThemeableMixin(WidgetBase)<MyThemeableWidgetProp
 
 The theme can be applied to individual widgets or to a project and property passed down to its children.
 
+##### Injecting a theme
+
+The theming system supports injecting a theme that is configured outside of the usual property passing down the widget tree.
```
dylans: "The theming system supports injecting a theme that is configured outside of the usual property passing down the widget tree. " perhaps -> "The theming system supports injecting a theme that is configured externally to the usual mechanism of passing properties down the widget tree. "?

README.md
```diff
@@ -694,6 +694,38 @@ class MyThemeableWidget extends ThemeableMixin(WidgetBase)<MyThemeableWidgetProp
 
 The theme can be applied to individual widgets or to a project and property passed down to its children.
 
+##### Injecting a theme
+
+The theming system supports injecting a theme that is configured outside of the usual property passing down the widget tree. 
+
+This is done using a `ThemeInjectorContext` instance that is passed to the `Injector` mixin along with the `ThemeInjector` class. Once the theme injector is defined in the registry the `theme` can be changed by called the `ThemeInjectorContext#set(theme: any)` API on the instance of the injector context.
```
dylans: "in the registry the `theme` can" -> "in the registry, the `theme` can"

dylans: "by called the" -> "by calling the"

README.md
```diff
@@ -694,6 +694,38 @@ class MyThemeableWidget extends ThemeableMixin(WidgetBase)<MyThemeableWidgetProp
...
+const themeInjectorContext = new ThemeInjectorContext(myTheme);
+
+// Create the base ThemeInjector using the singleton context and the Injector mixin
+const ThemeInjectorBase = Injector<ThemeInjectorContext, Constructor<ThemeInjector>>(ThemeInjector, themeInjectorContext);
+
+// Define the created ThemeInjector against the static key exported from `Themeable`
+registry.define(INJECTED_THEME_KEY, ThemeInjectorBase);
+```
+
+Once this is defined any themeable widgets without an explicit `theme` property will be controlled via the theme set within the `themeInjectorContext`. To change the theme simply call `themeInjectorContext.set(myNewTheme);` and all widgets that are using the injected theme will be updated to the new theme.
```
dylans: "Once this is defined any themeable widgets" -> "Once this theme injector is defined, any themeable widgets"

README.md
```diff
@@ -694,6 +694,38 @@ class MyThemeableWidget extends ThemeableMixin(WidgetBase)<MyThemeableWidgetProp
...
+
+To make this even easier `Themeable` exports a helper function wraps the behaviour defined above and returns the context, with a parameter for the `theme` and an optional `registry` to that the injector will be defined in. If a `registry` is not provided then the global `registry` is used.
```
dylans: "To make this even easier `Themeable` exports" -> "To make this even easier, `Themeable` exports"

dylans: behaviour -> behavior

dylans: "to that the injector will be defined in" -> "so that the injector will be defined in ______" in what?

src/Injector.ts
```diff
@@ -10,11 +10,31 @@ export interface GetChildren {
...
+/**
+ * The binding mappers for properties and children
+ */
+export interface Mappers {
+	getProperties: GetProperties;
+	getChildren: GetChildren;
+}
+
+/**
+ * Default noop Mappers for the container.
```
dylans: either remove the punctuation here or add to the comment before.

src/mixins/Themeable.ts
```diff
@@ -128,6 +134,67 @@ function createThemeClassesLookup(classes: ThemeClasses[]): ClassNames {
...
 /**
+ * The class for the theme injector context, used to control the theme once the
+ * theme injector has been defined in the registry.
+ */
+export class ThemeInjectorContext extends Evented {
+
+	private _theme: any;
+
+	/**
+	 * @param theme optional theme to initialize the context with
```
dylans: remove with at end of comment?

src/mixins/Themeable.ts
```diff
@@ -128,6 +134,67 @@ function createThemeClassesLookup(classes: ThemeClasses[]): ClassNames {
...
+	/**
+	 * Return the current theme
+	 */
+	public get theme(): any {
+		return this._theme;
+	}
+}
+
+/**
+ * Custom ThemeInjector class that listens to the invalidate event
```
dylans: ThemeInjector -> `ThemeInjector` ?

dylans: invalidate -> `invalidate` ?

src/mixins/Themeable.ts
```diff
@@ -128,6 +134,67 @@ function createThemeClassesLookup(classes: ThemeClasses[]): ClassNames {
...
+ */
+export class ThemeInjector extends BaseInjector<ThemeInjectorContext> {
+	constructor(context: ThemeInjectorContext) {
+		super(context);
+		this.own(context.on('invalidate', this.invalidate.bind(this)));
+	}
+}
+
+/**
+ * Convience function that given a theme and an optional registry, the theme
```
dylans: Convience -> Convenience

dylans: "that given" -> either "that is given" or "that, given"

src/mixins/Themeable.ts
```diff
@@ -128,6 +134,67 @@ function createThemeClassesLookup(classes: ThemeClasses[]): ClassNames {
...
+ * injector is defined against the registry and the theme context return.
```
dylans: "registry and the theme context return" -> either "registry, and the theme context is returned" or "registry, returning the theme context" ?

README.md
agubler: It's actually the `registry` that is passed that injector is defined in

agubler:
Waiting on a couple of PRs before landing this functionality #501 and #500

PR #479, opened 2017-05-05
support symbol as a label for registries

### LINKED ISSUE #478: Add symbol support for the registry
**Enhancement** 

Add symbol support for the `label` of a registry item; this will ensure there are no clashes between internally defined registry items and user defined registry items.

agubler:
**Type:** feature

**Description:**

Add support to use a `Symbol` as a key for defining an item in the registry and therefore when using `w`.

Requires https://github.com/dojo/core/pull/330

Resolves #478

### DIFF
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,6 +1,6 @@
 import { Evented } from '@dojo/core/Evented';
-import { WidgetBaseConstructor } from './interfaces';
-import WidgetRegistry, { RegistryLabel } from './WidgetRegistry';
+import { WidgetBaseConstructor, RegistryLabel } from './interfaces';
+import WidgetRegistry from './WidgetRegistry';
 
 export default class RegistryHandler extends Evented {
 	private _registries: { handle?: any, registry: WidgetRegistry }[] = [];
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,6 +1,6 @@
 import { Evented } from '@dojo/core/Evented';
 import { WidgetBaseConstructor, RegistryLabel } from './interfaces';
-import WidgetRegistry, { getLoadedEventType } from './WidgetRegistry';
+import WidgetRegistry, { WidgetRegistryEventObject } from './WidgetRegistry';
 
 export default class RegistryHandler extends Evented {
 	private _registries: { handle?: any, registry: WidgetRegistry }[] = [];
@@ -45,10 +45,12 @@ export default class RegistryHandler extends Evented {
 				return item;
 			}
 			else if (!registryWrapper.handle) {
-				registryWrapper.handle = registryWrapper.registry.on(getLoadedEventType(widgetLabel), () => {
-					this.emit({ type: 'invalidate' });
-					registryWrapper.handle.destroy();
-					registryWrapper.handle = undefined;
+				registryWrapper.handle = registryWrapper.registry.on(widgetLabel, (event: WidgetRegistryEventObject) => {
+					if (event.action === 'loaded') {
+						this.emit({ type: 'invalidate' });
+						registryWrapper.handle.destroy();
+						registryWrapper.handle = undefined;
+					}
 				});
 			}
 		}
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,6 +1,6 @@
 import { Evented } from '@dojo/core/Evented';
 import { WidgetBaseConstructor } from './interfaces';
-import WidgetRegistry from './WidgetRegistry';
+import WidgetRegistry, { RegistryLabel } from './WidgetRegistry';
 
 export default class RegistryHandler extends Evented {
 	private _registries: { handle?: any, registry: WidgetRegistry }[] = [];
@@ -31,13 +31,13 @@ export default class RegistryHandler extends Evented {
 		});
 	}
 
-	has(widgetLabel: string): boolean {
+	has(widgetLabel: RegistryLabel): boolean {
 		return this._registries.some((registryWrapper) => {
 			return registryWrapper.registry.has(widgetLabel);
 		});
 	}
 
-	get(widgetLabel: string): WidgetBaseConstructor | null {
+	get(widgetLabel: RegistryLabel): WidgetBaseConstructor | null {
 		for (let i = 0; i < this._registries.length; i++) {
 			const registryWrapper = this._registries[i];
 			const item = registryWrapper.registry.get(widgetLabel);
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,6 +1,6 @@
 import { Evented } from '@dojo/core/Evented';
 import { WidgetBaseConstructor, RegistryLabel } from './interfaces';
-import WidgetRegistry from './WidgetRegistry';
+import WidgetRegistry, { getLoadedEventType } from './WidgetRegistry';
 
 export default class RegistryHandler extends Evented {
 	private _registries: { handle?: any, registry: WidgetRegistry }[] = [];
@@ -45,7 +45,7 @@ export default class RegistryHandler extends Evented {
 				return item;
 			}
 			else if (!registryWrapper.handle) {
-				registryWrapper.handle = registryWrapper.registry.on(`loaded:${widgetLabel}`, () => {
+				registryWrapper.handle = registryWrapper.registry.on(getLoadedEventType(widgetLabel), () => {
 					this.emit({ type: 'invalidate' });
 					registryWrapper.handle.destroy();
 					registryWrapper.handle = undefined;
--- a/src/WidgetRegistry.ts
+++ b/src/WidgetRegistry.ts
@@ -13,6 +13,8 @@ export type WidgetRegistryItem = WidgetBaseConstructor | Promise<WidgetBaseConst
  */
 export const WIDGET_BASE_TYPE = Symbol('Widget Base');
 
+export type RegistryLabel = string | Symbol;
+
 /**
  * Widget Registry Interface
  */
@@ -24,7 +26,7 @@ export interface WidgetRegistry {
 	 * @param widgetLabel The label of the widget to register
 	 * @param registryItem The registry item to define
 	 */
-	define(widgetLabel: string, registryItem: WidgetRegistryItem): void;
+	define(widgetLabel: RegistryLabel, registryItem: WidgetRegistryItem): void;
 
 	/**
 	 * Return a WidgetRegistryItem for the given label, null if an entry doesn't exist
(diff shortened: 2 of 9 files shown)


src/WidgetRegistry.ts
```diff
@@ -13,6 +13,8 @@ export type WidgetRegistryItem = WidgetBaseConstructor | Promise<WidgetBaseConst
  */
 export const WIDGET_BASE_TYPE = Symbol('Widget Base');
 
+export type RegistryLabel = string | Symbol;
```
dylans: Any reason to not have this live in interfaces?

src/d.ts
```diff
@@ -71,7 +71,7 @@ export const registry = new WidgetRegistry();
 /**
  * Wrapper function for calls to create a widget.
  */
-export function w<W extends WidgetBaseInterface>(widgetConstructor: Constructor<W> | string, properties: W['properties'], children: W['children'] = []): WNode<W> {
+export function w<W extends WidgetBaseInterface>(widgetConstructor: Constructor<W> | string | Symbol, properties: W['properties'], children: W['children'] = []): WNode<W> {
```
dylans: If RegistryLabel was an interface, would it make sense to import it and then replace `| string | Symbol` -> `RegistryLabel` ?

src/interfaces.d.ts
```diff
@@ -269,7 +269,7 @@ export interface WNode<W extends WidgetBaseInterface = DefaultWidgetBaseInterfac
 	/**
 	 * Constructor to create a widget or string constructor label
 	 */
-	widgetConstructor: Constructor<W> | string;
+	widgetConstructor: Constructor<W> | string | Symbol;
```
dylans: If `RegistryLabel` was defined within interfaces, could use that here?

src/WidgetRegistry.ts
agubler: Yeah I'll move it there, then both the comments below could be addressed with the type.

agubler:
Have just added unit tests to use a symbol label, they currently fail as the label is used to emit when a registry item is loaded and a Symbol cannot be converted to a string (which the event type expects).

Need to resolve this :)

src/WidgetRegistry.ts
```diff
@@ -53,6 +53,13 @@ export function isWidgetBaseConstructor(item: any): item is WidgetBaseConstructo
 	return Boolean(item && item._type === WIDGET_BASE_TYPE);
 }
 
+function getEventType(label: RegistryLabel): RegistryLabel {
```
kitsonk: How is this functionality supposed to work with Symbols?

src/WidgetRegistry.ts
```diff
@@ -61,15 +68,16 @@ export class WidgetRegistry extends Evented implements WidgetRegistry {
...
-	has(widgetLabel: string): boolean {
+	has(widgetLabel: RegistryLabel): boolean {
 		return this.registry.has(widgetLabel);
 	}
 
-	define(widgetLabel: string, item: WidgetRegistryItem): void {
+	define(widgetLabel: RegistryLabel, item: WidgetRegistryItem): void {
 		if (this.registry.has(widgetLabel)) {
-			throw new Error(`widget has already been registered for '${widgetLabel}'`);
+			const regsitryLabelDesc = typeof widgetLabel === 'string' ? widgetLabel : widgetLabel.toString();
```
kitsonk: not all the branches of this are being tested, but `'foo'.toString()` is valid syntax, so it maybe just easier to do that or use explicit coercion of `String(widgetLabel)`;

src/WidgetRegistry.ts
```diff
@@ -78,7 +86,7 @@ export class WidgetRegistry extends Evented implements WidgetRegistry {
 			item.then((widgetCtor) => {
 				this.registry.set(widgetLabel, widgetCtor);
 				this.emit({
-					type: `loaded:${widgetLabel}`
+					type: getEventType(widgetLabel)
```
kitsonk: As above, how would I register for an event that has a Symbol label and how would that be different than other possible events?  Usually `type` is some sort of "class" in an event and there is other stuff in the event that talks about the specifics.  Shouldn't it really be `target`?

src/WidgetRegistry.ts
```diff
@@ -113,7 +121,7 @@ export class WidgetRegistry extends Evented implements WidgetRegistry {
 		promise.then((widgetCtor) => {
 			this.registry.set(widgetLabel, widgetCtor);
 			this.emit({
-				type: `loaded:${widgetLabel}`
+				type: getEventType(widgetLabel)
```
kitsonk: Sort of a macro thing... you are emitting the same event in 3 locations in code.  That seems like there is some duplicate logic here.

tests/unit/d.ts
```diff
@@ -99,6 +99,18 @@ registerSuite({
...
+			const properties: any = { id: 'id', classes: [ 'world' ] };
+			const symbolLabel = Symbol();
+
+			const dNode = w(symbolLabel, properties, [ w(WidgetBase, properties) ]);
+			assert.equal(dNode.type, WNODE);
+			assert.strictEqual(dNode.widgetConstructor, symbolLabel);
+			assert.deepEqual(dNode.properties, { id: 'id', classes: [ 'world' ]});
+			assert.lengthOf(dNode.children, 1);
+			assert.isTrue(isWNode(dNode));
+			assert.isFalse(isHNode(dNode));
```
kitsonk: No listening for the loaded event for a Symbol label tests?
No checking of the negative case of re-defining of a label with a Symbol?

src/WidgetRegistry.ts
agubler: It just returns the the symbol?

src/WidgetRegistry.ts
agubler: I'd noticed that from the codecov plugin; Thanks for the suggestions.

tests/unit/d.ts
agubler: I don't think either of these tests would belong in `d.ts`, more likely tests for the actual `WidgetRegistry.ts`. I'll have a look into it.

src/WidgetRegistry.ts
agubler: It could be abstracted to a function that accepts the widgetLabel (and the instance if not a method on the class) but it didn't really seem worthwhile to create a function (as this was existing code), but I could if you think it is needed?

```
emitLoadedEvent(widgetLabel: RegistryLabel) {
    this.emit({
        type: getEventType(widgetLabel)
    });
}
```

src/WidgetRegistry.ts
agubler: The functionality here hasn't changed as part of these changes. 

The type `loaded:${widgetLabel}` is actually a "class" of an event in this instance. If it just emitted `widget:loaded` every single widget would need to keep a registry of widgets that it is waiting on being loaded and use this to determine if the target is one of them before invalidating! 

The registration for the loaded events is actually done in the `RegistryHandler` which once emitted invalidates the widget it's associated to and removes the handle for the listener.

I did notice though I'd missed a location that was still using `loaded:${widgetLabel}` for both `symbol` and `string` labels so have fixed that and added tests.

agubler:
@kitsonk thanks for the feedback it was very useful. I think I've addressed or responded to most of you comments.

tests/unit/d.ts
kitsonk: Ok, I was only looking at what test cases got added then paying attention to where they got added.

src/WidgetRegistry.ts
kitsonk: I understood the code, but it was more of a broader question of if there is `type: 'loaded:foo'` for a string but `type: @@foo` for a Symbol that it seemed like a bigger challenge.  I know it is a broader question than this specific code, but it seems like symbols don't quite fit neatly into the `WidgetRegistry` event model.

agubler: It emits just the unique symbol as the type which as unique should be fine? the registry handler registers a listener for the symbol also and everything works in the same way as simply emitting `loaded:foo`

kitsonk: Well and that is the more macro question, if the event type for a loaded event emitted from the widget registry for a string label of `foo` is `loaded:foo` but for `Symbol('foo')` is `@@foo` then we can only ever have one type of event for symbol labels from `WidgetRegistry` forever.  That was part of my argument that actually the string label being part of the event type felt more like other meta data on the event than an actual event type.  Like `{ type: 'loaded', label }`, but I wasn't sure of the downstream impacts of that.

agubler: That's not how it works though? Did you see the reasoning for having a unique event per "loading" widget? Each symbol creates its own event, so you can have `const foo = Symbol()` & `const bar = Symbol()` and use both as registry labels for a widget that is not "loaded" and all registry handlers that had called `.get(foo)` or `.get(bar)` will be told when that widget is loaded as the registry emits `.emit({ type: foo })` or `.emit({ type: bar });`?

kitsonk: I guess what I am pointing out is that there are semantics implied in `{ type: 'loaded:foo' }` that are missing from `{ type: Symbol('foo') }`.  With one, you are being explicit that the event is a loaded event, the other, it is simply implied because that is the only event emitted.  I can understand that it works, but I am just pointing out there is an inconsistency.  Furthering that logic if `{ type: 'foo' }` wouldn't make sense, why does `{ type: Symbol('foo') }` make sense?  You _could_ in the future have some other event type for a string label, but you _cannot_ have another event type for a symbol label.

agubler: Ah I think I get what you mean now, let me see what I can do to address that - it is the extra context that you can never provide with symbol events as it would always have to be assumed as `loaded`. So a string label could support `unloaded` (just an example) but we couldn't do that for symbols.

agubler: Updated to expose an `action` from the `WidgetRegistryEvent`, so the type will always be the symbol or the string label and use an action of `loaded` (for now)

src/RegistryHandler.ts
```diff
@@ -45,10 +45,12 @@ export default class RegistryHandler extends Evented {
 				return item;
 			}
 			else if (!registryWrapper.handle) {
-				registryWrapper.handle = registryWrapper.registry.on(getLoadedEventType(widgetLabel), () => {
-					this.emit({ type: 'invalidate' });
-					registryWrapper.handle.destroy();
-					registryWrapper.handle = undefined;
+				registryWrapper.handle = registryWrapper.registry.on(widgetLabel, (event: WidgetRegistryEventObject) => {
+					if (event.action === 'loaded') {
```
kitsonk: just missing a negative test case that `event.action !== 'loaded'` is a noop

agubler: Thanks for keeping me honest

PR #480, opened 2017-05-05
Increasing shim version, issue #477

### LINKED ISSUE #477: Functional tests are failing for custom elements
**Bug** 

The functional tests for custom elements are currently failing for ios9.1 and safari 10,

rorticus:
**Type:** bug

**Description:**

Hopeful fix for custom element functional tests not working in Safari 10 and iOS 9.1. Increasing the custom elements shim version was enough to get them to pass in Safari 10 locally, I'm having issues running them in saucelabs, so I'm hoping Travis can do it for me 😛 

Resolves #477

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/grunt": "0.4.*",
     "@types/jsdom": "2.0.*",
     "@types/sinon": "^1.16.31",
-    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
+    "@webcomponents/custom-elements": "^v1.0.0-rc.3",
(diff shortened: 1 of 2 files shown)


rorticus:
Problems fixed in Safari 10, but there is still a problem with iOS... I think it's related to the shim we're using.

agubler:
Browserstack doesn't support anything greater than `9.1` it seems and when I ran kitchen sink on an ios sim it failed to load the `native-shim.js`.

Is there any way we can just ignore these functionals for ios 9.1?

agubler:
incorporated into https://github.com/dojo/widget-core/pull/482

PR #484, opened 2017-05-09
WIP: Add ability to serialize and deserialize to the Projector

### LINKED ISSUE #485: Serialize and deserialize a projection
**Feature**

In support of server side rendering, we need a mechanism that can take a `Projector` and serialize the virtual DOM in a way that it can be restored in another projector.  This allows context switching of the virtual DOM.

Things this feature needs to do technically:

- Provide a plain object structure which does not contain references to functions, DOM node instances, or other non-plain objects.
- Be able to hydrate the maquette projection where upon the first "proper" render, maquette will not remove and re-add the whole DOM tree.
- Allow a subsequent render of the projector add listeners and other references that are generated by the "stood up" application.

kitsonk:
**Type:** feature

**Description:**

This feature adds the ability to make it easy to perform server side rendering.

It adds a method to the `Projector` named `.serialize()` which will return a sanitized `VNode` that is the root of the current projection.  This then can be easily converted to a JSON string, transferred to a different context, and then deserialized by passing it as an argument to one of the attach methods on a projector (`.append()`, `.merge()`, `.replace()`).  During the attachment, this `VNode` will be used instead of rendering the `Projector`.

Resolves #485

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -1,7 +1,8 @@
 import global from '@dojo/core/global';
 import { assign } from '@dojo/core/lang';
 import { Handle } from '@dojo/interfaces/core';
-import { dom, Projection, ProjectionOptions, VNode, VNodeProperties } from 'maquette';
+import { ProjectionOptions, VNode, VNodeProperties } from '@dojo/interfaces/vdom';
+import { dom, Projection } from 'maquette';
 import 'pepjs';
 import cssTransitions from '../animations/cssTransitions';
 import { Constructor, DNode, WidgetProperties } from './../interfaces';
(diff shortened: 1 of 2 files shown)


kitsonk:
Todo:

- [ ] - More testing
- [ ] - Formal test cases

kitsonk:
I was also thinking that maybe `.serialize()` should actually be `.toJSON()` and it would integrate better, making `JSON.stringify(projector)` work.

src/mixins/Projector.ts
```diff
@@ -122,12 +137,57 @@ const eventHandlers = [
 	'onsubmit'
 ];
 
+/**
+ * A guard that ensure that the `value` is just a plain object
```
dylans: ensure -> ensures

kitsonk:
This is a bit of a rabbit hole, not what we need for SSR.  Closing for now, but will save the work.

PR #487, opened 2017-05-09
Allow re-use of DOM when merging on Projector

### LINKED ISSUE #486: Re-use DOM when merging with the Projector
**Feature**

The Projector should allow DOM to be re-used when merging.

The Projector should walk the `VNode` that is being merged, setting the DOM nodes, if they are already present in the DOM.  This coupled with a similar feature in maquette will allow the DOM to be re-used instead of being dropped an re-created in order to support easier server side rendering or render context switching.

kitsonk:
**Type:** feature

**Description:**

This PR provides the ability to re-use the DOM when using `projector.merge()`.  This requires AFASSoftware/maquette#99.

When merging, it will automatically _hydrate_ the render with the DOM nodes.  It is obviously incumbent upon the developer to ensure that the DOM that is being merged follows the same structure of the virtual DOM.  If any of the DOM is missing, but described in the virtual DOM, maquette will create those DOM nodes.

An example.  It will initially render with the button disabled when it receives the HTML, but when the app loads and is merged, the button will become active.  In addition the text will change from `Hello!` to `Hello, Dojo World!`.  The DOM nodes will be reused:

**index.html**

```html
<!DOCTYPE html>
<html>
<head>
	<title>dojo-test-app</title>
</head>
<body>
	<p>Welcome to dojo-test-app</p>
	<div id="my-app">
		<button type="button" disabled="">Click Me</button>
		<span>Hello!</span>
	</div>
</body>
</html>
```

**main.ts**

```ts
import { ProjectorMixin, AttachType } from '@dojo/widget-core/mixins/Projector';
import HelloWorld from './widgets/HelloWorld';

const root = document.getElementById('my-app') as HTMLDivElement;

const Projector = ProjectorMixin(HelloWorld);

projector.merge(root);
console.log('Merged!');
```

**widgets/HelloWorld.ts**

```ts
import { WidgetBase } from '@dojo/widget-core/WidgetBase';
import { WidgetProperties } from '@dojo/widget-core/interfaces';
import { v } from '@dojo/widget-core/d';

export default class HelloWorld extends WidgetBase<WidgetProperties> {
	render() {
		return v('div', {
			key: 'one'
		}, [ v('button', {
			type: 'button',
			disabled: false,
			onclick() {
				console.log('I was clicked!')
			}
		}, [ 'Click Me' ]),
		v('span', [ 'Hello, Dojo World!' ])]);
	}
}
```

Resolves #486

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/grunt": "0.4.*",
     "@types/jsdom": "2.0.*",
     "@types/sinon": "^1.16.31",
-    "@webcomponents/custom-elements": "^1.0.0-alpha.3",
+    "@webcomponents/custom-elements": "^v1.0.0-rc.3",
     "codecov.io": "0.1.6",
     "glob": "^7.0.6",
     "grunt": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -29,8 +29,7 @@
     "@dojo/core": "beta1",
     "@dojo/has": "beta1",
     "@dojo/i18n": "beta1",
-    "@dojo/shim": "beta1",
-    "maquette": "2.5.0"
+    "@dojo/shim": "beta1"
   },
   "devDependencies": {
     "@dojo/interfaces": "beta1",
@@ -54,6 +53,7 @@
     "typescript": "~2.3.2"
   },
   "dependencies": {
+    "maquette": "~2.5.0",
     "pepjs": "^0.4.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "@dojo/has": "beta1",
     "@dojo/i18n": "beta1",
     "@dojo/shim": "beta1",
-    "maquette": "2.4.3"
+    "maquette": "2.5.0"
   },
   "devDependencies": {
     "@dojo/interfaces": "beta1",
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -123,16 +123,15 @@ const eventHandlers = [
 	'onsubmit'
 ];
 
-function setDomNodes(vnode: VNode, root: Node | null = null) {
-	vnode.domNode = root;
-	if (vnode.children && root) {
-		const childNodes: Node[] = [];
-		for (let i = 0; i < root.childNodes.length; i++) {
-			if (root.childNodes[i].nodeType === 1) {
-				childNodes.push(root.childNodes[i]);
-			}
-		}
-		vnode.children.forEach((child, i) => setDomNodes(child, childNodes[i]));
+/**
+ * Internal function that maps existing DOM Elements to virtual DOM nodes.
+ * @param vnode The virtual DOM node
+ * @param domNode The Element, if any, to set on the virtual DOM node
+ */
+function setDomNodes(vnode: VNode, domNode: Element | null = null) {
+	vnode.domNode = domNode;
+	if (vnode.children && domNode) {
+		vnode.children.forEach((child, i) => setDomNodes(child, domNode.children[i]));
 	}
 }
 
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -47,6 +47,10 @@ export interface ProjectorMixin<P extends WidgetProperties> {
 
 	/**
 	 * Merge the projector onto the root.
+	 *
+	 * The `root` and any of its `children` will be re-used.  Any excess DOM nodes will be ignored and any missing DOM nodes
+	 * will be created.
+	 * @param root The root element that the root virtual DOM node will be merged with.  Defaults to `document.body`.
 	 */
 	merge(root?: Element): Handle;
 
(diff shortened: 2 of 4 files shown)


kitsonk:
This PR is now correctly failing as it needs the upstream maquette PR to function correctly.

johan-gorter:
Hello @kitsonk, is this method of reusing DOM nodes the most desired way for DOJO or are you also still also investigating other options? I would not like to have a feature in maquette that is not used by anyone.

kitsonk:
@johan-gorter I understand the concern.

We are fairly confident that for server-side rendering and progressive enhancement that not dropping the DOM and re-creating it will provide a better end-user experience.  While we don't have any empirical tests the prove there are flashes and the like as a large DOM restructure gets replaced with something almost identical, it is very logical it can and would happen.  By "assuming" the DOM structure, no matter how large, it is unlikely the user would have a negative experience as the application is progressively enhanced.

johan-gorter:
I do not think the screen will flash needlessly, but rather that recreating the DOM nodes could cause a performance penalty. Also (inner) scroll positions, focus and changes to <input>s would get lost. I think the approach of providing domNodes using VNode.domNode makes sense. I will merge it in.

johan-gorter:
Maquette version 2.5.0 is ready

kitsonk:
@johan-gorter thank you very much!

package.json
```diff
@@ -30,7 +30,7 @@
     "@dojo/has": "beta1",
     "@dojo/i18n": "beta1",
     "@dojo/shim": "beta1",
-    "maquette": "2.4.3"
+    "maquette": "2.5.0"
```
agubler: I know it was before, but do you think we should keep this pinned? or maybe use `~` or `^`?

kitsonk: We had had specific issues before with some compatibility.  Actually between `2.4.3` and `2.5.0`, there was a regression to some of our tests (#492).  It is risky to float it because it is such a fundamental dependency.

agubler: It just makes changes like this breaking as consumers will need to upgrade their Maquette version. I'd be happier if we actually specified Maquette as a real dependency (as it is), so consumers would just get the version of Maquette that is compatible with the version of `widget-core`

kitsonk: I think for `widget-core` we could make it a dependency.

Sort of a side note, how would `dojo build` resolve multiple versions of `maquette`?  For example if someone has it specified as one version in the `package.json` but includes `widget-core` so you have something like:

```
root
└ node_modules
 ├ @dojo
 │ └ widget-core
 │  └ node_modules
 │    └ maquette@2.5.0
 └ maquette@2.4.3

Comment (mwistrand) [inline package.json]: @kitsonk Regarding resolving multiple `maquette` versions, would this scenario ever happen since `maquette` is a peer dependency of `widget-core` rather than a dependency? Either way, webpack would know that `@dojo/widget-core/node_modules/maquette/dist/maquette.js` is different from `root/node_modules/maquette/dist/maquette.js` and use different integer mids for each.

Comment (kitsonk) [inline package.json]: @mwistrand but @agubler is suggesting making it a dependency, which raised the question in my mind of how that would be resolved with WebPack.  So the only case it would be _caveat emptor_ then would be under and AMD loader.  I guess considering that is the case, and we feel WebPack is going to be our primary end user environment, we can get rid of `peerDependencies` all together.  The main reason for it was to _do no harm_ when loading in a browser.  The thing we likely would need though is something that warned end users that duplicate packages were being included because semver dependencies could not be resolved, so people aren't scratching their head why their bundles are so big.

File: tests/unit/mixins/Projector.ts [hunk 2]
```diff
@@ -126,18 +126,96 @@ registerSuite({
...
+
+				assert.strictEqual(select.value, 'foo', 'foo should be selected');
+				assert.strictEqual(select.children.length, 3, 'should have 3 children');
+
+				assert.isFalse(onchangeListener.called, 'onchangeListener should not have been called');
+				assert.isFalse(onclickListener.called, 'onclickListener should not have been called');
+
+				const changeEvent = document.createEvent('Event');
+				changeEvent.initEvent('change', true, true);
+				select.onchange(changeEvent); // firefox doesn't like to dispatch this event
```
dylans: Maybe a reference to the cause of this issue, e.g. a known WebDriver defect

src/mixins/Projector.ts
```diff
@@ -122,6 +123,18 @@ const eventHandlers = [
...
 ];
 
+/**
+ * Internal function that maps existing DOM Elements to virtual DOM nodes.
+ * @param vnode The virtual DOM node
+ * @param domNode The Element, if any, to set on the virtual DOM node
+ */
+function setDomNodes(vnode: VNode, domNode: Element | null = null) {
+	vnode.domNode = domNode;
+	if (vnode.children && domNode) {
```
dylans: I'm not sure I get this check... meaning how does vnode.children exist and domNode not exist? It might be worth explaining this logic a bit better as I imagine it not being clear in 6 months.

tests/unit/mixins/Projector.ts
kitsonk: It isn't a WebDriver defect, it is a difference of implementation of browsers.  Contrary to popular belief, there are inconsistencies in implementation between browsers.  I suspect it has to do with onchange in firefox only occurring when you blur the element (which is different than other browsers) but it could also be firefox's trust issue with generated events.  Both of which weren't material to this test case, so I just moved on and noted what someone might consider an oddity.  If it were material to the test case, I would have expected more detailed information, or if it were a code path in the production code.

src/mixins/Projector.ts
kitsonk: The code is written in a way to deal with subsuming or not subsuming an existing DOM structure.  When merging, it is attempted to stitch together the existing DOM nodes, but if for some reason there is part of the DOM missing, it would result in a run time error.  Any missing DOM will be created by maquette.

tests/unit/mixins/Projector.ts
dylans: fair enough.

src/mixins/Projector.ts
dylans: Ok.

PR #491, opened 2017-05-10
updated typedoc to use TypeScript 2.3.2. 

devpaul:
**Type:** feature

**Description:**

* updated typedoc to use TypeScript 2.3.2. 
* Removed unused automated deployment files.

Improves the output greatly:

```
grunt api
Running "api:html" (api) task
Building API Documentation for "/Users/pshannon/src/dojo2/widget-core" to "_apidoc"

Using TypeScript 2.3.2 from /Users/pshannon/src/dojo2/widget-core/node_modules/typescript/lib
Warning: Unknown language: "tsx"
Rendering [========================================] 100%

Documentation generated at /Users/pshannon/src/dojo2/widget-core/_apidoc
```

I also removed files and settings related to the currently unused automation. It can be reinstated using grunt-dojo2-extras' `setupDeploy` task.

Related to dojo/grunt-dojo2-extras#23


kitsonk:
My concern is that we created `grunt-dojo2-extras` for things that weren't going to be part of our package build.  Stuff that was for non-npm published packages like `dojo/dojoio`, but we are now dragging it into our packages as a dependency.  I would really rather see this in `grunt-dojo2` than add another dependency.  I assumed this would apply to all npm published `@dojo` packages that we use `grunt-dojo2` with eventually?

Also, I don't like unpublished modules being merged into master.  If TypeDoc isn't published yet and we haven't verified the veracity of how things like default generics are being displayed, we shouldn't be merging this into master.

devpaul:
It sounds like we should put this on hold until typedoc publishes a release with TS 2.3.2. When that becomes available I'll add grunt-dojo2-extras as a dependency to grunt-dojo2 and remove the old (stale) typedoc task from grunt-dojo2.

PR #494, opened 2017-05-10
Adjust behaviour of invalidate during property and render phases

matt-gadd:
**Type:** bug

**Description:**
Before, calling `invalidate` in either render or property phases (ie `diffProperty`), could cause undesirable behaviour (including infinite loops). We now guard against this by:

a) an `invalidate` during a render phase becomes a no-op.
b) an `invalidate` during the properties phase will mark the widget as dirty, but will not emit the invalidate the `invalidated` event up the tree.

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -31,6 +31,12 @@ interface WidgetCacheWrapper {
 	used: boolean;
 }
 
+enum WidgetRenderState {
+	IDLE = 1,
+	PROPERTIES,
+	RENDER
+}
+
 /**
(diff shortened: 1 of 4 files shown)


agubler:
looks good, perhaps we should add something to the readme stating that widgets shouldn't be invalidated during a render and to guard users we ignore them?

matt-gadd:
there are currently no `invalidation` notes in the readme, so perhaps we should raise that as a separate issue.

agubler:
Yup that sounds reasonable.

PR #497, opened 2017-05-11
Add .sandbox() and .toHtml() to Projector

### LINKED ISSUE #493: Export HTML from Projector
**Feature**

In supporting the ability to do server side render, developers need an easy way to "grab" HTML which can be used to pre-render some DOM in one context and shift it to another context.  In order to facilitate that, we should add two methods to the Projector:

- `.toHtml()` - a method that will provide a text string of HTML that represents the currently rendered virtual DOM.
- `.sandbox()` - a peer method of `.attach()`, `.merge()`, and `.replace()` which creates a sandboxed root for the virtual DOM to render against that is not in the flow of the global `document` (and does not require a global `document` to be available.  Intended for server side rendering on Node.js, it still could be used in other environments to do rendering of virtual DOM without it impacting the global `document`.

From a usage perspective, it would look something like this:

```ts
import ProjectorMixin from '@dojo/widget-core/mixins/Projector';
import App from './widgets/App';

const Projector = ProjectorMixin(App);

projector.sandbox();
const html = projector.toHtml();
/* do something with the html */
```

Considerations:

- The current value display in `<select>` does not always represent the `<option selected>`.  When exporting HTML, if the markup originally had `<option selected>` the actual current visible value may not be accurate.  We should consider updating the exported HTML to represent that.
- There maybe other attributes which are disconnected from the actual visual state of the DOM.
- When running sandboxed, we should consider updating the renders synchronously.  This would ensure that whenever the HTML is exported, it accurately matches the render, versus possibly the render being updated asynchronously after the text it retrieved.

kitsonk:
**Type:** feature

**Description:**

This PR provides two additional APIs to the Projector to make it easy to perform server side rendering.

- `.sandbox()` - This will create a `DocumentFragment` that is used as the root for the projection.  This ensures any virtual DOM rendering does not impact the rendering of the global `document`.  It can take a single argument of type `Document` which can be used with something like `jsdom` where there is not a global `document` exposed.  It also switches the `Projector` into synchronous render mode, making it easier to ensure that when the projector is introspected, it is always in a fully rendered state.
- `.toHtml()` - Returns a `string` that provides the full HTML of the projection.  This method will throw if the projector is not attached.

Resolves #493

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -166,7 +166,6 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		private _root: Element;
 		private _async = true;
 		private _attachHandle: Handle;
-		private _attachType: AttachType;
 		private _projectionOptions: ProjectionOptions;
 		private _projection: Projection | undefined;
 		private _scheduled: number | undefined;
@@ -343,7 +342,6 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 					this._attachHandle = { destroy() { } };
 				}
 			});
-			this._attachType = type;
 
 			switch (type) {
 				case AttachType.Append:
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -283,10 +283,10 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		}
 
 		public toHtml(): string {
-			if (this.projectorState !== ProjectorAttachState.Attached) {
+			if (this.projectorState !== ProjectorAttachState.Attached || !this._projection) {
 				throw new Error('Projector is not attached, cannot return an HTML string of projection.');
 			}
-			return this._projection!.domNode.outerHTML;
+			return this._projection.domNode.outerHTML;
 		}
 
 		public __render__() {
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -1,4 +1,5 @@
 import global from '@dojo/core/global';
+import { createHandle } from '@dojo/core/lang';
 import { Handle } from '@dojo/interfaces/core';
 import { VNode } from '@dojo/interfaces/vdom';
 import { dom, Projection, ProjectionOptions, VNodeProperties } from 'maquette';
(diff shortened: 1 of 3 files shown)


kitsonk:
Related to #487

Todo:

- [ ] Update README.md (also with information from #487)

src/mixins/Projector.ts
```diff
@@ -128,12 +143,14 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		public projectorState: ProjectorAttachState;
 
 		private _root: Element;
+		private _async = true;
 		private _attachHandle: Handle;
+		private _attachType: AttachType;
```
kitsonk: Ooops I need to remove this... I thought I needed it for `.toHtml()` but didn't.

tests/unit/mixins/Projector.ts
```diff
@@ -110,6 +110,51 @@ registerSuite({
...
+						count++;
+						return v('div', [ String(count) ]);
+					}
+				}();
+
+				projector.sandbox();
+				assert.strictEqual(count, 1, 'render should have been called once');
+				assert.strictEqual(projector.root.firstChild!.textContent, '1', 'should have rendered "1"');
+				projector.invalidate();
+				assert.strictEqual(count, 2, 'render should have been called syncronously');
```
dylans: syncronously -> synchronously

src/mixins/Projector.ts
```diff
@@ -222,6 +261,13 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 			super.__setProperties__(properties);
 		}
 
+		public toHtml(): string {
+			if (this.projectorState !== ProjectorAttachState.Attached) {
+				throw new Error('Projector is not attached, cannot return an HTML string of projection.');
+			}
+			return this._projection!.domNode.outerHTML;
```
agubler: We try not to use `!` to skip possible null checks in `src` code, could we check that there is a projection in the catch above?

```ts
if (!this._projection || this.projectorState !== ProjectorAttachState.Attached) {
```

sebilasse:
@kitsonk @dylans 
How would

> .toHtml() - Returns a string that provides the full HTML of the projection. This method will throw if the projector is not attached.

deal with the scoped classes from css-modules ?

kitsonk:
@sebilasse 

> deal with the scoped classes from css-modules ?

It doesn't directly deal with them, nor does it concern itself with them.  It would be up to the implementor to ensure that the CSS modules used to generate the render have been transformed in the same way as what will be loaded on the client.  It is only one aspect of a solution.

sebilasse:
@kitsonk 
I see. In reality this would probably mean a complete rewrite of the postcss-import module because of the bug described in https://github.com/postcss/postcss-import/issues/180

PR #500, opened 2017-05-11
Bind to the original widget instance for injectors.

### LINKED ISSUE #496: Automatically use original widget scope as `bind` when using an Injector
**Bug** 

The injector pattern works by effectively "replacing" the original widget in the widget tree whilst maintaining a handle to the original `render` function. This means that the automatic decorating of `bind` in `WidgetBase` to `HNode`s and `WNode`s actually applies the scope of the Injector widget not the original widget itself.

That means the `bind` for the functions passed to `v` and `w` is not to the original widget as expected but the injector and therefore expected `properties` and other attributes are not available.

When using an injector it should be mandatory to pass the original scope in the `bind` property and this can be used to decorate all of the `DNode`s in an `afterRender`.

agubler:
**Type:** bug / feature

**Description:**

Change `bind` to be a mandatory property for the `Injector` and add an `afterRender` to decorate all HNode/WNode with original `bind`.

Resolves #496

### DIFF
--- a/src/Injector.ts
+++ b/src/Injector.ts
@@ -57,11 +57,8 @@ export function Injector<C, T extends Constructor<BaseInjector<C>>>(Base: T, con
 		protected decoratateBind(node: DNode) {
 			const { bind } = this.properties;
 			decorate(node, (node: HNode | WNode<DefaultWidgetBaseInterface>) => {
-				const { properties } = node;
-
-				if (!properties.bind) {
-					assign(properties, { bind });
-				}
+				const { properties = {} }: { properties: { bind?: any } } = node;
+				properties.bind = bind;
 			}, (node: DNode) => { return isHNode(node) || isWNode(node); });
 
 			return node;
--- a/src/Injector.ts
+++ b/src/Injector.ts
@@ -1,6 +1,14 @@
 import { assign } from '@dojo/core/lang';
-import { WidgetBase } from './WidgetBase';
-import { Constructor, DNode, WidgetProperties } from './interfaces';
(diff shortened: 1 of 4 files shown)


tests/unit/Injector.ts
```diff
@@ -19,13 +25,30 @@ registerSuite({
...
+			}
+		}
+
+		function testFunction(this: any) {
+			assert.strictEqual(this, bind);
+		}
+
 		const testChildren: DNode[] = [ 'child' ];
-		const render = (): DNode => { return 'Called Render'; };
+		const render = (): DNode => { return v('div', {}, [
```
agubler: This test will need to change once bind is not supported in as a property to `v`

src/Injector.ts
```diff
@@ -44,6 +53,17 @@ export function Injector<C, T extends Constructor<BaseInjector<C>>>(Base: T, con
 			super(context);
 		}
 
+		@afterRender()
+		protected decoratateBind(node: DNode) {
+			const { bind } = this.properties;
+			decorate(node, (node: HNode | WNode<DefaultWidgetBaseInterface>) => {
+				const { properties = {} }: { properties: { bind?: any } } = node;
```
kitsonk: Me and my annoying branch coverage... It is likely we aren't testing where bind is actually present, though if we are relying on typings it shouldn't be there.  We might have to instanbul ignore something here.

tests/unit/Injector.ts
agubler: done

PR #501, opened 2017-05-11
Ensure properties and children are 'reset' each render for a projector

### LINKED ISSUE #498: Properties and children become stateful in projectors
**Bug**

For all widgets in the widget tree `properties` and `children` are "reset" to the the object provided to `w` with widget `render`, this is not true for widgets that have been converted to a projector using the `ProjectorMixin`. 

Both the `properties` and `children` become stateful as the original user defined properties are never re-passed via `setProperties` or `setChildren` and as a result this can cause unexpected/inconsistent behaviour.

Any projector should reset it's properties and children to a copy of the last values passed to either `setProperties` or `setChildren` before the `render` - The `beforeRender` lifecycle hook should provide a mechanism to do achieve this.

agubler:
**Type:** bug

**Description:**

Ensure properties and children are reset each render back to the last received set by either `setProperties` or `setChildren`

Resolves #498

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -1,3 +1,4 @@
+import { assign } from '@dojo/core/lang';
 import global from '@dojo/core/global';
 import { Handle } from '@dojo/interfaces/core';
 import { dom, Projection, ProjectionOptions, VNodeProperties } from 'maquette';
@@ -123,7 +124,7 @@ const eventHandlers = [
 ];
 
 export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T): T & Constructor<ProjectorMixin<P>> {
-	return class extends base {
+	class Projector extends base {
 
 		public projectorState: ProjectorAttachState;
 
@@ -135,6 +136,8 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		private _paused: boolean;
 		private _boundDoRender: FrameRequestCallback;
 		private _boundRender: Function;
+		private _projectorChildren: DNode[];
+		private _projectorProperties: P;
 
 		constructor(...args: any[]) {
 			super(...args);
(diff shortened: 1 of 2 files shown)


src/mixins/Projector.ts
```diff
@@ -215,14 +218,22 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		}
 
 		public setChildren(children: DNode[]): void {
+			this._projectorChildren = [ ...children ];
```
kitsonk: Why wouldn't we be doing `super.__setChildren__(this._projectorChildren)`?  Feels like a little bit of indirection.  Also if we did this, wouldn't we just be able to `this.setChildren(this.children)` and `this.setProperties(this.properties)` below?

src/mixins/Projector.ts
```diff
@@ -215,14 +218,22 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
...
 			super.__setChildren__(children);
 		}
 
 		public setProperties(properties: P & { [index: string]: any }): void {
+			this._projectorProperties = assign({}, properties);
```
kitsonk: Could also do `{ ...properties }` now, though I guess with offloading to native, `assign` will always be quicker.

src/mixins/Projector.ts
agubler: We are taking a copy at the point in time the user has explicitly called `setProperties` or `setChildren`, because the properties and children that are passed can be decorated in lifecycle phases such as `beforeRender` - on a re-render for all other widgets in the widget tree a "fresh" properties object is used via the usage of `w()` but for the projector that is not the case and therefore the next render the possibly modified `properties` are used again (which means it can cause unexpected behaviour). So with these "copied" local value we re-set the `properties`/`children` in `WidgetBase` by calling the setters.

We could actually directly call the supers `__setProperties__` and `__setChildren__` from the `__render__` method as we only need the to take the copy when the user has actually called the methods intended for public consumption `setProperties` and `setChildren` - I can change that.

src/mixins/Projector.ts
agubler: I actually prefer the spreading but you get an error `Spread types may only be created from object types.` I think when the type is a generic.

src/mixins/Projector.ts
agubler: Ha nope of course I'm wrong, I do need to take the copy every time over wise my "frozen" properties/children will possibly be mutated then next time they are used to call `__setProperties__` or `__setChildren__`. Doh.

src/mixins/Projector.ts
kitsonk: Hmmm... I couldn't find anything on TypeScript issues, so I opened this: https://github.com/Microsoft/TypeScript/issues/15792

kitsonk: It was a dupe... PR Microsoft/TypeScript#13288 will resolve it.

PR #503, opened 2017-05-11
Better Error for using widgets multiple times

sebilasse:
As a widget author I'd like to know at least which widget caused the error
' It is recommended to provide a unique `key` property when using the same widget multiple times '

**Type:** bug / feature

**Description:**

Resolves #???

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -577,6 +577,9 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			}
 			if (!key && cachedChildren.length > 1) {
 				const errorMsg = 'It is recommended to provide a unique `key` property when using the same widget multiple times';
+				if (cachedChildren[0].child) {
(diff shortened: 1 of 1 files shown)


agubler:
@sebilasse Thank you for the contribution! Could I ask you to raise an issue for this enhancement so that we can fit it into our established project workflow? In case you want more details about our workflow then please see [CONTRIBUTING.md](https://github.com/dojo/meta/blob/master/CONTRIBUTING.md).

There are a couple of issues with the changes proposed:

1) `.name` is not an known property of `Function` without using the `es2015.core` TypeScript lib so causes a compile time error.
2) A side effect of the change causes multiple unit test failures (due to a failing test for the `warn` and also poorly stubbing console in the WidgetBase tests).

I've raised a pull request that should resolve your issue if you want to take a look at #504. Once you have raised an issue I'll reference it in that pull request and if you're happy with the change proposed, please go ahead and close this one.

PR #504, opened 2017-05-12
Include widget name in multiple widget warn message.

### LINKED ISSUE #505: Better Error for using widgets multiple times
In `widget-core/WidgetBase.ts`
As a widget author I'd like to know at least which widget caused the error
' It is recommended to provide a unique `key` property when using the same widget multiple times '

Would be solved by #504 ...
Thankyou @agubler

agubler:
**Type:** feature

**Description:**

Adds the `constructor.name` of the widget that has been used multiple times without a unique key warning.

Resolves #505


agubler:
Failing test on IE11 that needs to be investigated.

agubler:
IE11 doesn't support `Function.name` and returns undefined.

PR #510, opened 2017-05-13
Check children before setting widget as dirty

### LINKED ISSUE #509: The widget tree is being rendered every time any widget is invalidated.
**bug**

Currently a widget will be considered `dirty: true` every render regardless of whether the `children` or `properties` have actually changed. 

This is because although the `properties` are properly diff'd according to the specified `DiffType` or custom diff function, setting the `dirty` flag accordingly. However `children` are set on a widget **every** render and this automatically marks the widget as dirty and therefore will be run in the next render (plus all it's children).

If a widget has no children (i.e. an empty array) and the previous children were also empty then it should not be marked as dirty.

This was introduced by [children becoming mandatory on WNode and the children being defaulted to `[]` in `w`](https://github.com/dojo/widget-core/pull/444).

**Code**

```ts
let childRenderCalled = 0;
class ChildWidget extends WidgetBase {
	render() {
		childRenderCalled++;
		return v('div');
	}
}

class TestWidget extends WidgetBase {
	render() {
		return w(FooWidget, { key: '1' }, []);
	}
}

const widget: any = new TestWidget();
widget.__render__();
widget.invalidate();
widget.__render__();
```

**Expected behavior:**

expect `childRenderCalled === 1`

**Actual behavior:**

expect `childRenderCalled === 2`

agubler:
**Type:** bug

**Description:**

Check that either previous children or current children exist before marking a widget as `dirty`.

Resolves #509

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -365,12 +365,14 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 	}
 
 	public __setChildren__(children: (C | null)[]): void {
-		this._dirty = true;
-		this._children = children;
-		this.emit({
-			type: 'widget:children',
-			target: this
-		});
+		if (this._children.length || children.length) {
+			this._dirty = true;
+			this._children = children;
+			this.emit({
+				type: 'widget:children',
+				target: this
+			});
+		}
 	}
 
 	public __render__(): VNode | string | null {
@@ -581,9 +583,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 				this.emit({ type: 'error', target: this, error: new Error(errorMsg) });
 			}
 
-			if (Array.isArray(children)) {
-				child.__setChildren__(children);
-			}
+			child.__setChildren__(children);
 			return child.__render__();
 		}
 
--- a/tests/unit/WidgetBase.ts
+++ b/tests/unit/WidgetBase.ts
@@ -1419,6 +1419,28 @@ widget.__setProperties__({
 		widget.__render__();
 		assert.equal(foo, 2);
 	},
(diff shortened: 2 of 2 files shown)


kitsonk:
Is the issue wider than this?  Will children of any length, even if there are no differences effect this, or will only `__setChildren()__` be called when there is a _cascade_ of invalidation?  For example if I did this in a render, what would occur?

```ts
class ULWidget extends WidgetBase {
  render() {
    return v('ul', [ ...this.children ]);
  }
}

class StaticULWidget extends WidgetBase {
  render () {
    return v('div', [ w(ULWidget, [ v('li', [ 'Item One' ]) ] ]);
  }
}
```

I will always call `ULWidget` from `StaticULWidget` with exactly the same children.

agubler:
@kitsonk yes that would always "dirty" the `ULWidget` and was the desired behaviour when the changes to mark widgets as dirty in `__setChildren__` were implemented. 

It gets very complicated if we were to try and "diff" the children as we would essentially have to replicate the property diff'ing down the children tree which would then be done again as part of the properties lifecycle and to make it even more complicated we don't have access to any "custom" diff'ing, which means we cannot accurately diff at this point in the lifecycle anyway.

As a result a decision was made to simply mark all widgets that had children as dirty to ensure the children were re-rendered when their properties had changed. Unfortunately due to the changes that added a default empty array for children it meant that `__setChildren__` was being called every time regardless of whether there were ever children and there was no negative unit test to highlight the change of behaviour.

kitsonk:
> As a result a decision was made to simply mark all widgets that had children as dirty to ensure the children were re-rendered when their properties had changed.

Understood... I guess too that _static children_ is an edge case versus _no children_ being quite a bit more common.  I guess something to _keep an eye on_ IRL to see if there is a performance optimisation, though as you imply, the cure might be worse than the cause.

PR #516, opened 2017-05-17
Add higher order Container component to support registry items

### LINKED ISSUE #512: Support `registry` items using a Container
**Enhancement** 

It should be possible to support registry items using a container, currently this is not possible with the `Container` as a mixin (as it needs to extend a component constructor).

If the `Container` was converted to be a "higher order component" we could support both component constructor and registry items.

agubler:
**Type:** feature

**Description:**

Add support for registry items with the `Container` by converting the from a mixin to a higher order component.

Resolves #512


src/Container.ts
```diff
@@ -0,0 +1,28 @@
+import { WidgetBase } from './WidgetBase';
+import { Constructor, DNode, WidgetBaseInterface, RegistryLabel } from './interfaces';
+import { w } from './d';
+import { defaultMappers, BaseInjector } from './Injector';
+
+export function Container<W extends WidgetBaseInterface>(
+	component: Constructor<W> | RegistryLabel,
+	name: string,
```
matt-gadd: This should/could probably be a `RegistryLabel`

src/Container.ts
```diff
@@ -0,0 +1,28 @@
...
+	{ getProperties = defaultMappers.getProperties, getChildren = defaultMappers.getChildren }: any = defaultMappers
```
matt-gadd: the formatting of this is difficult to read, can you multi line it

matt-gadd: can the `any` be typed?

src/Container.ts
```diff
@@ -0,0 +1,28 @@
...
+): Constructor<WidgetBase<W['properties']>> {
+
+	return class extends WidgetBase<any> {
+		protected render(): DNode {
+			const { properties, children } = this;
+
+			return w<BaseInjector<any>>(name, {
+				bind: this,
+				render: () => { return w(component, properties, children); },
```
matt-gadd: if this explicitly returns you can drop the braces and `return` statement

src/Container.ts
agubler: It can. done.

src/Container.ts
agubler: It can, tend not to by default. Updated all the same.

PR #521, opened 2017-05-19
Adding support for widget meta, issue 507

### LINKED ISSUE #507: Meta mechanism for providing additional information inside a Widget
**Enhancement** 
As part of a bigger plan to reduce/remove dom access when authoring widgets, we need a way to flexibly allow users to access certain information about dom nodes and other things associated with them. A typical example in the http://github.com/dojo/widgets repo would be the need to access dimensions such as `offsetTop`, `scrollTop` etc from an element. 

Ideally this meta functionality in `WidgetBase` would be flexible to have implementations injected into it, and hooks so that we do not need to build every meta type into `WidgetBase` nor have to extend `WidgetBase` with mixins for each type. The return types should correctly type also. 

With the above in place the meta implementations would be the only escape hatches to access dom nodes via `WidgetBase` (other than on event objects), and we will then hopefully be able to deprecate the lifecycle hooks `afterCreateCallback`, `afterUpdateCallback` as well as removing `afterCreate` and `afterUpdate` from `VirtualDomProperties`.

Future additional meta implementations could include:
* State of animation players when using the Web Animation API
* Visibility/Intersection of elements using the Intersection Observer API

rorticus:
**Type:** feature

**Description:**

Infrastructure for providing meta information for a widget. One example of this (and included in the PR) is for getting DOM node dimensions:

```typescript
class MyWidget extends BaseWidget<WidgetProperties> {
    render() {
        const dimensions = this.meta(Dimensions);

        console.log(`I am this many tall! ${dimensions.height}`);
        
        return v('div');
    }
}
```

This will return a "zero state" on the initial render (as there is no dom node), but it will also invalidate the widget immediately and schedule another render, so the first thing the user _sees_ will be the render with the correct meta data.

New Meta options can be created simply:

```typescript
class AttributeMeta implements WidgetMeta {
    nodes: any;

    constructor(props: WidgetMetaProperties) {
        this.nodes = props.nodes;
    }

    getAttribute(key, attribute) {
        return this.nodes.get(key).getAttribute(attribute);
    }
}

/// meanwhile, in render,
render() {
    const attribute = this.meta(AttributeMeta).getAttribute("root", "someAttribute");

    // do something with attribute
}
```

Resolves #507


rorticus:
Added `Intersection` meta for more meta-y goodness.

```typescript
class MyWidget extends WidgetBase<any> {
    render() {
        const isVisibleInViewport = this.meta(Intersection).get('root') > 0;

        return v('div', { key: 'root' });
    }
}
```

src/WidgetBase.ts
```diff
@@ -421,8 +464,8 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 		return this._cachedVNode;
 	}
 
-	public invalidate(): void {
-		if (this._renderState === WidgetRenderState.IDLE) {
+	public invalidate(force?: boolean): void {
```
kitsonk: Can you explain why this is being added as part of this PR?  It doesn't feel germane to the intent of the PR, but maybe I am missing something.

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
...
+				this._invalidate();
+			});
+		}, {
+			rootMargin: '0px',
+			threshold: 0
+		});
+	}
+
+	has(key: string) {
+		const node = this._nodes.get(key);
```
kitsonk: Untested function.

tests/support/loadJsdom.ts
```diff
@@ -31,6 +31,13 @@ global.requestAnimationFrame = (cb: (...args: any[]) => {}) => {
 
 global.cancelAnimationFrame = () => {};
 
+global.IntersectionObserver = function () {
```
kitsonk: We should transition to test-extras loadJsdom.  I am not sure why we haven't moved in widget-core.  Why do we have to do this as well?

package.json
```diff
@@ -46,6 +46,7 @@
     "grunt-dojo2": "beta1",
     "grunt-tslint": "^5.0.1",
     "intern": "^3.4.1",
+    "intersection-observer": "^0.2.1",
```
kitsonk: This is only a dev dependency?  It feels like we are using a polyfill here.  This feels like we are introducing something fundamental as a dependency, like pointer events.  We should have a more formal discussion about introducing something like this.  While the [platform status seems](https://www.chromestatus.com/feature/5695342691483648) like it is headed the right direction, we still need to proactively make a decision.

tests/support/loadJsdom.ts
agubler: Using `test-extras` for unit testing in widget-core has issues, but we could probably load the Jsdom module. It will mean that we need to have a dev dependency on a released version of widget-core (due to peer deps requirements of test-extras).

matt-gadd: Have we thought about bringing the jsdom part to widget-core? Would be nice to not have the circular.

src/WidgetBase.ts
```diff
@@ -209,6 +222,9 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 
 	private _renderState: WidgetRenderState = WidgetRenderState.IDLE;
 
+	private _metaMap = new WeakMap<any, any>();
```
matt-gadd: i presume these can be typed?

src/WidgetBase.ts
```diff
@@ -239,6 +255,24 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+		if (!cached) {
+			cached = new MetaType({
+				nodes: this._nodeMap,
+				invalidate: this.invalidate.bind(this)
+			});
+
+			this._metaMap.set(MetaType, cached);
+		}
+
+		if (cached.requiresRender && !this._cachedVNode) {
```
matt-gadd: does `requiresRender` need to be more granular than this? what happens if I add a node in the `render` func later on?

tests/support/loadJsdom.ts
```diff
@@ -31,6 +31,13 @@ global.requestAnimationFrame = (cb: (...args: any[]) => {}) => {
...
+	return {
```
matt-gadd: does the polyfill not add this in node?

tests/unit/meta/Dimensions.ts
```diff
@@ -0,0 +1,119 @@
...
+
+	beforeEach() {
+		rAF = stub(global, 'requestAnimationFrame');
+	},
+
+	afterEach() {
+		rAF.restore();
+	},
+
+	'dimensions are blank on first render, set on second'(this: any) {
```
matt-gadd: do we think this is an implementation detail in name?

tests/unit/meta/Intersection.ts
```diff
@@ -0,0 +1,117 @@
...
+		const div = document.createElement('div');
+
+		document.body.appendChild(div);
+
+		const widget = new TestWidget();
+		widget.append(div);
+
+		resolveRAF();
+
+		const node = (<any> widget)._nodeMap.get('root');
```
matt-gadd: maybe we could just get this node out of the dom via a selector/id, then we wouldn't have to access internals?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
+import { WidgetMeta, WidgetMetaProperties } from '../WidgetBase';
+import Map from '@dojo/shim/Map';
+import WeakMap from '@dojo/shim/WeakMap';
+import global from '@dojo/core/global';
+
+import 'intersection-observer';
+
+export default class Intersection implements WidgetMeta {
+	private _nodes: Map<string, any>;
+	private _listeners: WeakMap<Element, any>;
```
matt-gadd: can this be typed?

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
+import { WidgetMeta, WidgetMetaProperties } from '../WidgetBase';
+import Map from '@dojo/shim/Map';
+
+export interface DimensionResults {
```
matt-gadd: missing the `offset` group

src/WidgetBase.ts
```diff
@@ -263,7 +297,9 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			if (!properties.bind) {
 				properties.bind = this;
 			}
-		}, (node: DNode) => { return isHNode(node) || isWNode(node); });
+		}, (node: DNode) => {
```
agubler: Thanks for the formatting... 😄

src/WidgetBase.ts
```diff
@@ -280,6 +317,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 	 */
 	private afterUpdateCallback(element: Element, projectionOptions: ProjectionOptions, vnodeSelector: string,
 		properties: VNodeProperties, children: VNode[]): void {
+		this._setNode(element, projectionOptions, vnodeSelector, properties, children);
```
agubler: Why do you need to pass more than the `element` and `properties` as that looks like it's all the `this._setNode` function needs.

src/WidgetBase.ts
agubler: I will let @rorticus correct me (or maybe @matt-gadd) but I think is to support the mechanism which forces an immediate re-render when the meta is can only be collected once the Elements are created (and therefore missed by the first render). With the `renderState` introduced, normal invalidation triggered in the `render` lifecycle is a noop and therefore `force` allows that to be overridden in special circumstances.

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
...
+	scrollTop: number;
+	scrollWidth: number;
+	top: number;
+	width: number;
+}
+
+export default class Dimensions implements WidgetMeta {
+	private _nodes: Map<string, any>;
+
+	readonly requiresRender = true;
```
agubler: I do like the explicit `public` visibility keyword (for properties and methods)

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	constructor(properties: WidgetMetaProperties) {
+		this._nodes = properties.nodes;
+	}
+
+	has(key: string): boolean {
+		return !!this._nodes.get(key);
```
agubler: Can you not use `this._nodes.has(...`?

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
...
+				scrollWidth: node.scrollWidth,
+				bottom: clientRect.bottom,
+				height: clientRect.height,
+				left: clientRect.left,
+				right: clientRect.right,
+				top: clientRect.top,
+				width: clientRect.width
+			};
+		}
+		else {
```
agubler: don't need the `else` branch as you are returning from the if branch

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
...
+
+	constructor(properties: WidgetMetaProperties) {
+		this._nodes = properties.nodes;
+	}
+
+	has(key: string): boolean {
+		return !!this._nodes.get(key);
+	}
+
+	get(key: string): DimensionResults {
```
agubler: I think `DimensionResults` should probably be either `Readonly<DimensionResults>` or the interface itself have  readonly properties.

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
...
+		});
+	}
+
+	has(key: string) {
+		const node = this._nodes.get(key);
+
+		return node && this._listeners.has(node);
+	}
+
+	get(key: string) {
```
agubler: What is the return type?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
...
+				this._listeners.set(entry.target, entry);
+				this._invalidate();
+			});
+		}, {
+			rootMargin: '0px',
+			threshold: 0
+		});
+	}
+
+	has(key: string) {
```
agubler: missing `: boolean`

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
...
+import Map from '@dojo/shim/Map';
+import WeakMap from '@dojo/shim/WeakMap';
+import global from '@dojo/core/global';
+
+import 'intersection-observer';
+
+export default class Intersection implements WidgetMeta {
+	private _nodes: Map<string, any>;
+	private _listeners: WeakMap<Element, any>;
+	private _invalidate: Function;
```
agubler: We can type this better `() => void`

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
+import { WidgetMeta, WidgetMetaProperties } from '../WidgetBase';
+import Map from '@dojo/shim/Map';
+import WeakMap from '@dojo/shim/WeakMap';
+import global from '@dojo/core/global';
+
+import 'intersection-observer';
+
+export default class Intersection implements WidgetMeta {
```
agubler: Can we do a named export and then also export default below?

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,64 @@
...
+	right: number ;
+	scrollHeight: number;
+	scrollLeft: number;
+	scrollTop: number;
+	scrollWidth: number;
+	top: number;
+	width: number;
+}
+
+export default class Dimensions implements WidgetMeta {
```
agubler: Can we do a named export and export default below please?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,53 @@
...
+
+	get(key: string) {
+		// is this node set up?
+		const node = this._nodes.get(key);
+
+		if (node) {
+			if (this._listeners.has(node)) {
+				return this._listeners.get(node).intersectionRatio;
+			}
+			else {
```
agubler: else branch is redundant.

tests/support/loadJsdom.ts
agubler: That would be nice, but would need to live in `src` so that it could be used externally...

tests/unit/meta/Dimensions.ts
agubler: Perhaps something like `dimensions are correct configured`? (even if in the test we do two renders, etc?)

tests/support/loadJsdom.ts
kitsonk: Only thought is that we would want `jsdom` as a dev dependency, but then we have a dev only module of `loadJsdom`?  That would essentially make it a hard dependency, right?  Even though it would be built out, it feels a little bit _dirty_, IMO.

src/WidgetBase.ts
rorticus: You are correct @agubler ! Normally the widget won't schedule a re-render if there is already a render happening, but some meta require a DOM node to work correctly, so we need to render twice in order to provide one.

src/meta/Intersection.ts
rorticus: Removed 😛

tests/support/loadJsdom.ts
rorticus: Removed this code from loadJSdom, this discussion can be continued when the intersection observer PR is in!

package.json
rorticus: Removed

src/WidgetBase.ts
rorticus: Yes! Sorry. Should be typed now.

src/WidgetBase.ts
rorticus: OK, removed `requiresRender` completely and we're relying on the passed in `invalidate` property. If a widget requires a node it can call `invalidate` itself.

tests/support/loadJsdom.ts
rorticus: Removed this from the PR...

tests/unit/meta/Dimensions.ts
rorticus: Removed from the PR...

tests/intern.ts
```diff
@@ -64,7 +64,8 @@ export const loaderOptions = {
 		{ name: 'maquette', location: 'node_modules/maquette/dist', main: 'maquette' },
 		{ name: 'pepjs', location: 'node_modules/pepjs/dist', main: 'pep' },
 		{ name: 'grunt-dojo2', location: 'node_modules/grunt-dojo2'},
-		{ name: 'sinon', location: 'node_modules/sinon/pkg', main: 'sinon' }
+		{ name: 'sinon', location: 'node_modules/sinon/pkg', main: 'sinon' },
+		{ name: 'intersection-observer', location: 'node_modules/intersection-observer', main: 'intersection-observer' }
```
matt-gadd: this won't be needed for now :)

rorticus: Oopsie! Removed.

README.md
```diff
@@ -1007,6 +1008,25 @@ The initialization function is run from the context of the HTML element.
 It should be noted that children nodes are removed from the DOM when widget instantiation occurs, and added as children
 to the widget instance.
 
+#### Meta
```
agubler: Maybe we should talk about the mechanism here and how to implement a meta module? then we can add sections for each of the meta (Dimensions & Intersection Observers etc) as they are implemented?

src/WidgetBase.ts
```diff
@@ -209,6 +221,9 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 
 	private _renderState: WidgetRenderState = WidgetRenderState.IDLE;
 
+	private _metaMap = new WeakMap<WidgetMetaConstructor<any>, WidgetMeta>();
+	private _nodeMap = new Map<any, any>();
```
agubler: `Map<string, Element>()`?

src/WidgetBase.ts
agubler: Just making sure you didn't miss this comment ^^^

tests/unit/WidgetBase.ts
```diff
@@ -5,7 +5,9 @@ import * as assert from 'intern/chai!assert';
 import { stub, spy } from 'sinon';
 import { v, w, registry } from '../../src/d';
 import { DNode, WidgetProperties } from '../../src/interfaces';
-import { WidgetBase, diffProperty, DiffType, afterRender, beforeRender, onPropertiesChanged } from '../../src/WidgetBase';
+import {
```
agubler: If we are going to do multi lines can we do?

```ts
import {
    WidgetBase,
    diffProperty
    ...
} from '../../src/WidgetBase';
```

src/WidgetBase.ts
rorticus: Thanks for calling that out. Updated.

README.md
rorticus: Put in a quick example of meta. I have examples for Dimensions and Intersection ready to go into their own PRs. What kind of info do you think you we need here?

src/WidgetBase.ts
rorticus: Got it!

tests/unit/WidgetBase.ts
rorticus: Updated!

README.md
agubler: Perhaps we should talk about how to create a custom meta module? Maybe show a trivial example?

rorticus: Does the blurb I added make sense then?

rorticus:
Updated to include a `requireNode` mechanism for meta to use. When a node is added with `requireNode`, if the node does not exist, the widget will be rerendered. if the node does not exist on the next render, an error will be thrown.

```typescript
get(key: string) {
    this.widgetMetaProperties.requireNode(key);
}
```

tests/unit/meta/meta.ts
```diff
@@ -145,9 +140,9 @@ registerSuite({
 		const widget = new TestWidget();
 		widget.append(div);
 
-		resolveRAF();
-		resolveRAF();
-
-		assert.strictEqual(renders, 1, 'expected one renders');
+		assert.throws(() => {
+			resolveRAF();
```
agubler: Do you need both of these `resolveRAF()` calls?

pottedmeat: I think the first one performs an invalidate which is resolved in the second one.

src/mixins/Projector.ts
```diff
@@ -189,7 +189,9 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T)
 			this._boundDoRender = this.doRender.bind(this);
 			this._boundRender = this.__render__.bind(this);
 
-			this.own(this.on('widget:children', this.invalidate));
+			this.own(this.on('widget:children', () => {
+				this.invalidate(true);
```
agubler: is this needed?

README.md
```diff
@@ -1069,6 +1070,66 @@ The initialization function is run from the context of the HTML element.
...
+
+Widget meta is used to access information that you would normally only have access to on a DOM element. For example, the dimensions of an HTML node.  You can access and respond to meta data during a widget's render operation.
```
pottedmeat: I know that this has a lot of features specific to DOM, but could/would we ever recommend using this for non-DOM meta?

README.md
```diff
@@ -1069,6 +1070,66 @@ The initialization function is run from the context of the HTML element.
...
+import {WidgetMeta} from "@dojo/widget-core/WidgetBase";
+
+class UsefulMeta implements WidgetMeta {
+	private _props: WidgetMetaProperties;
+	
+	constructor(properties: WidgetMetaProperties) {
+		this._props = properties;
+	}
+	
+	nodeAccessingMethod(key: string) {
```
pottedmeat: We can make this a real example. I like the `has` method from the tests.

src/WidgetBase.ts
```diff
@@ -59,6 +59,19 @@ interface DiffPropertyConfig {
 	diffFunction?<P>(previousProperty: P, newProperty: P): PropertyChangeRecord;
 }
 
+export interface WidgetMetaConstructor<T> {
```
pottedmeat: Why are these in `WidgetBase` and not `interfaces`?

src/WidgetBase.ts
```diff
@@ -409,11 +454,22 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+			});
+			this._requiredNodes.clear();
+
 			let dNode = render();
 			const afterRenders = this.getDecorator('afterRender');
 			afterRenders.forEach((afterRenderFunction: Function) => {
 				dNode = afterRenderFunction.call(this, dNode);
 			});
+
+			this._nodeMap.clear();
```
pottedmeat: Won't this be a problem if, on a re-render, the nodes in `_requireNodes` aren't created or updated and are then not passed into `_nodeMap` through `_setNode` which will throw the error above?

src/WidgetBase.ts
pottedmeat: I don't think this is actually working as expected. Since meta relies on having access to DOM, it means it’s hitting the next `requestAnimationFrame` where the projector actually creates the DOM, which means this forced invalidate isn’t happening synchronously. If we have to render in a new animation frame anyway, we should just invalidate using that API.

pottedmeat: In this current PR, this is what's already happening - the forced invalidate merely sets the `_dirty` flag, the actual render won't happen until the next animation frame.

rorticus:
@agubler @matt-gadd You mind giving another quick review? Neil made some great updates that mean we don't need to modify the `invalidate` command to support forcing invalidation.

Thanks!

src/WidgetBase.ts
```diff
@@ -241,6 +259,33 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 		this._checkOnElementUsage();
 	}
 
+	protected meta<T extends WidgetMeta>(MetaType: WidgetMetaConstructor<T>): T {
+		let cached = this._metaMap.get(MetaType);
+		if (!cached) {
+			const boundInvalidate = this.invalidate.bind(this);
+			const invalidate = () => {
```
pottedmeat: I know I made this an arrow function, but it could (should?) just be a plain old function

src/WidgetBase.ts
```diff
@@ -59,6 +60,19 @@ interface DiffPropertyConfig {
 	diffFunction?<P>(previousProperty: P, newProperty: P): PropertyChangeRecord;
 }
 
+export interface WidgetMetaConstructor<T> {
+	new (properties: WidgetMetaProperties): T;
+}
+
+export interface WidgetMetaProperties {
+	nodes: Map<string, any>;
```
pottedmeat: Could `any` be `HTMLElement` instead? Or something similar?

agubler: At least `Element` indeed

src/WidgetBase.ts
rorticus: Yeah I guess it isn't using anything scoped so we should probably remove the arrow function.

agubler:
@rorticus What's the status on this? I believe @matt-gadd mentioned there were some additional changes that you were working on?

rorticus:
@agubler I'll talk to @pottedmeat and see where we are at. @matt-gadd mentioned we could now create a `MetaBase` that includes the constructor and `has` methods, which seems like a pretty good idea.

pottedmeat:
Yeah, I was in the middle of testing them. I can push them out this morning perhaps.

agubler:
@pottedmeat testing the implementations? We should be able to merge the mechanism without having raised the implementations, I am wanting to do a release soon and would like at least this PR in if possible.

rorticus:
Updated to the latest from @pottedmeat .

src/RegistryHandler.ts
```diff
@@ -1,5 +1,5 @@
 import { Evented } from '@dojo/core/Evented';
-import { WidgetBaseConstructor, RegistryLabel } from './interfaces';
+import { Constructor, RegistryLabel, WidgetBaseInterface } from './interfaces';
```
agubler: Looks like this has included changes from commits that are already in master?

src/WidgetBase.ts
```diff
@@ -265,15 +305,33 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+		}, (node: DNode) => {
+			return isHNode(node) || isWNode(node);
+		});
 		return node;
 	}
 
+	protected applyDeferredProperties(element: any, properties: VNodeProperties) {
+		const key = String(properties.key);
+		if (this._deferredProperties.has(key)) {
+			(this._deferredProperties.get(key) || []).forEach(({ object, propertyName, callback }: { object: string, propertyName: string, callback: Function }) => {
```
agubler: This is quite long, can we split it up.

src/WidgetBase.ts
```diff
@@ -265,15 +305,33 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+		}
+	}
+
 	/**
 	 * vnode afterCreate callback that calls the onElementCreated lifecycle method.
 	 */
 	private afterCreateCallback(element: Element, projectionOptions: ProjectionOptions, vnodeSelector: string,
 		properties: VNodeProperties, children: VNode[]): void {
+		this._setNode(element, properties);
+		this.applyDeferredProperties(<HTMLElement> element, properties);
```
agubler: Why do we need to cast this?

src/WidgetBase.ts
```diff
@@ -265,15 +305,33 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
 				properties.bind = this;
 			}
-		}, (node: DNode) => { return isHNode(node) || isWNode(node); });
+		}, (node: DNode) => {
+			return isHNode(node) || isWNode(node);
+		});
 		return node;
 	}
 
+	protected applyDeferredProperties(element: any, properties: VNodeProperties) {
```
agubler: Is this part of the meta changes?

src/WidgetBase.ts
```diff
@@ -626,9 +701,45 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			return this.dNodeToVNode(child);
 		});
 
+		this._prepareDeferredProperties(dNode);
+
 		return dNode.render();
 	}
 
+	private _prepareDeferredProperties(node: DNode) {
```
agubler: Again feels like this is a change that is actually unrelated to the meta config changes.

src/WidgetRegistry.ts
```diff
@@ -45,7 +45,7 @@ export interface WidgetRegistry {
 	 * @param widgetLabel The label of the widget to return
 	 * @returns The WidgetRegistryItem for the widgetLabel, `null` if no entry exists
 	 */
-	get(widgetLabel: RegistryLabel): WidgetBaseConstructor | null;
+	get<T extends WidgetBaseInterface = WidgetBaseInterface>(widgetLabel: RegistryLabel): Constructor<T> | null;
```
agubler: This changes are already in master...

src/interfaces.d.ts
```diff
@@ -374,3 +375,16 @@ export interface WidgetBaseInterface<
 	 */
 	invalidate(): void;
 }
+
+export interface WidgetMeta {
```
agubler: should we have some doc on these public interfaces?

src/RegistryHandler.ts
pottedmeat: Yeah. Fixing now.

src/WidgetBase.ts
```diff
@@ -241,6 +248,33 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+			const boundInvalidate = this.invalidate.bind(this);
+			const invalidate = function () {
+				global.cancelAnimationFrame(scheduled);
+				scheduled = global.requestAnimationFrame(boundInvalidate);
+			};
+			let scheduled: number | undefined;
+			cached = new MetaType({
+				nodes: this._nodeMap,
+				invalidate,
+				requireNode: (key: string) => {
```
matt-gadd: can this now be in base? and we pass the `_requiredNodes` as a prop like `nodes` and `invalidate`

pottedmeat: I think this code is here more for the error reporting at the end of render.

matt-gadd: this builds up the required nodes, and can actually be an implementation detail i think?

tests/unit/meta/meta.ts
```diff
@@ -0,0 +1,201 @@
...
+
+		const widget = new TestWidget();
+		widget.append(div);
+
+		assert.throws(() => {
+			resolveRAF();
+		});
+	},
+
+	'base meta has no-op functions'() {
```
matt-gadd: this test has no assertions and isn't very well typed?

tests/unit/meta/meta.ts
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		const div = document.createElement('div');
+
+		const widget = new TestWidget();
+		widget.append(div);
+
+		resolveRAF();
+
+		assert.strictEqual(renders, 3, 'expected two renders');
```
matt-gadd: expects 2 renders, but asserts 3?

tests/unit/meta/meta.ts
pottedmeat: I put it there just for test coverage. I'm not able to make the class abstract because it would require anything extending it to implement those no-op methods. There's probably an Istanbul instruction to ignore methods, though.

tests/unit/meta/meta.ts
```diff
@@ -0,0 +1,201 @@
...
+			}
+		}
+
+		const div = document.createElement('div');
+
+		const widget = new TestWidget();
+		widget.append(div);
+		const meta = widget.getMeta();
+
+		assert.isTrue((<any> meta).nodes.size > 0);
```
matt-gadd: can we get this so it types? really trying to cut down on the `any`'s in tests.

src/meta/Base.ts
```diff
@@ -0,0 +1,23 @@
...
+		this.invalidate = properties.invalidate;
+		this.requireNode = properties.requireNode;
+	}
+
+	public has(key: string): boolean {
+		this.requireNode(key);
+		return this.nodes.has(key);
+	}
+
+	protected invalidate(): void {}
```
matt-gadd: should this be `private`?

src/meta/Base.ts
```diff
@@ -0,0 +1,23 @@
...
+
+	protected requireNode(key: string): void {}
```
matt-gadd: same with this, should this be `private`?

tests/unit/meta/meta.ts
pottedmeat: It's a protected instance variable. Should I just cast it to `{ nodes: Map<string, HTMLElement> }`?

src/meta/Base.ts
pottedmeat: Some subclasses like `Intersection` will call invalidate when they receive events.

src/meta/Base.ts
pottedmeat: Subclasses will call this directly to notify the widget system they require that node.

src/meta/Base.ts
matt-gadd: yup makes sense, i was being silly here.

tests/unit/meta/meta.ts
matt-gadd: if its `protected` then we're testing an implementation detail. if there is no other way you can write a friendly public function on `TestMeta` that exposes it, that way you won't have to `any` it.

src/meta/Base.ts
```diff
@@ -0,0 +1,23 @@
+import Map from '@dojo/shim/Map';
+import { WidgetMeta, WidgetMetaProperties } from '../interfaces';
+
+export class Base implements WidgetMeta {
+	protected nodes: Map<string, HTMLElement>;
+
+	constructor(properties: WidgetMetaProperties) {
+		this.nodes = properties.nodes;
+		this.invalidate = properties.invalidate;
```
matt-gadd: its kind of odd to be overriding class methods on instantiation.

matt-gadd: what happens on extension?

pottedmeat: So the plan was to allow people to write their own meta class without having to subclass anything. But that's not absolutely necessary and I can move everything into the base class and force people to subclass it.

pottedmeat: I could also just store `invalidate` to `this._invalidate` then write an `invalidate` function that calls `this._invalidate()`

src/WidgetBase.ts
```diff
@@ -209,6 +211,10 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 
 	private _renderState: WidgetRenderState = WidgetRenderState.IDLE;
 
+	private _metaMap = new WeakMap<WidgetMetaConstructor<any>, MetaBase>();
```
agubler: Do you think we should add a line break between these like all the other class properties?

src/WidgetBase.ts
```diff
@@ -241,6 +247,20 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
+		if (!cached) {
+			cached = new MetaType({
+				nodes: this._nodeMap,
+				requiredNodes: this._requiredNodes,
+				invalidate: this.invalidate.bind(this)
+			});
+			this._metaMap.set(MetaType, cached);
+		}
+
+		return cached as T;
```
agubler: Why do we need to cast this?

tests/unit/meta/meta.ts
```diff
@@ -0,0 +1,189 @@
...
+				});
+			}
+		}
+
+		const div = document.createElement('div');
+
+		const widget = new TestWidget();
+		widget.append(div);
+
+		assert.throws(() => {
```
matt-gadd: can we assert the error message?

src/meta/Base.ts
```diff
@@ -0,0 +1,39 @@
...
+	protected nodes: Map<string, HTMLElement>;
+
+	constructor(properties: WidgetMetaProperties) {
+		this._invalidate = properties.invalidate;
+		this._requiredNodes = properties.requiredNodes;
+
+		this.nodes = properties.nodes;
+	}
+
+	has(key: string): boolean {
```
agubler: I know it defaults to `public`, but can we be explicit please.

src/meta/Base.ts
```diff
@@ -0,0 +1,39 @@
...
+	private _invalidate: () => void;
+	private _invalidating: number;
+	private _requiredNodes: Set<string>;
+
+	protected nodes: Map<string, HTMLElement>;
+
+	constructor(properties: WidgetMetaProperties) {
+		this._invalidate = properties.invalidate;
+		this._requiredNodes = properties.requiredNodes;
+
```
agubler: any reason for this line break?

tests/unit/meta/meta.ts
```diff
@@ -0,0 +1,189 @@
...
+function resolveRAF() {
+	for (let i = 0; i < rAF.callCount; i++) {
+		rAF.getCall(i).args[0]();
+	}
+	rAF.reset();
+}
+
+registerSuite({
+	name: 'meta base',
+
```
agubler: lot's of line breaks, are they needed?

src/WidgetBase.ts
agubler: Or remove all the line breaks but have one or the other :)

src/WidgetBase.ts
pottedmeat: Because `T` extends `MetaBase`, when I just say `let cached: T = ` it complains that `MetaBase is not assignable to T` because `T` is a subclass of `MetaBase`.

tests/unit/meta/meta.ts
pottedmeat: I'm not sure what you mean

src/WidgetBase.ts
```diff
@@ -409,11 +437,22 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			const render = beforeRenders.reduce((render, beforeRenderFunction) => {
 				return beforeRenderFunction.call(this, render, this._properties, this._children);
 			}, this.render.bind(this));
+
+			this._requiredNodes.forEach((element, key) => {
```
pottedmeat: I think it's not a private function because it's so similar to the code structure of the surrounding beforeRender/afterRenders calls right before and after it. It could certainly be a private function and, if so, should I do the same thing with the code around it?

I moved it before beforeRender and it works fine.

tests/unit/meta/meta.ts
matt-gadd: Well you're not really asserting that this is the correct error currently, just that it throws. Asserting the error message is useful ie you should get back ```Required node test not found```

pottedmeat: Ah, okay. Thanks

src/WidgetBase.ts
agubler: In a separate PR I've already abstracting the beforeRenders and afterRenders into private functions.

pottedmeat: Cool, I'll take a look at it.

agubler: Could the function actually just be a `beforeRender` anyway?

pottedmeat: Yeah, I moved it and it worked fine.

agubler: Sorry I meant a function decorated with `@beforeRender()`?

src/meta/Base.ts
pottedmeat: I was just separating private/public. I'll remove.

tests/unit/meta/meta.ts
pottedmeat: No, it's just the way I was asked to do it in another PR I had worked on. I'll remove.

src/WidgetBase.ts
pottedmeat: Oh, probably. I'll try it out.

PR #525, opened 2017-05-19
Deffering custom element children until after the node is created

### LINKED ISSUE #517: Creating Custom Element throws error
It's currently not possible to create a custom element programmatically using `document.createElement`, you get the following error:
```
DOMException: Failed to execute 'createElement' on 'Document': 
The result must not have children
```
The spec says the following on elements:
```
The element must not gain any attributes or children, 
as this violates the expectations of consumers who use the createElement 
or createElementNS methods.
```

rorticus:
**Type:** bug

**Description:**

Creating a custom element manually (`document.createElement`) causes an error because elements are not supposed to be created with children (the projector). We're now deferring the appending of the projector until after the element is attached to the dom.

Resolves #517

### DIFF
--- a/src/registerCustomElement.ts
+++ b/src/registerCustomElement.ts
@@ -25,10 +25,20 @@ export function registerCustomElement(descriptorFactory: CustomElementDescriptor
 	let widgetInstance: WidgetBase<any>;
 
 	customElements.define(descriptor.tagName, class extends HTMLElement {
+		private _isAppended = false;
+		private _appender: Function | null = null;
+
 		constructor() {
 			super();
 
-			initializeElement(this);
+			this._appender = initializeElement(this);
+		}
+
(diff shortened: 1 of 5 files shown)


src/registerCustomElement.ts
```diff
@@ -25,10 +25,20 @@ export function registerCustomElement(descriptorFactory: CustomElementDescriptor
...
+
 		constructor() {
 			super();
 
-			initializeElement(this);
+			this._appender = initializeElement(this);
+		}
+
+		connectedCallback() {
+			if (!this._isAppended && this._appender) {
```
agubler: you are missing a test branch here though.

rorticus: Yup, just saw that. Working on that now!

PR #528, opened 2017-05-19
Refactoring DOMWrapper to set properties on wrapped nodes, issue 518

### LINKED ISSUE #518: Can not operate on children reactively in Custom Elements
**Bug**
When using a Custom Element, any children are orphaned and inserted inside `DomWrapper`, so they don't get removed when we later render our widget in Maquette. The `DomWrapper` widgets are available as children in the parent widget, but passing properties to them doesn't work; so there is no way to currently reactively handle children inside a render function.

rorticus:
**Type:** feature

**Description:**

Setting `DomWrapper` properties now sets the properties of the wrapped node.

Resolves #518

### DIFF
--- a/src/util/DomWrapper.ts
+++ b/src/util/DomWrapper.ts
@@ -1,6 +1,5 @@
 import { WidgetBase } from './../WidgetBase';
 import { DNode, WidgetProperties } from './../interfaces';
-import { isHNode } from '../d';
 import { assign } from '@dojo/core/lang';
 import { VNode } from '@dojo/interfaces/vdom';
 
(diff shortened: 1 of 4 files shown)


matt-gadd:
@rorticus I think I may have a slightly nicer solution that at least hands off to maquette rather than having to transfer the properties ourselves. And it also types nicely for consumers. POC here, lemme know what you think:
https://gist.github.com/matt-gadd/e3ce8091f95c966a3399e4eab7271978

matt-gadd:
In fact we might be able to leverage https://github.com/AFASSoftware/maquette/pull/99, so we don't have to do the replacement node hack at all.

kitsonk:
possibly, I think though that code path ATM would only work on projector attachment. Worth having a peak though because obviously getting maquette to do the changes would be better.

matt-gadd:
@kitsonk sorry I meant in custom elements themselves (which have there own projector), because then we can consume the children without overwriting them.

kitsonk:
Yes, if they use dom.merge the sub dom should be automagically stitched together.

PR #540, opened 2017-05-25
Switch mixins to not require widgets to extend their properties interface

### LINKED ISSUE #524: Not possible to pull from a passed registry via a Container
**Enhancement**

I expected to be able to source my injector from a scoped/injected registry rather than just a global one. To do this I tried to mixin the RegistryMixin:

**Code**

```
const FooContainer = RegistryMixin(container(Foo, 'foo-state'));
```
```
registry.define('foo-state', Injector(BaseInjector, {}));
```
```
render() {
    return w(FooContainer, { registry });
}
```

**Expected behavior:**

expected to be able to inject the registry.

**Actual behaviour:**

a typings error, because the `registry` property does not exist on the properties of `Foo`.

agubler:
**Type:** feature

**Description:**

Currently it isn't possible add mixin capabilities unless an author has explicitly extended the relevant property interfaces at design time.

```ts
interface MyWidgetProperties extends WidgetProperties { }

class MyWidget extends WidgetBase<MyWidgetProperties> { }
```

As the widget's `MyWidgetProperties` doesn't extend the properties required for any mixin capabilities it means that the mixins cannot be used.

```ts
// errors as `MyWidgetProperties` does not extend `ThemeableProperties`
const MyThemeableWidget = ThemeableMixin(MyWidget); 
```

This change loosens the typings for mixins to accept `any` properties and adds an intersection to the return type to ensure that the extended widget has the correct property shape.

```ts
// An interface that defines the mixins properties
interface MyMixin {
    properties: MyMixinProperties;
}
// mixin return type
function MyMixin<T extends Constructor<WidgetBase<any>>>(Base: T):T & Constructor<MyMixin> {
    return class extends Base { }
}
```
If the `properties` need to be referenced within the mixin then the attribute will need to be added within the class typed to the mixins property interface. 

```ts
function MyMixin<T extends Constructor<WidgetBase<any>>>(Base: T):T & Constructor<MyMixin> {
    return class extends Base { 
        public readonly properties: MyMixinProperties;
        
        public myFunction() {
            const { myMixinProperty } = this.properties;
            // do other things
        }
    }
}
```
Resolves #524

### DIFF
--- a/src/mixins/I18n.ts
+++ b/src/mixins/I18n.ts
@@ -62,10 +62,14 @@ export interface I18nMixin {
 	 * The localized messages, along with a `format` method for formatting ICU-formatted templates.
 	 */
 	localizeBundle<T extends Messages>(bundle: Bundle<T>): LocalizedMessages<T>;
+
+	properties: I18nProperties;
 }
 
-export function I18nMixin<T extends Constructor<WidgetBase<I18nProperties>>>(base: T): T & Constructor<I18nMixin> {
+export function I18nMixin<T extends Constructor<WidgetBase<any>>>(base: T): T & Constructor<I18nMixin> {
 	class I18n extends base {
+
+		public properties: I18nProperties;
 
 		constructor(...args: any[]) {
 			super(...args);
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -159,8 +159,8 @@ function setDomNodes(vnode: VNode, domNode: Element | null = null) {
 	}
 }
 
-export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T): T & Constructor<ProjectorMixin<P>> {
-	class Projector extends base {
+export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T): T & Constructor<ProjectorMixin<P>> {
+	class Projector extends Base {
 
 		public projectorState: ProjectorAttachState;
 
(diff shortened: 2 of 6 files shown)


src/mixins/I18n.ts
```diff
@@ -62,11 +62,15 @@ export interface I18nMixin {
 	 * The localized messages, along with a `format` method for formatting ICU-formatted templates.
 	 */
 	localizeBundle<T extends Messages>(bundle: Bundle<T>): LocalizedMessages<T>;
+
+	properties: I18nProperties;
 }
 
-export function I18nMixin<T extends Constructor<WidgetBase<I18nProperties>>>(base: T): T & Constructor<I18nMixin> {
+export function I18nMixin<T extends Constructor<WidgetBase<any>>>(base: T): T & Constructor<I18nMixin> {
```
kitsonk: Does the default screw things up here?  I just always hate `any`, it feels dirty...

agubler: With it being `I18nProperties` means that the author of a the widget needs to have extended the `I18nProperties` at design time in order for the mixin to be able to be applied.

So I couldn't decorate mixin behaviour on to a widget unless the author originally had the foresight to extend the appropriate property object (or objects). This is especially sucky for things like the `Container` that assume the properties interface of the widget it is wrapping.

I picked `any` because in the future I would like to actually remove the requirement for consumers to even extend the `WidgetProperties` (and just have then automagically available in all scenarios).

Hopefully makes sense!

kitsonk: Yup. 👍

PR #543, opened 2017-05-26
Support null and string render result from a projector

### LINKED ISSUE #532: Projector should support render being returned null or a string
**Enhancement**
Currently the `ProjectorMixin` narrows the functionality of a widget in that you cannot return `null` or a `string` from the widgets `render` call. This can be confusing in that you can use widget `x` in a tree of widgets and it will work just fine say returning a string, but as soon as its used with the `ProjectorMixin` it will throw an error.

It would be nice if the `ProjectorMixin` could handle these scenarios more gracefully.

agubler:
**Type:** feature

**Description:**

Adds support for all attach types to return either `null` or `string` from a top level projector. Maquette doesn't support changing the `vnodeSelector` of the root `VNode`, so need to ensure that they remain consistent when the top level node switches between a `VNode` and `null` or `string`.

##### Merge

For `merge`, if `null` or `string` is returned from `__render__` then a `VNode` is created using the `tagName` from the `root` node, with the `vnode.text` set for `string`s. When a `VNode`s is returned as the top level node, the `vnodeSelector` is set to the `root`s `tagName`.

Assuming a `root` node of `my-app`:

* Returning `null` will render `<my-app></my-app>`
* Returning `string` will render `<my-app>string</my-app>`
* Returning `VNode` will render `<my-app>...</my-app>`

##### Append & Replace

For `append` and `replace` it is not possible to know the correct `tagName` therefore a wrapper `VNode` is created with a `vnodeSelector` of `span`. If the `VNode` is returned for the first `render` then the actual `tagName` is stored and used if subsequent `__render__`s return `string` or `null`.

Assuming a `root` node of `my-app` & first render being `null` or `string`:

* Returning `null` will render `<span></span>`
* Returning `string` will render `<span>string</span>`
* Returning `VNode` (div) will render `<span><div>...</div></span>`

Assuming a `root` node of `my-app` & first render being `VNode` with a `vnodeSelector` of `div`:

* Returning `null` will render `<div></div>`
* Returning `string` will render `<div>string</div>`
* Returning `VNode` will render `<div>...</div>`

Resolves #532

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -175,6 +175,8 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(base: T)
 		private _boundRender: Function;
 		private _projectorChildren: DNode[];
 		private _projectorProperties: P;
+		private _rootTagName: string;
+		private _attachType: AttachType;
 
 		constructor(...args: any[]) {
 			super(...args);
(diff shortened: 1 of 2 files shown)


kitsonk:
Nothing specifically wrong with this PR, but it feels like a lot of code to support what is essentially a _nice_ to have.  Do we have legitimate use cases where we have someone trying to widget that only renders `null` or a test string?  I can appreciate it is trying to be more _correct_ but is the extra weight worth eliminating a run-time error that is likely to actually be something unintended in the first place?

agubler:
@kitsonk The main reason is consistency - you can have a widget that returns `null` or a `string` and it works fine as part of the widget tree but if a user wants to just use that widget and wraps using the `ProjectorMixin` then it will fail (not a good experience). A mixin shouldn't really ever "reduce" functionality.

agubler:
It could be that the widget is lazily loading a child component and whilst that occurs returns `null`, a common pattern. We certainly don't want to throw an error in that instance.

kitsonk:
> It could be that the widget is lazily loading a child component and whilst that occurs returns null, a common pattern. We certainly don't want to throw an error in that instance.

What would it in that case... I was trying to follow the code, but it seems if that is the case the outer projector creates some sort of outer `span` wrapper.  If the widget did load, it would now always be wrapped in a `span`?

agubler:
> What would it in that case... I was trying to follow the code, but it seems if that is the case the outer projector creates some sort of outer span wrapper. If the widget did load, it would now always be wrapped in a span?

Yes for `append` and `replace` that would be the case when if the widget returns `null` or `string` for the first render. This is because we don't know what the tagName will be of the eventually returned node, and Maquette doesn't allow changing the tagName of the root node so we need to continue to wrap the widget in a `span`

If the first `render` returns `VNode` then we store that tagName and use that if any subsequent renders return `null` or `string`

For `merge` then we do, because it is the tag name of the root element we are attaching too so we just use that consistently through out.

agubler:
@kitsonk Agreed, that run time errors are good when a user makes a mistake but we also drive to be consistent for all widgets regardless of whether they are actually the root. 

Another use case could be a custom element that has some state controlled externally and in certain conditions return `null` (or a `string`) to not throw an error, as custom elements have the projector mixed in.

PR #554, opened 2017-05-30
Simplify DomWrapper  

### LINKED ISSUE #553: DomWrapper has issues with certain VirtualDomProperties
**Bug**

Some of the `VirtualDomProperties` cause issues with maquette when they are detected in `DomWrapper` properties (e.g. `classes`).

matt-gadd:
**Type:** feature

**Description:**
Thanks to maquette (https://github.com/AFASSoftware/maquette/pull/103), we no longer need to do any dom replacement hacks in `DomWrapper`, this also has the nice side effect of resolving https://github.com/dojo/widget-core/issues/553. I have also added some supporting test cases. 

Resolves #553

### DIFF
--- a/src/util/DomWrapper.ts
+++ b/src/util/DomWrapper.ts
@@ -7,29 +7,26 @@ export interface DomWrapperOptions {
 	onAttached?(): void;
 }
 
-export type DomWrapper = Constructor<WidgetBase<VirtualDomProperties & WidgetProperties>>;
+export type DomWrapperProperties = VirtualDomProperties & WidgetProperties;
+
+export type DomWrapper = Constructor<WidgetBase<DomWrapperProperties>>;
 
 export function DomWrapper(domNode: Element, options: DomWrapperOptions = {}): DomWrapper {
-	return class extends WidgetBase<VirtualDomProperties & WidgetProperties> {
-		private _vNode: VNode;
-		private _firstRender = true;
+	return class extends WidgetBase<DomWrapperProperties> {
 
 		protected onElementCreated(element: Element, key: string) {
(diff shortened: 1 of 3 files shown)


src/util/DomWrapper.ts
```diff
@@ -7,29 +7,26 @@ export interface DomWrapperOptions {
...
+			const vNode = super.__render__() as VNode;
+			vNode.domNode = domNode;
+			return vNode;
 		}
 
 		protected render() {
-			const properties = this._firstRender ? {} as any : this.properties;
-			properties.bind && delete properties.bind;
+			const properties: any = this.properties;
+			delete properties.bind;
```
kitsonk: should we be doing a `delete` on a read only getter property?  That seem dangerous.  I didn't notice that before.  Would `const { bind, ...properties } = this.properties;` be a bit better?  I suspect no-unused will get in the way though.

matt-gadd: sorry yeah, i meant to spread it then delete it, good catch 👍 . When https://github.com/dojo/widget-core/issues/544, is fixed this won't be needed at all.

sebilasse:
The templates which are used by `CLI` have own `package.json` which should be updated as well (?)
For example https://github.com/dojo/cli-create-app/blob/master/src/templates/package.json

agubler:
@sebilasse They will be, when we bump `cli-create-app` to `beta2` or perhaps beforehand if there is a need for the created `app template` to use the `next` tag.

In fact there is an issue open on `@dojo/cli-create-app` to that very change https://github.com/dojo/cli-create-app/issues/82

PR #555, opened 2017-05-30
Support an array of `DNode`s from `render`

### LINKED ISSUE #533: Consider supporting an array of DNodes being returned from render
**Enhancement** 
It would be nice for widgets to be able to return an array of `DNodes` from a `render` call. This would allow authors to say create a container widget that creates a number of `li` elements, but don't have to create the outer `ul` element.

agubler:
**Type:** feature

**Description:**

Adds support to return an array of `DNode`s from a widget `render`, widgets used as a projector that return an array of nodes from the render will be automatically wrapped in a `span` (append and replace) or the tag name of the root DOM Element (merge).

Example Usage:

```ts
class MyListWidget extends WidgetBase {
    protected render(): DNode[] {
        return [
            v('li', { key: '1' }, [ 'Item One' ]),
            v('li', { key: '2' }, [ 'Item Two' ]),
            v('li', { key: '3' }, [ 'Item Three' ]),
            v('li', { key: '4' }, [ 'Item Four' ]),
            v('li', { key: '5' }, [ 'Item Five' ]),
            v('li', { key: '6' }, [ 'Item Six' ]),
            v('li', { key: '7' }, [ 'Item Seven' ])
        ];
    }
}
```
Resolves #533

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -411,9 +411,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			afterRenders.forEach((afterRenderFunction: Function) => {
 				dNode = afterRenderFunction.call(this, dNode);
 			});
-
 			const widget = this.dNodeToVNode(dNode);
-
 			this.manageDetachedChildren();
 			if (widget) {
 				this._cachedVNode = widget;
@@ -551,6 +549,8 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 	 * @param dNode the dnode to process
 	 * @returns a VNode, string or null
 	 */
+	private dNodeToVNode(dNode: DNode): VNode | string | null;
+	private dNodeToVNode(dNode: DNode[]): (VNode | string | null)[];
 	private dNodeToVNode(dNode: DNode | DNode[]): (VNode | string | null)[] | VNode | string | null {
 
 		if (typeof dNode === 'string' || dNode === null) {
@@ -558,9 +558,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 		}
 
 		if (Array.isArray(dNode)) {
-			return dNode.map((node) => {
-				return this.dNodeToVNode(node) as VNode | string | null;
-			});
+			return dNode.map((node) => this.dNodeToVNode(node));
 		}
 
 		if (isWNode(dNode)) {
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -168,7 +168,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 	/**
 	 * cachedVNode from previous render
 	 */
-	private _cachedVNode?: VNode | string;
+	private _cachedVNode?: (VNode | string)[] | VNode | string;
 
 	/**
 	 * internal widget properties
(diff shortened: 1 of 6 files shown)


src/WidgetBase.ts
```diff
@@ -168,7 +168,7 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 	/**
 	 * cachedVNode from previous render
 	 */
-	private _cachedVNode?: VNode | string;
+	private _cachedVNode?: (VNode | string | null)[] | VNode | string;
```
agubler: We use `VNode | string | null` often enough to export this as a known type from `interfaces`, any one have a suggestion for a name?

src/WidgetBase.ts
```diff
@@ -411,7 +411,9 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
 			afterRenders.forEach((afterRenderFunction: Function) => {
 				dNode = afterRenderFunction.call(this, dNode);
 			});
+
```
agubler: remove the random line breaks

src/mixins/Projector.ts
```diff
@@ -304,18 +304,13 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T)
...
 				}
 
-				result = {
-					domNode: null,
-					vnodeSelector: this._rootTagName,
-					properties: {},
-					children: undefined,
-					text: result === null ? undefined : result
-				};
+				result = h(this._rootTagName, {}, result);
```
agubler: turns out `h()` is good for creating `VNode`s and will do all the flattening and filtering needed 😄

src/WidgetBase.ts
```diff
@@ -549,12 +551,18 @@ export class WidgetBase<P extends WidgetProperties = WidgetProperties, C extends
...
-	private dNodeToVNode(dNode: DNode): VNode | string | null {
+	private dNodeToVNode(dNode: DNode | DNode[]): (VNode | string | null)[] | VNode | string | null {
 
 		if (typeof dNode === 'string' || dNode === null) {
 			return dNode;
 		}
 
+		if (Array.isArray(dNode)) {
+			return dNode.map((node) => {
+				return this.dNodeToVNode(node) as VNode | string | null;
```
agubler: Will always return a `VNode | string | null`, but the way the interface is need to cast - should probably add the correct overloads to make this play nicely.

src/d.ts
```diff
@@ -48,6 +48,7 @@ export function isHNode(child: DNode): child is HNode {
  */
 export function decorate(dNodes: DNode, modifier: (dNode: DNode) => void, predicate?: (dNode: DNode) => boolean): DNode;
 export function decorate(dNodes: DNode[], modifier: (dNode: DNode) => void, predicate?: (dNode: DNode) => boolean): DNode[];
+export function decorate(dNodes: DNode | DNode[], modifier: (dNode: DNode) => void, predicate?: (dNode: DNode) => boolean): DNode | DNode[];
```
agubler: Allow decorate to operate on a node that is typed as `DNode | DNode[]`, previous API only allowed the node to be typed as only one or the other

src/WidgetBase.ts
kitsonk: `NotWNode`? 😂 `VNodeChild`? or something with `Child` in it?  Man this is hard.

agubler: I cannot think of a good enough one that describes the type properly!

PR #556, opened 2017-05-30
remove 'bind' from public consumer properties

### LINKED ISSUE #544: Bind shouldn't be included onto the core consumer properties
**Bug**

When `bind` is automatically decorated in the `afterRender` it should be added outside of the consumer properties bag so that it is not available on `this.properties` and could be inadvertently passed through to children by spreading properties (`...properties`).

agubler:
**Type:** bug

**Description:**

`bind` is a special internal property that gets decorated onto nodes return from `render`, this should not be available for consumers from `this.properties` - `bind` is not on the API but if the properties are spread then `bind` will get passed on without the consumer knowing.

`bind` is now removed from widget properties as part of the `__setProperties__` process and only used to bind functions that are passed in the `properties` object.

Renamed the `Injector` property `bind` to `scope` as this is public so we do want all the standard `__setProperties__` processing.

Resolves #544

### DIFF
--- a/src/Container.ts
+++ b/src/Container.ts
@@ -20,7 +20,7 @@ export function Container<W extends WidgetBaseInterface>(
 			const { properties, children } = this;
 
 			return w<BaseInjector<any>>(name, {
-				bind: this,
+				scope: this,
 				render: () => w(component, properties, children),
 				getProperties,
 				properties,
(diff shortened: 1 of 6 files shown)


sebilasse:
@agubler This is nice. **Is it already in `next` on npm ?** 

For me many things got clear in the last days and now I am writing dojo2 widgets the whole day and this one is the last which is a bit hindering. E.g. when I do
```TypeScript
const {
      header, meta, description, extra,
      color, centered, fluid, link, inline, raised,
      href, image, onClick, className = null, ...rest
} = this.properties;
```
always have to remove `bind` from ...rest 

btw - et. al : thanks and cheers for the overall 🐉2 progress !

PR #578, opened 2017-06-16
Meta dimensions

rorticus:
**Type:** feature

**Description:**

Adds dimensional meta using the new meta system. We can use this to get exciting things like node position and size!

```typescript
render() {
    const dimensions = this.meta(Dimensions).get('someNode');

    console.log('node width: ', dimensions.width);
    // ...
}
```

### DIFF
--- a/src/meta/Dimensions.ts
+++ b/src/meta/Dimensions.ts
@@ -1,61 +1,87 @@
 import { Base } from './Base';
+import { deepAssign } from '@dojo/core/lang';
 
-export interface DimensionResults {
-	bottom: number;
-	height: number;
+export interface TopLeft {
 	left: number;
-	offsetHeight: number;
-	offsetLeft: number;
-	offsetTop: number;
-	offsetWidth: number;
-	right: number ;
-	scrollHeight: number;
-	scrollLeft: number;
-	scrollTop: number;
-	scrollWidth: number;
 	top: number;
+}
+
+export interface BottomRight {
+	bottom: number;
+	right: number ;
+}
+
+export interface Size {
+	height: number;
 	width: number;
 }
+
+export interface DimensionResults {
+	position: TopLeft & BottomRight;
+	offset: TopLeft & Size;
+	size: Size;
+	scroll: TopLeft & Size;
+}
+
+const defaultDimensions = {
+	offset: {
+		height: 0,
+		left: 0,
+		top: 0,
+		width: 0
+	},
+	position: {
+		bottom: 0,
+		left: 0,
+		right: 0,
+		top: 0
+	},
+	scroll: {
+		height: 0,
+		left: 0,
+		top: 0,
+		width: 0
+	},
+	size: {
+		width: 0,
+		height: 0
+	}
+};
 
 export class Dimensions extends Base {
 	public get(key: string): Readonly<DimensionResults> {
 		this.requireNode(key);
 
 		const node = this.nodes.get(key);
-		const {
-			bottom = 0,
-			height = 0,
-			left = 0,
-			right = 0,
-			top = 0,
-			width = 0
-		} = (node ? node.getBoundingClientRect() : {});
-		const {
-			scrollLeft = 0,
-			scrollTop = 0,
-			scrollHeight = 0,
-			scrollWidth = 0,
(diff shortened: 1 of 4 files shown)


src/meta/Dimensions.ts
```diff
@@ -0,0 +1,63 @@
+import { Base } from './Base';
+
+export interface DimensionResults {
```
matt-gadd: should we perhaps make these slightly nicer than the dom ones in terms of accessing? ie have separate objects for each, like `size`, `position`, `scroll` etc?

rorticus: I could see it either way. on the one hand we've already made it nicer because you don't have to remember where the properties are actually coming from, you just have a big property bag. On the other hand, it might be nice to separate them out into groups like that.

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,63 @@
...
+	scrollHeight: number;
+	scrollLeft: number;
+	scrollTop: number;
+	scrollWidth: number;
+	top: number;
+	width: number;
+}
+
+export class Dimensions extends Base {
+	get(key: string): Readonly<DimensionResults> {
```
agubler: can we be specific for the visibility? i.e. `public`?

rorticus: Done!

src/meta/Dimensions.ts
matt-gadd: I just cringe a bit when i see `marginTop` (which I think is missing btw?) or `scrollTop`, I can see why they're that way on DOM elements as they specifically do not nest properties. Would it not be immediately obvious what those properties belong to anyway if they were in groups?

rorticus: Yeah, like I said, I can see where you're coming from. I can go ahead and split these into groups. Is `marginTop` an actual property or is only on `style.marginTop`? If we start including style properties, maybe we should come up with a list of what we'll need to include, i.e., margins, padding, borders, etc. I think all of those affect position and size?

matt-gadd: ah yes i'm being silly, `marginTop` is a style property my bad!

rorticus: Yeah but when i think about it, _should_ we include some of those? :)

src/meta/Dimensions.ts
```diff
@@ -0,0 +1,63 @@
...
+	right: number ;
+	scrollHeight: number;
+	scrollLeft: number;
+	scrollTop: number;
+	scrollWidth: number;
+	top: number;
+	width: number;
+}
+
+export class Dimensions extends Base {
```
agubler: A really minor nit, but do you think this would be a little "nicer" with something like this?

```ts
const defaultDimensions = {
	bottom: 0,
	height: 0,
	left: 0,
	right: 0,
	top: 0,
	width: 0,
	scrollLeft: 0,
	scrollTop: 0,
	scrollHeight: 0,
	scrollWidth: 0,
	offsetLeft: 0,
	offsetTop: 0,
	offsetWidth: 0,
	offsetHeight: 0
};

export class Dimensions extends Base {
	public get(key: string): Readonly<DimensionResults> {
		this.requireNode(key);
		const node = nodes.get(key);
		if (!node) { 
			return { ...defaultDimensions };
		}
		const boundingDimensions = node.getBoundingClientRect();

		return {
			scrollLeft: node.scrollLeft,
			scrollTop: node.scrollTop,
			scrollHeight: node.scrollHeight,
			scrollWidth: node.scrollWidth,
			offsetLeft: node.offsetLeft,
			offsetTop: node.offsetTop,
			offsetWidth: node.offsetWidth,
			offsetHeigh: node.offsetHeight,
			bottom: boundingDimensions.bottom,
			height: boundingDimensions.height,
			left: boundingDimensions.left,
			right: boundingDimensions.right,
			top: boundingDimensions.top,
			width: boundingDimensions.width,
		};
	}
}
```

rorticus: Updated!

src/meta/Dimensions.ts
rorticus: OK, updated to include groups

PR #583, opened 2017-06-20
Add DomWrapper to README

### LINKED ISSUE #545: Document DomWrapper in the README
**Bug** 

We are missing any documentation of the DomWrapper in the README.

kitsonk:
**Type:** documentation

**Description:**

Resolves #545 

[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1191,7 +1191,7 @@ class IsTallMeta extends MetaBase {
 
 `DomWrapper` is used to wrap DOM that is created _outside_ of the virtual DOM system.  This is the main mechanism to integrate _foreign_ components or widgets into the virtual DOM system.
 
-The `DomWrapper` generates a class/constructor function that is then used as a widget constructor in the virtual DOM.  `DomWrapper` takes up to two arguments.  The first argument is the DOM node that it is wrapping.  The second is an optional set of options.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1186,6 +1187,47 @@ class IsTallMeta extends MetaBase {
...
+import WidgetBase from '@dojo/widget-core/WidgetBase';
+import createComponent from 'third/party/library/createComponent';
+
+export default class WrappedComponent extends WidgetBase {
+    private _component: any;
+    private _onAttach = () => {
+        this._component = createComponent(this._root);
+    }
+    private _root: HTMLDivElement;
+    private _WrappedDom: Constructor<WidgetBase<VirtualDomProperties & WidgetProperties>>;
```
agubler: I think you can use `DomWrapper` for this... it is a type and function.

kitsonk: 🙈 you are right!

README.md
```diff
@@ -1186,6 +1187,46 @@ class IsTallMeta extends MetaBase {
 }
 ```
 
+#### DomWrapper
+
+`DomWrapper` is used to wrap DOM that is created _outside_ of the virtual DOM system.  This is the main mechanism to integrate _foreign_ components or widgets into the virtual DOM system.
+
+The `DomWrapper` generates a class/constructor function that is then used as a widget class in the virtual DOM.  `DomWrapper` takes up to two arguments.  The first argument is the DOM node that it is wrapping.  The second is an optional set of options.
```
matt-gadd: should we mention that the component that `DomWrapper` produces allows you to pass `VirtualDomProperties` just like you would use on a normal `v` call?

PR #591, opened 2017-06-23
Remove events for properties and children changes

### LINKED ISSUE #589: Remove `children:changed` and `properties:changed` event
**Enhancement**

With the introduction of `RenderStates`, `invalidate` can be used everywhere instead of setting `this._dirty` which means that the events are no longer required schedule a render within the projector (the only place internally that the events are currently listened on).

This also removes unsupported (and undocumented) hooks into a widget, where instead exposed support lifecycle hooks should be used.

agubler:
**Type:** feature

**Description:**

Eventing for properties and children changes are not required. Replacing the listeners on these events by calling `scheduleRender` on invalidate and calling invalidate within the `__setProperties__` and `__setChlildren__` calls.

Resolves #589

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -1,6 +1,4 @@
-import { BaseEventedEvents, Evented } from '@dojo/core/Evented';
-import { EventedListenerOrArray } from '@dojo/interfaces/bases';
-import { Handle } from '@dojo/interfaces/core';
+import { Evented } from '@dojo/core/Evented';
 import { ProjectionOptions, VNodeProperties } from '@dojo/interfaces/vdom';
 import Map from '@dojo/shim/Map';
 import Promise from '@dojo/shim/Promise';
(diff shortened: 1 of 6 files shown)


kitsonk:
Not opposed to this change, but it will break `test-extras` when it lands.  Opened dojo/test-extras#52 for tracking it.

Also, should we even have `Evented` in `WidgetBase` at all?  It would only encourage people to _think_ it is appropriate to retain a reference to a widget instance and use `.on()` which is clearly anti-pattern for `widget-core`.  All of the widget core's higher order events are callbacks at properties at the moment.  While we have discussed providing a higher order eventing system, the way of registering is very unlikely to be based on `Evented`, right?

agubler:
This leaves the one required event for invalidating up the widget tree, so for now at least we do still need to extend `Evented`

agubler:
@kitsonk it doesn't actually break `test-extras` with these changes, the listeners on these events are not actually required (all unit tests pass).

I'll raise the changes on `test-extras`.

PR #597, opened 2017-06-26
Enable Injector to work with Widgets the use local registries and meta

### LINKED ISSUE #595: Meta/Registry does not work correctly with Widgets that have Themeable mixed in
**Bug**

`meta` leverages the in-built `@beforeRender` and `@afterRender` hooks to check that required nodes exist and also to clear the node map after the render.

This is fine unless it is used in a widget that leverages an injector style pattern to perform some logic or change behaviours (also using the `@beforeRender`), in `ThemeableMixin`'s case to swap the render function for one that returns an injected widget context instead.

For `meta` this means that the internal node map is actually cleared before the node check that is returned via the `@beforeRender` is executed and therefore throws an error when the expected node in the map no longer exists.

It is necessary to ensure that the clear operations is only called after the check function, not after every render (regardless of which render function was called) - This can be done by actually putting the clear operation in the `@beforeRender` check function before calling the original `render` function.

This is also an issue for Widgets that have the `RegistryMixin` applied and in the render are referencing a `widgetLabel` that has only been defined in a locally scoped registry that has been passed via properties.

One potential resolution would be for the `Injector` to override the `get registries` and return the `registries` from the instance passed via `scope` property (this is the original widget instance). Another approach could be to allow registries handlers to be merged, so that the injector can take the registries from the component it's "wrapped" and add them to it's own registry handler.

These errors are reproducible by adding `ThemeableMixin` to the test widgets in both the `meta` and `registry` unit tests.

agubler:
**Type:** bug

**Description:**

Attempt to enable use of Injectors to be used in a `@beforeRender()` with `meta` and local registries.

Resolves #595

### DIFF
--- a/src/Injector.ts
+++ b/src/Injector.ts
@@ -12,6 +12,7 @@ import {
 	DNode,
 	WidgetProperties
 } from './interfaces';
+import RegistryHandler from './RegistryHandler';
 
 export interface GetProperties {
 	<C, P extends WidgetProperties>(inject: C, properties: P): any;
@@ -65,7 +66,7 @@ export class Context<T = any> extends Evented {
 }
 
 export interface InjectorProperties extends WidgetProperties {
-	scope: any;
+	scope: WidgetBase;
 	render(): DNode;
 	getProperties?: GetProperties;
 	properties: any;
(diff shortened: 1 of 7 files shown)


src/Injector.ts
```diff
@@ -131,6 +132,10 @@ export function Injector<C extends Evented, T extends Constructor<BaseInjector<C
 
 			return render();
 		}
+
+		public get registries(): RegistryHandler {
```
agubler: Don't like this, this probably needs to be fixed up once the new registry changes are in.

PR #598, opened 2017-06-26
Cannot warn or use the previous DNodes when a supported DNode type of `undefined` is returned.

### LINKED ISSUE #585: Shouldn't report a warning for not returning DNodes from afterRender when the initial DNode is `null` or `undefined`
**Bug**

At the moment there is a warning if `DNode` is not returned from an `afterRender`, this should also take into account the initial `DNode` value and not report if the initial value is also `undefined` and `null`

agubler:
**Type:** feature

**Description:**

It was fine to warn when `undefined` was not a supported `DNode` value but now that `undefined` is a supported type, we cannot warn or automatically replace with the previous value.

Resolves #585

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -587,12 +587,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		const afterRenders = this.getDecorator('afterRender');
 
 		return afterRenders.reduce((dNode: DNode | DNode[], afterRenderFunction: AfterRender) => {
-			const updatedDNode = afterRenderFunction.call(this, dNode);
-			if (!updatedDNode) {
-				console.warn('DNodes not returned from afterRender, using existing dNodes');
-				return dNode;
-			}
-			return updatedDNode;
+			return  afterRenderFunction.call(this, dNode);
 		}, dNode);
 	}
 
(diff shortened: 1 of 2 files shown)


kitsonk:
Is it a valid/legitimate use case that the original `DNode` was of a value, but the return type is actually `undefined`?  Ergo, could it be done in cases where the `previous !== undefined && current === undefined`?

Tangental to this, but related, would it make sense at the decorator API to allowing passing of an optional argument that was something like `observe: boolean` which would _always_ ignore what is returned (and would even have an overload of `(callback: (...) => void, observe: true)`?

I think just "observing" the render is a legitimate use case.  You want to watch but you want to contract to not modify the render.

agubler:
> Is it a valid/legitimate use case that the original DNode was of a value, but the return type is actually undefined? Ergo, could it be done in cases where the previous !== undefined && current === undefined?

Okay, so I originally did this - however this would mean that if the `afterRender` legitimately returned `undefined` it would use the previous `dNode` value - this doesn't seem correct.

> I think just "observing" the render is a legitimate use case. You want to watch but you want to contract to not modify the render.

Personally think that the `afterRender` is only there to work with the decorating the dNodes, not to add any side affect like storing state.

kitsonk:
> Personally think that the `afterRender` is only there to work with the decorating the dNodes, not to add any side affect like storing state.

👍 Ok, if we find a legitimate use case for just being able to get notified of when something renders, we can revisit.  I will admit everything I have seen to date is some sort of _hack_ for something we wanted to support in another way.

PR #599, opened 2017-06-27
README Refresh

agubler:
**Type:** feature

**Description:**

A proposed re-work of the existing readme, with some sections re-written, re-worded or just re-ordered.

Attempted to add an overview headlines for widget-core, a "basics" section that describes Dojo 2 widget-core and it's basic usage, then advanced section that has more details.

Once we have a documentation site, I think most of these sections can be moved to there and expanded.

#### Missing Sections

- [x] Classes & Themes
- [ ] Registry
- [ ] Injector/Container
- [ ] SSR (should we include this yet?)
- [ ] Advanced section for theming regarding injecting theme


README.md
```diff
@@ -4,47 +4,41 @@
...
+
+-----
+
+ - [Installation](#installation)
+ - [Features](#features)
+ 	- [Basic Widgets](#basic-widgets)
+		- [Rendering a Widget in the DOM](#rendering-a-widget-in-the-dom)
+		- [Widgets and Properties](#widgets-and-properties)
+		- [Composing Widgets](#composing-widgets)
+		- [Extracting Widgets](#extracting-widgets)
```
kitsonk: Decomposing or destructuring instead of extracting?  Extracting seems like you are going to be taking references out of a larger widget.

README.md
```diff
@@ -56,186 +50,318 @@ npm install @dojo/has
...
+### Mixins
 
-In additional to the programatic functions `v` and `w`, widget-core optionally supports the use of the `jsx` syntax known as [`tsx`](https://www.typescriptlang.org/docs/handbook/jsx.html) in TypeScript.
+Dojo 2 makes use of mixins to decorate additional functionality and properties to existing widgets. Mixins provide a mechanism that allows loosely coupled design and composition of behaviors into existing widgets without having to change the base widget.
 
-To start to use `jsx` in your project the widgets need to be named with a `.tsx` extension and some configuration is required in the project's `tsconfig.json`:
+TypeScript supports mixins using a constructor type of `new (...args: any[]) => any;` that enables a class to be passed as a function argument and extended to add new functionality.
 
-Add the configuration options for `jsx`:
+Example mixin that adds method `setState` and `readonly` `state` property:
```
kitsonk: Do we really want to show a mixin we consider anti-pattern?  I can't think of another good example ATM though.

README.md
```diff
@@ -56,186 +50,318 @@ npm install @dojo/has
...
 ```
 
-Once the project is configured, `tsx` can be used in a widget's `render` function simply by importing the `tsx` function as `import { tsx } from '@dojo/widget-core/tsx';`
+Examples of Dojo 2 mixins can be seen with `ThemeableMixin` and `I18nMixin` that are described in [Classes & theming](#classes--theming) and [Internationalization](#internationalization) sections.
 
-```tsx
-class MyWidgetWithTsx extends WidgetBase<MyProperties> {
-	protected render(): DNode {
-		const { clear, properties: { completed, count, activeCount, activeFilter } } = this;
+### Classes & Theming
```
kitsonk: Styling and Theming (or preface with CSS Classes, to avoid confusion with ES Classes?)

README.md
```diff
@@ -56,186 +50,318 @@ npm install @dojo/has
...
+}
 ```
 
-**All** widgets should extend from this class.
+## Key Principles
+
+These are some of the **important** principles to keep in mind when creating and using widgets:
 
-#### The 'properties' lifecycle
+1. The widget's *`__render__`*, *`__setProperties__`*, *`__setChildren__`* functions should **never** be called or overridden
```
kitsonk: some of these are sentences with a full stop, some aren't... should be consistent

Also, if you expounded on _why_ some of these are principles... like:

1. The widget's *`__render__`*, *`__setProperties__`*, *`__setChildren__`* functions should **never** be called or overridden.  These are the internal methods of the widget APIs and their behaviour can change in the future, causing regressions in your application.

It is human nature to do something you are told not to do without justification. 😈

README.md
```diff
@@ -381,464 +509,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
+        if (node) {
+            return node.offsetHeight >= minHeight;
+        }
+        return false;
+    }
 }
 ```
 
-The theme can be applied to individual widgets or to a project and property passed down to its children.
+### Tsx Support
```
kitsonk: `TSX support`? JSX always appears all-caps: https://facebook.github.io/react/docs/jsx-in-depth.html and https://www.typescriptlang.org/docs/handbook/jsx.html, also even TypeScript seems to just refer to it as JSX.

README.md
agubler: I thought exactly the same! But couldn't think of another simple one. I guess it is not so much of an anti-pattern per se but something that we don't want to support out of widget-core.

README.md
agubler: Yup I think that is a nice idea.

README.md
agubler: I'm happy to just use `JSX`....

agubler:
@dylans do you fancy giving this a review?

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 
 You can also use the [dojo cli](https://github.com/dojo/cli) to create a complete Dojo 2 skeleton application.
 
 ## Features
 
-Constructing your own widgets (Custom widgets) is simple and straightforward.
-The smallest `@dojo/widget-core` example looks like this:
+### Basic Widgets
+
+Dojo 2 applications use a concept called Virtual DOM (vdom) to represent what should be shown on the screen. These vdom nodes are plain javascript objects that are not expensive to create, unlike browser DOM elements. Dojo 2 uses these vdom elements to synchronize and update the browser DOM so that application shows the expected view.
```
dylans: Virtual DOM (vdom) should eventually link to https://github.com/dojo/dojo.io/blob/master/site/source/tutorials/working_with_virtual_dom.md when it is added to the dojo.io website

dylans: Dojo 2 applications use a concept called Virtual DOM (vdom) to represent what should be shown on the screen  -> Dojo 2 applications use the Virtual DOM (vdom) paradigm to represent the content that should be shown in the view

dylans: javascript - JavaScript

dylans: that are not expensive to create, unlike browser DOM elements -> that are more efficient to create from a performance perspective than browser DOM elements

dylans: so that application shows the expected view -> so that the application shows the expected view

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+The following will create a `HNode` that represents a simple `div` DOM element, with a text node child: `Hello, Dojo 2!`:
 
 ```ts
-class MyWidget extends WidgetBase<WidgetProperties> {
-	render() {
-       return v('h1', [ 'Hello, Dojo!' ]);
+v('div', [ 'Hello, Dojo 2!' ])
+```
+
+The second vdom type, `WNode`s represent widgets. A widget is a class that extends `WidgetBase` from `@dojo/widget-core/WidgetBase` and implements a `render` function that returns one of the Dojo 2 vdom types (known as a `DNode`). Widgets are used to represent reusable, independent sections of a Dojo 2 application.
```
dylans: `WNode`s represent widgets -> `WNode`, represent widgets

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+class HelloDojo extends WidgetBase {
+	protected render(): DNode {
+		return v('div', [ 'Hello, Dojo 2!' ]);
 	}
 }
+```
+
+#### Rendering a Widget in the DOM
+
+To display your new component in the browser you will need to decorate it with some functionality needed to "project" the widget into the browser. This is done using the `ProjectorMixin` from `@dojo/widget-core/mixins/Projector`.
```
dylans: "in the browser " -> "in the view " (given that SSR or headless browsers may also be used, seems useful to be more general?

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+```ts
+const Projector = ProjectorMixin(HelloDojo);
+const projector = new Projector();
 
-const Projector = ProjectorMixin(MyWidget);
+projector.append();
+```
+
+By default the projector will attach the widget to the `document.body` in the DOM, but this can be overridden by passing a reference to the Element to attach to.
```
dylans: "to the Element to attach to" -> "to the preferred parent DOM Element"

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+Consider the following in your HTML file:
+
+```html
+<div id="my-app"></div>
+```
+
+You can target this Element like so:
```
dylans: like so: -> :

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+const Projector = ProjectorMixin(HelloDojo);
 const projector = new Projector();
 
 projector.append(root);
 ```
 
-This code renders a `h1` element onto the page, that says "Hello, Dojo!".
+#### Widgets and Properties
+
+We have created a widget used to project our `HNode`s into the DOM, however widgets can be composed of other widgets and `properties` are used to determine if a widget needs to be re-rendered.
```
dylans: `properties` are used -> `properties` which are used

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 
-### Overview
+Properties are available on the the widget instance, defined by an interface and passed as a generic to the `WidgetBase` class when creating your custom component. The properties interface should extend the base `WidgetProperties` provided from `@dojo/widget-core/interfaces`:
```
dylans: generic -> `generic` (and perhaps worth a link to https://www.typescriptlang.org/docs/handbook/generics.html ?)

README.md
```diff
@@ -4,47 +4,41 @@
...
-        - [Meta](#meta)
-		- [DomWrapper](#domwrapper)
-    - [Key Principles](#key-principles)
-    - [API](#api)
+widget-core is library to create powerful, composable user interface widgets.
+
+* **Reactive & unidirectional:** Widget-core follows core reactive principles to ensure predictable consistent behavior.
+* **Encapsulated Widgets:** Create independent encapsulated widgets that can be wired together to create complex and beautiful user interfaces.
+* **DOM Abstractions:** Widget-core provides API abstractions to avoid needing to access or manipulate the DOM outside of the reactive render life-cycle.
+* **I18n & Themes:** Widget-core provides core mixins to enable internationalization and theming support for your components.
```
dylans: components -> widgets (I would suggest saving the word component for when we talk about web components)

README.md
dylans: component -> widget (I would suggest saving the word component for when we talk about web components)

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
-We also make use of a VirtualDOM (VDOM) in Dojo 2.
-In order to interact with our VDOM, you need to pass it [HyperScript](https://github.com/dominictarr/hyperscript).
-In Dojo 2 we provide 2 functions that make interacting with the VDOM, easy and intuitive: `v` and `w`.
+		return v('div', [ `Hello, ${name}` ]);
+	}
+}
+```
 
-### `v` & `w`
+New properties are compared with the previous properties to determine if a widget requires re-rendering, by default Dojo 2 uses the `auto` diffing strategy, that performs a shallow comparison for objects and arrays, ignores functions (except classes that extend WidgetBase) and a reference comparison for all other values.
```
dylans: re-rendering, by -> re-rendering. By

dylans: WidgetBase -> `WidgetBase`

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 
-These functions express structures that will be passed to the VDOM.
+#### Composing Widgets
 
-`v` creates nodes that represent DOM tags, e.g. `div`, `header` etc.
-This function allows Dojo 2 to manage lazy hyperscript creation and element caching.
+As mentioned, often widgets are composed of other widgets in their `render` output. This promotes widget reuse across an application (or multiple applications) and helps enables widget best practices.
```
dylans: helps enables -> promotes

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 ```ts
-import { DNode, HNode, WNode } from '@dojo/widget-core/interfaces';
+const Projector = ProjectorMixin(App);
+const projector = new Projector();
+
+projector.append();
 ```
 
-#### `v`
+**Note:** Widgets must return a single top level `DNode` from the `render` method, which is why the `Hello` widgets were wrapped in `div`.
```
dylans: in `div` -> in a `div` element

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+		return v('ul', { classes: { list: true } }, items.map((item) => {
+			const { id, highlighted, content } = item;
+			const classes = highlighted ? { highlighted: true } : { highlighted: false };
+
+			return v('li', { key: id, classes }, [ content ]);
+		});
+	}
+}
 ```
+The `List` widget does works as expected and displays the list in the browser but is hard to change, add functionality and reuse parts.
```
dylans: does works -> works

dylans: "but is hard to change, add functionality and reuse parts." -> "but is difficult to reuse, modify, and/or extend"

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+**Note:** When working with children arrays with the same type of widget or Element it is important to add a unique `key` property or attribute so that Dojo 2 can identify the correct node when updates are made.
```
dylans: Element it -> Element, it

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 ```
 
-As well as interacting with the VDOM by passing it HyperScript, you can also pass it Dojo 2 Widgets or Custom Widgets using the `w` function.
+If we try to use the `onItemClick` function in the current `List` widget, we would need to wrap it in another function in order to pass the item's `id`.
 
-#### `w`
+This would mean a new function would be created every render but Dojo 2 does not support changing listener functions after the first render and this would **error**.
 
-The following code creates a widget using the `widgetConstructor` and `properties`.
+To resolve this, the list item can be extracted:
```
dylans: extracted -> extracted into a separate widget

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+		const { id, content, highlighted } = this.properties;
+		const classes = highlighted ? { highlighted: true } : { highlighted: false };
+
+		return v('li', { key: id, classes, onclick: this.onClick }, [ content ]);
+	}
+}
 ```
 
-The following code creates a widget using the `widgetConstructor`, `properties` and `children`
+Using the `ListItem` we can simplify the `List` slightly and also add the on click functionality that we required:
```
dylans: on click -> onclick

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 ```ts
-w(WidgetClass, properties);
-w(WidgetClass, properties, children);
+// interface for the extended API
+interface StateMixin {
+	readonly state: Readonly<any>;
+	setState(state: any): void;
+}
+
+// function that accepts a class that extends WidgetBase and returns the extended class with the StateMixin
```
dylans: WidgetBase -> `WidgetBase` and StateMixin -> `StateMixin` ?

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+Examples of Dojo 2 mixins can be seen with `ThemeableMixin` and `I18nMixin` that are described in [Classes & theming](#classes--theming) and [Internationalization](#internationalization) sections.
 
-### tsx
+### Styling & Theming
 
-In additional to the programatic functions `v` and `w`, widget-core optionally supports the use of the `jsx` syntax known as [`tsx`](https://www.typescriptlang.org/docs/handbook/jsx.html) in TypeScript.
+#### Overview
 
-To start to use `jsx` in your project the widgets need to be named with a `.tsx` extension and some configuration is required in the project's `tsconfig.json`:
+Dojo 2 widget-core provides `ThemeableMixin` to decorate a widget with theming functionality and a `@theme` decorator to specify the classes available to the widget. Both of which come from `@dojo/widget-core/mixins/Themeable`.
```
dylans: Both of which come from -> Both `ThemeableMixin` and `@theme` are provided by

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
 
-Add the configuration options for `jsx`:
+To specify the classes for a widget, an interface needs to be imported with named exports for each class and passed to the `@theme` decorator. Importing the interface provides intellisense for the class names and passing this via the `@theme` decorator tells the `ThemeableMixin` which classes can be themed.
```
dylans: intellisense -> IntelliSense / auto-complete

dylans: tells -> informs

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+```typescript
+export const classNameOne: string;
+export const classNameTwo: string;
 ```
-"jsx": "react",
-"jsxFactory": "tsx",
+</details>
+
+
+**Important:** at runtime a javascript file is required to provide the processed css class names.
```
dylans: javascript -> JavaScript

dylans: css -> CSS

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+The `ThemeableMixin` provides a method available on the instance `this.classes()` that consumes string class names and converts them into the expected class object when a widget is rendered. Classes that are passed to `this.classes()` are made available to be themed, as described [here](applying-a-theme).
```
dylans: described [here](applying-a-theme) -> described in the [applying a theme](applying-a-theme) section

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+However, it is not always desirable to allow consumers to override styling that may be required for a widget to function correctly, these classes are known as "fixed". Fixed classes are passed by using a chained function `this.classes().fixed()`.
```
dylans: correctly, these classes -> correctly. These classes which should not be overridden

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+	protected render(): DNode {
+		return v('div', { classes: this.classes(css.root).fixed(css.rootFixed) });
+	}
+}
 ```
 
-Include `.tsx` files in the project:
+#### Writing a theme
+
+Themes are typescript modules that export an object that contains css-modules files keyed by a widgets css file name.
```
dylans: typescript -> TypeScript

dylans: css file -> CSS file

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+// myTheme/theme.ts
+import * as myWidget from './styles/myWidget.m.css';
+
+export default {
+	'myWidget': myWidget
+}
 ```
 
-Once the project is configured, `tsx` can be used in a widget's `render` function simply by importing the `tsx` function as `import { tsx } from '@dojo/widget-core/tsx';`
+In the above example, the new `root` class provided by the theme for `myWidget` will replace the `root` class that was provided by the original `myWidget.m.css`. This means the `myWidget` will now have color set to blue, but will no longer receive the styles from the `root` class in the original css.
```
dylans: have color -> have its color

dylans: css -> CSS

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
-				<span classes={this.classes(css.count)}>
-					<strong>{`${activeCount}`}</strong>
-					<span>{`${count}`}</span>
-				</span>
-				<TodoFilter activeFilter={activeFilter} />
-				{ completed ? ( <button onclick={clear} /> ) : ( null ) }
-			</footer>
-		);
-	}
+To apply a theme to a widget, simply require it as a module and pass it to a widget via its properties. It is important to ensure that any child widgets created within your widgets render function are passed the `theme` or they will not be themed.
```
dylans: your widgets render -> you widget's render

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+// ...
+render() {
+	return w(TabPanel, { theme: myTheme } });
 }
 ```
 
-**Note:** Unfortunately `tsx` is not directly used within the module so will report as an unused import so would be needed to be ignored by linters.
+#### Passing extra classes
+
+Sometimes you may need to apply positioning or layout styles to a child widget. As it is not possible to pass `classes` directly to `WNodes` they provide an `extraClasses` property that can be used to target themeable classes used within its `render` function. In most cases this should only target the `root` class and apply positioning adjustments. The classes passed via `extraClasses` are outside of the theming mechanism and thus will not be effected by a change in `theme`.
```
dylans: directly to `WNodes` they provide an `extraClasses` property that can be used to target themeable classes used within its `render` function. -> directly to virtualized widget nodes. `WNodes` thus provide an `extraClasses` property to target themeable classes within its `render` function.

README.md
```diff
@@ -56,186 +50,416 @@ npm install @dojo/has
...
+
+### Internationalization
 
-### Writing Custom Widgets
+Widgets can be internationalized by adding the `I18nMixin` mixin from `@dojo/widget-core/mixins/I18n`. [Message bundles](https://github.com/dojo/i18n) are localized by passing them to `localizeBundle`.
 
-The `WidgetBase` class provides the functionality needed to create Custom Widgets.
-This functionality includes caching and widget lifecycle management.
+If the bundle supports the widget's current locale, but those locale-specific messages have not yet been loaded, then the default messages are returned.
+The widget will be invalidated once the locale-specific messages have been loaded.
```
dylans: loaded, triggering a re-render with the localized message content.

README.md
```diff
@@ -305,9 +529,11 @@ class MyWidget extends WidgetBase<WidgetProperties> {
 }
 ```
 
-#### Render Lifecycle
+### Widget Registry
+
+Section Coming Soon!
```
dylans: TODO: Section

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
 }
 ```
 
-#### Projector
+### DOM Wrapper
```
dylans: DOM Wrapper -> `DOMWrapper`

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-There are 3 ways that a projector widget can be added to the DOM - `.append`, `.merge`, `.replace`, or `.sandbox`, depending on the type of attachment required.
+|Name|Description|
+|-|-|
+|`onAttached`|A callback that is called when the wrapped DOM is flowed into the virtual DOM|
 
- - `append`  - Creates the widget as a child to the projector's `root` node
- - `merge`   - Merges the widget with the projector's `root` node
- - `replace` - Replace the projector's `root` node with the widget
- - `sandbox` - Create a document fragment as the projector's `root` node
+As an example, we want to integrate a 3rd party library where we need to pass the component factory a _root_ element and then flow that into our virtual DOM.  In this situation we don't want to create the component until the widget is being flowed into the DOM, so `onAttached` is used to perform the creation of the component:
```
dylans: As an example, we -> For example, if we

dylans: component factory ... update to more recent terminology

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
+    public render() {
+        return w(this._WrappedDom, { key: 'wrapped' });
+    }
+}
 ```
 
-###### .toHtml()
-
-`.toHtml()` returns a string which represents the current render of the `Projector`.  While it can be used with any attachment mode, it is most effective when using `.sandbox()`, as this mode operates synronously, ensuring that the string returned accuretly represents the current rendered DOM structure.  Building on the example from above:
+The properties which can be set on `DomWrapper` are the combination of the `WidgetBaseProperties` and the `VirtualDomProperties`, which means effectively you can use any of the properties passed to a `v()` node and they will be applied to the wrapped DOM node.  For example the following would set the classes on the wrapped DOM node:
```
dylans: classes -> `classes`

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-*Handler passed via properties*
-
-```ts
-class MyWidget extends WidgetBase<WidgetProperties> {
-	onClick(): void {
-		this.properties.mySpecialFunction();
-	}
-}
-```
+If an HTML node is required to calculate the meta information, a sensible default will be returned and your widget will be automatically re-rendered to provide more accurate information.
```
dylans: Feels like we should add an extra sentence as this is really beneficial, e.g. something like "Meta allows us to preserve the concepts of reactivity in a virtual DOM where traditional vdom approaches would simply provide a reference to a DOM node" but something better?

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-tabPanel
-├── createTabPanel.ts
-└── styles
-    └── tabPanel.css
-```
+##### Implementing Custom Meta
 
-The `baseClasses` css must contain a complete set of all of the classes you wish to apply to a widget as all theme and extra classes are limited by the classnames made available here.
-Classnames are locally scoped as part of building a Dojo application. A theme `key` is generated at build time to locate the themes for each class where a theme is set.
+You can create your own meta if you need access to DOM nodes.
```
dylans: "you need access to DOM nodes" -> perhaps "if you want to preserve reactive principles with a DOM node"?

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-
-```ts
-import { WidgetProperties } from '@dojo/widget-core/interfaces';
-import { WidgetBase } from '@dojo/widget-core/WidgetBase';
-import { ThemeableMixin, ThemeableProperties, theme } from '@dojo/widget-core/mixins/Themeable';
-import * as baseClasses from './styles/tabpanel.css';
+Meta classes are provided with a few hooks into the widget, passed to the constructor:
 
-interface MyThemeableWidgetProperties extends WidgetProperties, ThemeableProperties;
+* `nodes` - A map of `key` strings to DOM elements. Only `v` nodes rendered with `key` properties are stored.
```
dylans: Only `v` nodes rendered with `key` properties are stored. -> reword to make this clearer

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
+* `requireNode` - A method that accept a `key` string to inform the widget it needs a rendered DOM element corresponding to that key. If one is available, it will be returned immediately. If not, the widget will be re-rendered and if the node does not exist on the next render, an error will be thrown.
```
dylans: that accept -> that accepts

dylans: error -> **error**

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-				v('li', { classes: this.classes(tab) }, [ 'tab1' ])
-				// ...
-			]);
-	}
-}
-```
+Extending the base class found in `meta/Base` will automatically add these hooks to the class instance as well as providing a `has` method:
 
-##### Applying a Theme
+* `has(key: string)` - A method that returns `true` if the DOM element with the passed key exists in the rendered DOM.
```
dylans: key -> `key`

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
 
-The theming system supports injecting a theme that is configured externally to the usual mechanism of passing properties down the widget tree.
+In addition to creating widgets functionally using the `v()` and `w()` functions from `@dojo/widget-core/d`, Dojo 2 optionally supports the use of the `jsx` syntax known as [`tsx`](https://www.typescriptlang.org/docs/handbook/jsx.html) in TypeScript.
 
-This is done using a `ThemeInjectorContext` instance that is passed to the `Injector` mixin along with the `ThemeInjector` class. Once the theme injector is defined in the registry, the `theme` can be changed by calling the `ThemeInjectorContext#set(theme: any)` API on the instance of the injector context.
-
-```ts
-// Create the singleton injector context
-const themeInjectorContext = new ThemeInjectorContext(myTheme);
+To start to use `jsx` in your project the widgets need to be named with a `.tsx` extension and some configuration is required in the project's `tsconfig.json`:
```
dylans: "to use `jsx` in your project the widgets" -> "using `jsx` in your project, widgets"

README.md
```diff
@@ -381,464 +607,224 @@ class MyBaseClass extends WidgetBase<WidgetProperties> {
...
-			}),
-			w(Button, {
-				// Passing a formatted message string to a child widget.
-				label: messages.format('itemCount', { count: 2 })
-			})
-		]);
-	}
-}
-```
+**Note:** Unfortunately `tsx` is not directly used within the module so will report as an unused import so would be needed to be ignored by linters.
```
dylans: "so will report as an unused import so would be needed to be ignored by linters." -> "so the `tsx` module will report as an unused import, and will need to be ignored by linters."

README.md
```diff
@@ -1016,188 +1002,9 @@ The initialization function is run from the context of the HTML element.
...
-2. Except for projectors, you should **never** need to deal directly with widget instances
-3. Hyperscript should **always** be written using the @dojo/widget-core `v` helper function
-4. **Never** set state outside of a widget instance
-5. **Never** update `properties` within a widget instance
+It should be noted that children nodes are removed from the DOM when widget instantiation occurs, and added as children to the widget instance.
 
-### API
+## API
 
 [API Documentation](http://dojo.io/api/widget-core/v2.0.0-alpha.28/)
```
dylans: Update to https://dojo.io/api/widget-core/v2.0.0-beta1.6/

agubler:
@dylans you have requested changes on this, feels like we could get this version in and iterate (it's been open for a while)

PR #601, opened 2017-06-27
Support original `afterRenders` when replacing render with an `Injector`

### LINKED ISSUE #600: I18nMixin unit test fail using a widget with ThemeableMixin applied and a theme in the registry
@agubler commented on [Tue Jun 27 2017](https://github.com/dojo/core/issues/343)

**Bug**

6 unit tests fail when using a widget that has `ThemeableMixin` applied and a `theme` is available in the global registry.

To reproduce add `ThemeableMixin` to the `I18n` test widget, as follows:

```ts
class Localized extends I18nMixin(ThemeableMixin(WidgetBase))<I18nProperties> {}
```

And ensure that a theme has been registered in the widgets registry (usually global).

agubler:
**Type:** bug

**Description:**
Further support in `WidgetBase` and `Injector` to enable correct `afterRenders` to be executed when the original `DNode`s are returned from the `render` function passed to the `Injector`.

Breaking change as updated `this.registries` to `this.getRegistries()` so that `Injector` can override and call the `super.getRegistries()` with the original widgets scope passed as the properties.

Resolves #600

### DIFF
--- a/src/Injector.ts
+++ b/src/Injector.ts
@@ -1,7 +1,6 @@
 import { assign } from '@dojo/core/lang';
 import { Evented } from '@dojo/core/Evented';
 import {
-	afterRender,
 	diffProperty,
 	WidgetBase
 } from './WidgetBase';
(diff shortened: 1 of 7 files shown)


agubler:
Weird codecov issues.

PR #604, opened 2017-06-27
Support typings for ThemeableMixin extraClasses

### LINKED ISSUE #603: Enable typings for `extraClasses`
**Enhancement**

It would be good if the keys of the `extrasClasses` property for `ThemeableMixin` were typed to the injected css keys.

Given the following css interface and widget:

```ts
// css interface
const classOne: string;
```

```ts
import * as css from './the/file/another';

interface MyProperties extends ThemeableProperties {
    anotherProperty: string;
}

class MyWidget extends WidgetBase<MyProperties> {}
```

The only supported key for `extraClasses` would be `classOne` but currently anything can be passed and a widget user will not get any sort of warning.

```ts
// this will not error
w(MyWidget, { extraClasses: { unsupportedClassName: 'my-extra-class' });
```

We could enable a widget author to specify the shape of the css allowed as a generic to `ThemeableMixin` meaning that we can guard `extraClasses` correctly.

```ts
import * as css from './the/file/another';

interface MyProperties extends ThemeableProperties<typeof css> {
    anotherProperty: string;
}

class MyWidget extends WidgetBase<MyProperties> {}
```

```ts
// this will now error
w(MyWidget, { extraClasses: { unsupportedClassName: 'my-extra-class' });
```

agubler:
**Type:** feature

**Description:**

Support typings for `extrasClasses` property on `ThemeableProperties`.

Resolves #603

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -325,7 +325,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		return this._properties;
 	}
 
-	public __setProperties__(originalProperties: P): void {
+	public __setProperties__(originalProperties: this['properties']): void {
 		const { bind, ...properties } = originalProperties as any;
 		const changedPropertyKeys: string[] = [];
 		const allProperties = new Set([...Object.keys(properties), ...Object.keys(this._properties)]);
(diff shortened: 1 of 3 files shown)


src/WidgetBase.ts
```diff
@@ -325,7 +325,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		return this._properties;
 	}
 
-	public __setProperties__(originalProperties: P): void {
+	public __setProperties__(originalProperties: this['properties']): void {
```
kitsonk: cool! 👍

src/mixins/Themeable.ts
```diff
@@ -26,10 +26,10 @@ export type ClassNames = {
 /**
  * Properties required for the themeable mixin
  */
-export interface ThemeableProperties extends WidgetProperties {
+export interface ThemeableProperties<T = any> extends WidgetProperties {
```
kitsonk: Is `any` the true default of this, nothing more narrow, even a simple `{ [key: string]: any; }`?

PR #605, opened 2017-06-27
Include ThemeableMixin by default

### LINKED ISSUE #581: Consider applying core mixins to WidgetBase by default
**Enhancement**

The main core mixins, `Themeable`, `I18n` and `Registry` provide functionality that could be automatically mixed in to `WidgetBase` by default. This would remove possible friction for consumer to need to deal with the mixin concept early on in Dojo 2 adoption.

The mixins are all additive functionality and shouldn't impact consumers who only need core `WidgetBase` functionality.

This would leave:

1) `Stateful`, which I think we should consider removing in favour of promoting use of typed class properties for any required internal state.
2) `Projector`, which would remain as an optional mixin to decorate a widget with functionality required to attach to the DOM.

agubler:
**Type:** feature

**Description:**

Proposal includes `ThemeableMixin` by default for `WidgetBase`. Renames the original `WidgetBase.ts` to `Base.ts` and adds a new `WidgetBase.ts` module that imports `Base` plus other named exports from `Base.ts` and `ThemeableMixin` and exports `ThemableMixin(Base)` as `WidgetBase` (and also re-exports all the other exports from `Base.ts`).

This should be backwards compatible with existing usages of `WidgetBase` even if they have `ThemeableMixin` applied.

This requires #601 to fix the failing unit test.

Resolves #581


agubler:
Open to any better suggestions on naming :)

kitsonk:
Not that I have been a fan of it, but it does seem to be the case here... what if we made _WidgetBase_ and abstract class and named it _Widget_ and make the `.render()` abstract.  That way no one would be able to use it (in TypeScript) without extending it and implementing `.render()`.  I think we _might_ be scared of using _Widget_ because people might actually use it/abuse it.  Looking at a few other frameworks, they do use things like `Element` or `Component` for that _kernel_.

kitsonk:
Ignore me... I didn't notice the full context that `WidgetBase` is still the main thing that end user will utilise.

agubler:
The merges for this PR would be "tricky" so closing and will re-raise the change if there is consensus to do so.

PR #606, opened 2017-06-30
Fix compile errors from latest dependencies

agubler:
**Type:** bug

**Description:**

Fix compile errors from dep version upgrades.


tests/unit/RegistryHandler.ts
```diff
@@ -64,7 +66,7 @@ registerSuite({
 		assert.isTrue(registryHandler.has(bar));
 	},
 	'get'() {
-		const promise = new Promise((resolve) => {
+		const promise = new Promise<WidgetBaseConstructor>((resolve) => {
```
kitsonk: _sigh_ I thought I had the typings right with 2.4.1 for that to work without needing to cast...

PR #608, opened 2017-07-05
Remove aggressive node requires

### LINKED ISSUE #607: Meta is too aggressive when marking nodes required
**Bug** 

Nodes are being added to `_requireNodes` when calling `has` and in `requireNode` when a node already exists in rendered DOM.

This creates problems in `has` when the node using that key may or may not be rendered and the code is merely inquiring as to its existence. By marking that key as being required, an incorrect error will be thrown.

We also run into a problem with `requireNode` when, during a render, we are able to successfully extract meta information from a node during a render but do not return that node's representation in the return value from the render.

pottedmeat:
**Type:** bug

**Description:**

Resolves #607

### DIFF
--- a/src/meta/Base.ts
+++ b/src/meta/Base.ts
@@ -17,7 +17,6 @@ export class Base {
 	}
 
 	public has(key: string): boolean {
-		this.requireNode(key);
 		return this.nodes.has(key);
 	}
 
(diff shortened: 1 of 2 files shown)


agubler:
@pottedmeat I am not sure I understand the purpose of this change, especially in relation to adding a `forcedInvalidate`, if this is now outdated or no longer required could you please close? 

Otherwise could you please provide more information about the need for some of these changes. Thanks!

PR #615, opened 2017-07-14
Adding deferred properties

rorticus:
**Type:** feature

**Description:**

Throwing this out here to get some feedback. Currently, deferred properties let you apply styles **after** a virtual node is rendered.

For example, this will create a `div` that is as wide as it is tall:

```typescript
render() {
    return v('div', {
        key: 'root',
        styles: {
            width: asyncProperty(() => {
                const dim = this.meta(Dimensions).get('root');
                return dim.size.height + 'px';
            })
        }
    }, ['some text']);
}
```

Note that the height is calculated on the height of the node _without a `width` style_. Also note that once the property has been applied, you've just fixed the width of the node and subsequent calls will return your first value.

Take a look at the `deferredProperties.html` example to see how it would work in the case of a panel component that animates to show the current panel.

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -52,6 +52,12 @@ interface ReactionFunctionConfig {
 	reaction: DiffPropertyReaction;
 }
 
+interface DeferredProperty {
+	object: 'properties' | 'styles';
+	propertyName: string;
+	callback: () => any;
+}
+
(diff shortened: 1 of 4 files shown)


rorticus:
Closing this in favor of @pottedmeat 's amazing meta changes

PR #619, opened 2017-07-25
Benchmark tweaks

agubler:
**Type:** feature

**Description:**

Some performance tweaks based on the `js-framework-benchmark` results.

Before and after (tweak) results running on my mac with 20 iterations.

interactive_results

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -159,11 +159,6 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	 */
 	private _decoratorCache: Map<string, any[]>;
 
-	/**
-	 * set of render decorators
-	 */
-	private _renderDecorators: Set<string>;
-
 	private _registries: RegistryHandler;
 
 	/**
@@ -194,7 +189,6 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this._properties = <P> {};
 		this._cachedChildrenMap = new Map<string | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>();
 		this._diffPropertyFunctionMap = new Map<string, string>();
-		this._renderDecorators = new Set<string>();
 		this._bindFunctionPropertyMap = new WeakMap<(...args: any[]) => any, { boundFunc: (...args: any[]) => any, scope: any }>();
 		this._registries = new RegistryHandler();
 		this._registries.add(registry);
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -5,7 +5,7 @@ import '@dojo/shim/Promise'; // Imported for side-effects
 import Set from '@dojo/shim/Set';
 import WeakMap from '@dojo/shim/WeakMap';
 import { isWNode, registry, v } from './d';
-import { auto, reference } from './diff';
+import { auto } from './diff';
 import {
 	AfterRender,
 	BeforeRender,
@@ -117,7 +117,6 @@ export function handleDecorator(handler: (target: any, propertyKey?: string) => 
 /**
  * Main widget base for all widgets to extend
  */
-@diffProperty('bind', reference)
 export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends Evented implements WidgetBaseInterface<P, C> {
 
 	/**
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -87,6 +87,7 @@ export function beforeRender(method?: Function) {
 export function diffProperty(propertyName: string, diffFunction: DiffPropertyFunction, reactionFunction?: Function) {
 	return handleDecorator((target, propertyKey) => {
 		target.addDecorator(`diffProperty:${propertyName}`, diffFunction);
+		target.addDecorator('registeredDiffProperty', propertyName);
 		if (reactionFunction || propertyKey) {
 			target.addDecorator('diffReaction', {
 				propertyName,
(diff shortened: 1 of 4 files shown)


src/WidgetBase.ts
```diff
@@ -328,41 +292,57 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+			}
+			checkedProperties.push(propertyName);
 			const previousProperty = this._properties[propertyName];
-			const newProperty = properties[propertyName];
-			const diffFunctions: DiffPropertyFunction[] = this.getDecorator(`diffProperty:${propertyName}`);
-
-			if (diffFunctions.length === 0) {
-				diffFunctions.push(auto);
+			const newProperty = this._bindFunctionProperty(properties[propertyName], bind);
+			if (registeredDiffPropertyNames.indexOf(propertyName) !== -1) {
```
kitsonk: I don't know if it is any more performant, but you could use `@dojo/shim/array.includes()` here which offloads to native...

src/WidgetBase.ts
```diff
@@ -328,41 +292,57 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+				runReactions = true;
+				const diffFunctions = this.getDecorator(`diffProperty:${propertyName}`);
+				for (let i = 0; i < diffFunctions.length; i++) {
+					const result = diffFunctions[i](previousProperty, newProperty);
+					if (result.changed && changedPropertyKeys.indexOf(propertyName) === -1) {
```
kitsonk: `@dojo/shim/array.includes()`?

src/WidgetBase.ts
```diff
@@ -328,41 +292,57 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+				if (args.changed) {
+					reaction.call(this, args.previousProperties, args.newProperties);
+				}
+			});
+		}
 
 		this._properties = diffPropertyResults;
 
-		if (changedPropertyKeys.length) {
+		if (changedPropertyKeys.length > 0) {
```
kitsonk: 💯

src/WidgetBase.ts
```diff
@@ -328,41 +292,57 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-
-		this._mapDiffPropertyReactions(properties, changedPropertyKeys).forEach((args, reaction) => {
-			if (args.changed) {
-				reaction.call(this, args.previousProperties, args.newProperties);
 			}
-		});
+		}
+
+		if (runReactions) {
+			this._mapDiffPropertyReactions(properties, changedPropertyKeys).forEach((args, reaction) => {
```
kitsonk: speculating, but with us targeting the down emit and access `this` in the arrow function, it might be an opportunity to create a bound callback that we cache if this gets called a lot?

src/WidgetBase.ts
```diff
@@ -646,11 +630,23 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-		.map((child: DNode) => {
-			return this._dNodeToVNode(child);
-		});
+		if (dNode.properties.key !== undefined) {
+			dNode.properties.afterCreate = this._afterCreateCallback;
+			dNode.properties.afterUpdate = this._afterUpdateCallback;
+		}
+
+		if (dNode.properties.bind === undefined) {
+			(<any> dNode).properties.bind = this;
```
kitsonk: why if `dNode.properties.bind` didn't require a cast on `dNode`, it does now?

agubler: Because it is readonly...

src/WidgetBase.ts
agubler: I "think" that `indexOf` performs better across our supported platforms, although native `includes` is slightly better on chrome and firefox according to this benchmark: https://jsperf.com/array-indexof-vs-includes

src/WidgetBase.ts
agubler: It's not exactly a hot path as only gets called if there is a registered `diffProperty`.

src/WidgetBase.ts
kitsonk: 👍

src/WidgetBase.ts
kitsonk: 👍

agubler:
@kitsonk any more feedback / comments on this one?

src/WidgetBase.ts
```diff
@@ -198,13 +191,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this._properties = <P> {};
 		this._cachedChildrenMap = new Map<string | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>();
 		this._diffPropertyFunctionMap = new Map<string, string>();
-		this._renderDecorators = new Set<string>();
 		this._bindFunctionPropertyMap = new WeakMap<(...args: any[]) => any, { boundFunc: (...args: any[]) => any, scope: any }>();
 		this._registries = new RegistryHandler();
 		this._registries.add(registry);
 		this.own(this._registries);
+		this._boundRenderFunc = this.render.bind(this);
```
kitsonk: More of general observation of why we don't use initializers generally... like:

```ts
class WidgetBase {
    private _properties = {} as P;
    private _registries = new RegistryHandler();
}
```

It would cut out a fair amount of typing and is even more readable IMO.

src/WidgetBase.ts
kitsonk: 👍

src/WidgetBase.ts
```diff
@@ -198,13 +191,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this._properties = <P> {};
```
kitsonk: Would `Object.create(null)` perform better?  I don't if this gets overwritten or not, but if it doesn't, I suspect it will be a minor performance improvement in property lookups.

agubler: I think generally it does, I don't think it'll reflect in the benchmarks but I'll update it regardless.

src/WidgetBase.ts
agubler: I prefer initialising inline also, seemed unrelated to these changes though.

kitsonk: True... 🤷‍♂️

src/WidgetBase.ts
agubler: Actually `hasOwnProperty` is used by Maquette on properties so we need it to be an object.

kitsonk: DOH! Sorry, I think I remember that now.  😢 Is it worth it to PR maquette to use `Object.hasOwnProperty.call(props, key)` so that `Object.create(null)` can be used, or is it just a minimal improvement thing that it wouldn't matter?

agubler: The honest answer is I don't know how significant it would be, I do suspect minimal.

kitsonk: Ok, other fruit to squeeze... 🍋

Issue #571, opened 2017-06-12
Scrolling, Gesture and Dragging abstraction via meta


kitsonk:
**Enhancement** 

Implementing gestures and scrolling present several challenges:

* They require intimate knowledge of the DOM and DOM events.
* They are somewhat inconsistent across browsers.
* To truly model the user action, it requires the widget to break encapsulation of the DOM to listen for events outside of its own DOM structure.
* Currently require duplication of code across widgets needing to support these features.

Therefore we should consider abstracting these features via the meta API.  Conceptually an interaction would look something like this:

* Widget indicates that it wasn't to know about scrolling for itself or one of its keys.
* The meta manager ensure that it is listening for the proper events if it isn't already.
* When the scroll event occurs that applies to the widget, it invalidates the widget.
* When the widget renders, it requests the information about the scroll position.

Additional considerations:

- Integration and further abstraction of the Intersection Observers meta?
- Implement a full drag and drop solution, including target acceptance and other features over time?
- For scrolling, it would be a good idea to deal in abstract concepts and likely would need to integrate into the dimensions meta as well.  For example, it would be good if the widget could offload the calculation of the changes from a relative value to a specific value.  Specifically, given a horizontal scroll event, the widget would have offered up some relative scale to meta on subscription (e.g. `100`) and the change would be relative. (e.g. `1` or `-1`) and to the meta dimensions it could say "tell me what my absolute scroll top would be if the change was a relative `-1` out of `100`).  That way, the widget would have no need to be concerned about its absolute values, it simply would trust whatever is returned is accurate.

Prior art/references:

- [HammerJS](https://hammerjs.github.io/)
- [Dojo 1 DND](https://github.com/dojo/dojo/tree/master/dnd)
- [pep.js](https://github.com/jquery/PEP) - Already included with widget core
- [touch-action](https://www.w3.org/TR/pointerevents/#the-touch-action-css-property) (can effect when the action is triggered, reducing delay)

Refs: #507

kitsonk:
Drag Meta does not fully resolve issue.

kitsonk:
We have delivered part of this, the rest should be delivered as we re-factor/address widgets where such functionality is required.

agubler:
Drag meta has been available for sometime now and there no concrete requirements for additional work. Closing in favour of specific issues when use-cases are identified.

PR #620, opened 2017-07-25
Add MetaWithOptions and requireNode callback

pottedmeat:
**Type:** feature

**Description:**

* Adds an optional callback to `requireNode`
* Adds Meta/option bundling

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -16,12 +16,17 @@ import {
 	RegistryLabel,
 	Render,
 	VirtualDomNode,
-	WidgetMetaConstructor,
 	WidgetBaseConstructor,
 	WidgetBaseInterface,
 	WidgetProperties
 } from './interfaces';
 import MetaBase from './meta/Base';
+import {
+	WidgetMetaConstructor,
+	WidgetMetaOptions,
+	WidgetMetaRequiredNodeCallback
+} from './meta/interfaces';
+import { MetaWithOptionsRef } from './meta/MetaWithOptions';
 import RegistryHandler from './RegistryHandler';
 import { isWidgetBaseConstructor, WIDGET_BASE_TYPE } from './WidgetRegistry';
 
@@ -179,11 +184,11 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 
 	private _renderState: WidgetRenderState = WidgetRenderState.IDLE;
 
-	private _metaMap = new WeakMap<WidgetMetaConstructor<any>, MetaBase>();
+	private _metaMap = new Map<WidgetMetaConstructor<any> | MetaWithOptionsRef<any, any>, MetaBase<any>>();
 
 	private _nodeMap = new Map<string, HTMLElement>();
 
-	private _requiredNodes = new Set<string>();
+	private _requiredNodes = new Map<string, WidgetMetaRequiredNodeCallback[]>();
 
 	/**
 	 * @constructor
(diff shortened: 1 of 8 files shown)


agubler:
@pottedmeat Would like to see an issue for bug/feature enhancement changes if possible. It would help us understand the goal of the changes.

src/WidgetBase.ts
```diff
@@ -269,6 +284,25 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		return node;
 	}
 
+	@afterRender()
+	protected notifyMetaKeys(node: DNode | DNode[]): DNode | DNode[] {
```
agubler: Would like to try and avoid using `afterRenders` in `WidgetBase` after analysing the results from `js-framework-benchmark`

src/meta/Base.ts
```diff
@@ -1,38 +1,61 @@
...
-import Set from '@dojo/shim/Set';
-import { WidgetMetaProperties } from '../interfaces';
+import {
+	WidgetMetaOptions,
+	WidgetMetaProperties,
+	WidgetMetaRequiredNodeCallback
+} from './interfaces';
 
-export class Base {
+export abstract class Base<T, O extends WidgetMetaOptions = WidgetMetaOptions> {
```
agubler: I don't understand the first generic?

src/meta/MetaWithOptions.ts
```diff
@@ -0,0 +1,18 @@
+import MetaBase from './Base';
+import { WidgetMetaConstructor, WidgetMetaOptions } from './interfaces';
+
+function MetaWithOptions<T extends MetaBase<any>, O extends WidgetMetaOptions>(MetaType: WidgetMetaConstructor<T, O>, options: O): MetaWithOptionsRef<T, O> {
+	return new MetaWithOptionsRef(MetaType, options);
+}
+
+export class MetaWithOptionsRef<T extends MetaBase<any>, O extends WidgetMetaOptions> {
+	private MetaType: WidgetMetaConstructor<T, O>;
```
agubler: underscores for privates

src/WidgetBase.ts
```diff
@@ -206,15 +211,25 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
 
-	protected meta<T extends MetaBase>(MetaType: WidgetMetaConstructor<T>): T {
-		let cached = this._metaMap.get(MetaType);
+	protected meta<T extends MetaBase<any>>(MetaOrMetaWithOptions: WidgetMetaConstructor<T> | MetaWithOptionsRef<T, any>): T {
+		let cached = this._metaMap.get(MetaOrMetaWithOptions);
 		if (!cached) {
+			let options: WidgetMetaOptions;
+			let MetaType: WidgetMetaConstructor<T>;
+			if (MetaOrMetaWithOptions instanceof MetaWithOptionsRef) {
+				options = (<any> MetaOrMetaWithOptions).options;
```
agubler: this feels like a hack

kitsonk:
Also, as a general comment, I don't know if it calls for a slightly different approach, because in trying to create a meta manager for gestures (#571) I am realising it might just be easier to be specific in the `get()` API to pass options instead of the meta registration.

For example, I am thinking it will be something like this...

```ts
class FooWidget extends WidgetBase {
  render() {
    const scrollResult = this.meta(Gestures).get('root', { type: 'scroll' });
    const dragResult = this.meta(Gestures).get('draghandle', { type: 'drag' });
  }
}
```

This means the meta `Base` would stay the same without having to worry about abstract handling of options which isn't meaningful for every manager and makes it easier to provide different types of results which are totally under the control of manager instance, versus having potentially multiple managers per widget.

src/WidgetBase.ts
```diff
@@ -269,6 +284,25 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+			decorate(node, (node: any) => {
+				const { properties = {} }: { properties: { key?: string } } = node;
+				if (properties.key) {
+					keys.push(properties.key);
+				}
+			}, (node: DNode) => {
+				return isHNode(node) || isWNode(node);
+			});
+			this._metaMap.forEach((meta) => {
+				meta.onKeysUpdated(keys);
```
agubler: Doesn't feel like `onKeysUpdated` should be on the public meta API

PR #623, opened 2017-07-26
Ensure _decoratorCache uses class-level decorators

### LINKED ISSUE #622: addDecorator called at the method level does not use decoratorMap
**Bug** 

Calling `addDecorator` from within an instantiated widget - specifically with a decorator such as `afterRender` which takes place after `render` - will add the callback to `_decoratorCache` without adding class-level decorators from `decoratorMap`.

pottedmeat:
**Type:** bug

**Description:**

addDecorator called at the method level does not use decoratorMap to add class-level decorators to the cache.

Resolves #622


agubler:
Can we have a test for this change please?

pottedmeat:
Added

agubler:
@pottedmeat This has conflicts that need to be resolved :)

PR #624, opened 2017-07-27
Add Hover meta manager

kitsonk:
**Type:** feature

**Description:**

This PR adds a hover meta manager.  In part is raised to discuss some features of meta to help flesh out some of the architecture as well as provide the ability to detect a hover state in the DOM without having to store that state in the widget.

Usage would be something like:

```ts
class SomeWidget extends ThemeableMixin(WidgetBase) {
  render() {
    return v('div', {
      classes: this.classes(this.meta(Hovering).get('root') && css.hovering),
            key: 'root'
        }, [ 'Hello World!' ]);
  }
}
```

Things to note about the PR, which I think would extend to other DOM meta managers is:

- Currently there is no easy/built-in way of managing some state information related to the actual DOM nodes, the Hover manager has to create that, but I know there is other state information relating it back to a key that is cached in the base class.  We might want to provide some easy mechanism to cache/set state related to the nodes.
- There is no way of managing event handlers.  It is logical that meta managers will want/need to hook event handlers, either at the node level or at some delegate level.  They are easy to create, but it raises lifecycle concerns in my mind and potentially memory leaks.  The Base doesn't inherit from Destroyable so, there is no way to own a destruction handle.  I am not entirely sure there is a concern, but it feels dirty adding things and not cleaning up after myself.

Refs #571

### DIFF
--- a/src/meta/Hover.ts
+++ b/src/meta/Hover.ts
@@ -1,0 +1,53 @@
+import WeakMap from '@dojo/shim/WeakMap';
+import { Base } from './Base';
+import { WidgetMetaProperties } from '../interfaces';
+
+interface HoverInfo {
+	hovering: boolean;
+}
+
+export default class Hover extends Base {
+	private _hoverInfo: WeakMap<HTMLElement, HoverInfo>;
+
+	constructor(properties: WidgetMetaProperties) {
+		super(properties);
+		this._hoverInfo = new WeakMap();
+	}
+
+	/**
+	 * Returns `true` if the mouse is hovering over the node, identified by `key`, otherwise returns `false`
+	 * @param key The key of the node that the hover information is required for
+	 */
+	public get(key: string): boolean {
+		this.requireNode(key);
+
+		const node = this.nodes.get(key);
+
+		if (!node) {
+			return false;
+		}
+
+		let hoverInfo = this._hoverInfo.get(node);
+
+		if (!hoverInfo) {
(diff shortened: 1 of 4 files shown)


kitsonk:
I was thinking about this more...

It is nice to get references to the nodes, but there is a bit a boiler plate to deal with the node not being created yet for a manager.  I wonder if Base could extract it further, or delegate to the Widget Core doing stuff with a node... sort of imply that if you interact with it, it becomes required, and allow it to set things like event listeners without having to worry about it really being there.  Something like:

```
this.nodes('key').addEventListener('click', () => {});
```

Where it returns not the real node instance, persay, but a "proxy" with certain key APIs that are dealt with when the node is really created.

At the very least I think we need to find a way to delegate the adding of listeners.

kitsonk:
@agubler, @matt-gadd, and myself chatted... while this PR is _functional_ it is trivial and we agreed that we would want to _draw the line_ at this:

* When you do not have to introspect the `e.target` of an event listener in order to handle the listener (e.g. in the case of hover, `onmouseenter` and `onmouseexit` can be added and used without the need of looking at the event at all) then the _right_ pattern is to add these listeners and model any state in the widget instance.
* There is a bit of a fuzzy line when you are needing to look at other parts of the event structure, but we would want to take those on a case by case basis.

The rest of the questions raised in this PR though for discussion will be taken up by looking at POCs for other meta providers or other issues, so we are going to close this PR.

PR #628, opened 2017-08-01
Support default properties

### LINKED ISSUE #466: Add default generics where appropriate in widget-core

agubler:
**Type:** feature

**Description:**

Add support for specifying defaults for optional properties. Once defaulted, the properties will not be considered optional in usage within the widget. I was not able to restrict the default properties to the properties interface, which means "extra" properties can be added and would be available within the widget's use of `this.properties`.

Example usage:

```ts
interface TestProps {
	a: string;
	b?: string;
	c: string;
	d?: string;
}

function num(a: string) {}

class TestOne extends WidgetBase<TestProps> {

	defaultProperties = { b: '1' };

	myFuncOne() {
		num(this.properties.b); // uses the defaulted value and doesn't error for `strictNullChecks`
	}

	myFuncTwo() {
		num(this.properties.d); // no default, so fails the `strictNullChecks`
	}
}

class TestTwo extends WidgetBase<TestProps> {

	defaultProperties = { d: '1' };

	myFuncOne() {
		num(this.properties.b); // no default, so fails the `strictNullChecks`
	}

	myFuncTwo() {
		num(this.properties.d); // uses the defaulted value and doesn't error for `strictNullChecks`
	}
}
```

Resolves #466

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -187,6 +187,8 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 
 	private _requiredNodes = new Set<string>();
 
+	public readonly defaultProperties = {};
+
 	/**
 	 * @constructor
 	 */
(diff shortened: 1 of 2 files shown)


agubler:
Needs some tests

agubler:
Unfortunately this has a side affect of causing an error when trying to spread properties: `error TS2698: Spread types may only be created from object types.`

Need to look at whether this can be avoided.

agubler:
The typings causing issues on this one.

PR #636, opened 2017-08-17
Remove Global Registry

### LINKED ISSUE #613: ProjectorMixin.setProperties() type guard
**Bug**

The method `.setProperties()` on the `ProjectorMixin` doesn't appear to guard against whatever the base properties are.  Instead it show as typed to accept any properties.  I haven't dug into why this is the case.

agubler:
**Type:** feature

**Description:**

Remove global registry, registry mixin and support default registry in `WidgetBase`.

* If a widget is not passed a `registry` property a new registry will be created and passed to children widgets as `defaultRegistry`
* If a widget is passed a `registry` and doesn't receive a `defaultRegistry` property then it will be "promoted" to the `defaultRegistry` and automatically passed to children.
* If a widget is passed a `registry` and a `defaultRegistry` then only the `defaultRegistry` will be passed to children automatically.

Resolves #613


src/WidgetBase.ts
```diff
@@ -291,6 +293,33 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		return this._properties;
 	}
 
+	@diffProperty('registry', reference)
+	protected diffPropertyRegistry(previousProperties: any, newProperties: any): void {
+		if (this._registries.defaultRegistry === this._defaultRegistry) {
+			this._registries.remove(this._defaultRegistry);
+		}
+		const result = this._registries.replace(previousProperties.registry, newProperties.registry);
```
kitsonk: Small nit, but you repeatedly access `this._registries`.  `const { _registries } = this;` would allow a build optimizations to occur and would reduce the code size.

src/WidgetBase.ts
```diff
@@ -291,6 +293,33 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+		if (!result) {
+			this.getRegistries().add(newProperties.registry);
```
kitsonk: Why access the public method here, when in the rest of the code, you are not accessing that way, but instead directly to the private member?

src/WidgetBase.ts
```diff
@@ -291,6 +293,33 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+		}
+	}
+
+	@diffProperty('defaultRegistry', reference)
+	protected diffPropertyDefaultRegistry(previousProperties: any, newProperties: any): void {
+		if (newProperties.defaultRegistry === undefined) {
```
kitsonk: again `const { _registries } = this` would provide a build time optimization.

src/WidgetBase.ts
agubler: I can change it, but curious what build optimisations would currently occur?

src/WidgetBase.ts
agubler: you're right, should just access the private member...

src/WidgetBase.ts
kitsonk: Currently, out of the box uglify would leave a `this.whatever` alone, since it cannot be renamed, since it does not have a local variable to deal with.  Changing it to a local variable, when you have multiple access to it, will allow uglify to shorten the variable, so something like this:

```js
const whatever = this.whatever;
if (!this.whatever.has()) {
    this.whatever.add({});
}
```

can be re-written to:

```js
const a=this.whatever if(!a.has()){a.add({})}
```

Instead of:

```
if(!this.whatever.has()){this.whatever.add({})}
```

It really only matters when you have several access to the same member in a block.  There is also a small small performance gain of lookup, though most runtimes would optimise that anyways.

agubler: Thanks for the explanation.

PR #637, opened 2017-08-17
Drag Meta Provider

kitsonk:
**Type:** Feature

**Description:**

This is a proof of concept meta provider which provides drag information to a widget instance.  The provider provides a `DragResults` every time it is called.  Those drag results provide if the node is being currently dragged `isDragging` and any change in position from the last time the provider was called.

An example of how this would look in a real world implementation would be something like this:

```ts
export interface ScrollBarProperties extends WidgetProperties {
  horizontal: boolean;
  position: number;
  size?: number;
  sliderMin?: number;
  sliderSize?: number;
  visible?: boolean;
  onScroll?(delta: number): void;
}

class ScrollBar extends WidgetBase<ScrollBarProperties> {
  const { key, horizontal } = this.properties;
  let isDragging = false;
  if (onScroll) {
    const dragResult = this.meta(Drag).get('slider');
    const delta = horizontal ? dragResult.delta.x : dragResult.delta.y;
    isDragging = dragResult.isDragging;
    onScroll(delta);
  }

  // some calculated styles to position the slider
  // the UI can also render different if the slider is being dragged
  return v('div', { key }, [ v('div', { key: 'slider' }) ]);
}
```

Because the DragController tracks the events from the window level, but is able to attribute it back to a particular node, tracking of the events keeps the drag state accurate.  Here is an example of it in action:

dragging_example

Fixes: #571 (in part)


kitsonk:
We should support pointer events as part of this provider and it should not need to deal with touch/mouse events then.

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
+import global from '@dojo/shim/global';
+import { assign } from '@dojo/shim/object';
+import WeakMap from '@dojo/shim/WeakMap';
+import { Base } from './Base';
+
+export interface DragResults {
+	/**
+	 * The movement of pointer during the duration of the drag state
+	 */
+	delta: Position;
```
agubler: nit: should we declare the interface before we use it? (I thought it was an from an import)

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+function getPosition(e: PointerEvent): Position {
+	return {
+		x: e.pageX,
+		y: e.pageY
+	};
+}
+
+/**
+ * Return the delta position between two positions
+ * @param start The first posistion
```
agubler: posistion  -> position

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+	};
+}
+
+class DragController {
+	private _nodeMap = new WeakMap<HTMLElement, NodeData>();
+	private _dragging: HTMLElement | undefined = undefined;
+
+	private _getData(target: HTMLElement): { state: NodeData, target: HTMLElement } | undefined {
+		if (this._nodeMap.has(target)) {
+			return { state: this._nodeMap.get(target)!, target };
```
agubler: it would be so good if TS could know that we are in a block narrowed by `this._nodeMap.has(target)`

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+		}
+		if (target.parentElement) {
+			return this._getData(target.parentElement);
+		}
+	}
+
+	private _onDragStart = (e: PointerEvent) => {
```
agubler: any reason why `private _onDragStart = (e: PointerEvent) => {` over `private _onDragStart(e: PointerEvent) {`?

agubler: nit: could we have `event` over `e`?

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+		state.dragResults = {
+			delta: getDelta(state.start, state.last),
+			isDragging: false
+		};
+		state.invalidate();
+		this._dragging = undefined;
+	}
+
+	constructor() {
+		const win: Window = global.window;
```
agubler: `window`?

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+		win.addEventListener('pointermove', this._onDrag, true);
+		win.addEventListener('pointerup', this._onDragStop, true);
+	}
+
+	public get(node: HTMLElement, invalidate: () => void): DragResults {
+		const { _nodeMap } = this;
+		// first time we see a node, we will initialize its state
+		if (!_nodeMap.has(node)) {
+			_nodeMap.set(node, {
+				dragResults: {
```
agubler: can you use `emptyResults`?

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+			state.dragResults.delta = { x: 0, y: 0 };
+		}
+
+		return dragResults;
+	}
+}
+
+const controller = new DragController();
+
+export default class Drag extends Base {
```
agubler: generally we do a named export and a default export across widget-core.

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+	private boundInvalidate = this.invalidate.bind(this);
```
agubler: `_boundInvalidate`

src/meta/Drag.ts
```diff
@@ -0,0 +1,170 @@
...
+ * A frozen empty result object, frozen to ensure that no one downstream modifies it
+ */
+const emptyResults = Object.freeze({
+	delta: Object.freeze({ x: 0, y: 0 }),
+	isDragging: false
+});
+
+/**
+ * Return the x/y position for an event
+ * @param e The MouseEvent or TouchEvent
```
agubler: now the `PointerEvent`?

src/meta/Drag.ts
agubler: apart from it being frozen... could we just deep assign the `emptyResults` when we return them?

src/meta/Drag.ts
kitsonk: Why the lambda, is because the handler needs to be bound to the instance.  When used as a handler, the `this` scope is the `window` I believe.

src/meta/Drag.ts
kitsonk: Don't want to mask the global.  We have tended to now do that, using `doc` or `win` when using a reference to a potential global.

src/meta/Drag.ts
kitsonk: well, empty results is valid here, I will use that

src/meta/Drag.ts
kitsonk: no problem

src/meta/Drag.ts
kitsonk: 👍

src/meta/Drag.ts
kitsonk: I was favouring alphabetisation over use before declaration... one form of clarity causes another issue

src/meta/Drag.ts
kitsonk: See: Microsoft/TypeScript#11117

src/meta/Drag.ts
agubler: I guess so.

src/meta/Drag.ts
agubler: Not surprised there’s an issue. That would be nice.

src/meta/Drag.ts
agubler: Gotcha re scope. Over having a separate variable that binds scope to the method (we have used both patterns)

kitsonk: Yeah... I like this better for class _methods_ where it effects the code readability less, IMO.  Using the bound variable makes a bit of indirection when reading the code.

agubler: Agreed, for private methods.

src/meta/Drag.ts
```diff
@@ -0,0 +1,182 @@
...
+function getDelta(start: Position, current: Position): Position {
+	return {
+		x: current.x - start.x,
+		y: current.y - start.y
+	};
+}
+
+class DragController {
+	private _nodeMap = new WeakMap<HTMLElement, NodeData>();
+	private _dragging: HTMLElement | undefined = undefined;
```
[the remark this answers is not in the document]
kitsonk: It could be, but considering I was explicitly assigning `undefined`, the property will always be on the instance.  Felt more accurate to not be optional.

src/meta/Drag.ts
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		return dragResults;
+	}
+}
+
+const controller = new DragController();
+
+export class Drag extends Base {
+	private _boundInvalidate: () => void = this.invalidate.bind(this);
```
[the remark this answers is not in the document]
kitsonk: It is its own `.invalidate` which isn't auto bound I believe... because this doesn't come from `WidgetBase`.  I will double check though.

tests/unit/meta/Drag.ts
```diff
@@ -0,0 +1,437 @@
...
+				delta: { x: -5, y: -10 },
+				isDragging: false
+			}
+		], 'dragging should be attributed to parent node');
+
+		widget.destroy();
+		document.body.removeChild(div);
+	},
+
+	'non draggable node'() {
```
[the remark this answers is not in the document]
kitsonk: Rename it to?

tests/unit/meta/Drag.ts
```diff
@@ -0,0 +1,437 @@
...
+						width: '100px',
+						height: '100px'
+					}
+				}, [
+					v('div', {
+						innerHTML: 'Hello World',
+						key: 'child1'
+					}),
+					v('div', {
+						innerHTML: 'Hello Wolrd',
```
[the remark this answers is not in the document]
kitsonk: 👍

tests/unit/meta/Drag.ts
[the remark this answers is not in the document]
pottedmeat: Non listening node?

src/meta/Drag.ts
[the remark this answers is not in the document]
kitsonk: While the properties passed in at creation provides a bound invalidate, that is a private property on base and the public `invalidate()` calls that private property that is copied out, therefore when called from event handlers, the `Base.invalidate()` loses its context and needs to be statically bound.  This is likely undesirable, and we should consider ensuring that `Base.invalidate()` will always call the right invalidate, but for now, this needs to stay this way to work.  I will open an issue to consider the change to `Base`.

PR #640, opened 2017-08-22
Update requireNode to accept a callback

### LINKED ISSUE #607: Meta is too aggressive when marking nodes required

### LINKED ISSUE #639: Add callback support to requireNode to bypass automatic invalidation
**Enhancement** 

In cases where simple access to the DOM allows a Meta Type to return a valid value - as is the case with Dimensions Meta - an immediate re-rendering when a node is absent makes sense as it will return a meaningful value in the next render cycle.

But in cases where the value is derived asynchronously - as is the case with Intersection Meta - the immediate re-render has no effect. With these use cases, the Meta Type has its own invalidation and simply needs to have access to the node to set up an event listener. We can do this through a callback.

pottedmeat:
**Type:** bug / feature

**Description:**

Removes aggressive `.requireNode` calls where they aren't needed.

Adds callback support to `.requireNode` and adds tests to ensure it works with asynchronous invalidation.

Resolves #607 
Resolves #639

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -295,13 +295,13 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 
 	private _setNode(element: Element, properties: VNodeProperties): void {
 		const key = String(properties.key);
+		this._nodeMap.set(key, <HTMLElement> element);
 		const callbacks = this._requiredNodes.get(key);
 		if (callbacks) {
 			for (const callback of callbacks) {
 				callback.call(this, element);
 			}
 		}
-		this._nodeMap.set(key, <HTMLElement> element);
 	}
 
 	public get properties(): Readonly<P> & Readonly<WidgetProperties> {
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -2,7 +2,6 @@ import { Evented } from '@dojo/core/Evented';
 import { ProjectionOptions, VNodeProperties } from '@dojo/interfaces/vdom';
 import Map from '@dojo/shim/Map';
 import '@dojo/shim/Promise'; // Imported for side-effects
-import Set from '@dojo/shim/Set';
 import WeakMap from '@dojo/shim/WeakMap';
 import { isWNode, v, isHNode } from './d';
 import { auto, reference } from './diff';
@@ -18,7 +17,8 @@ import {
 	WidgetMetaConstructor,
 	WidgetBaseConstructor,
 	WidgetBaseInterface,
-	WidgetProperties
+	WidgetProperties,
+	WidgetMetaRequiredNodeCallback
 } from './interfaces';
 import MetaBase from './meta/Base';
 import RegistryHandler from './RegistryHandler';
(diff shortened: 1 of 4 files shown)


src/meta/Base.ts
```diff
@@ -26,11 +24,18 @@ export class Base {
...
+		const node = this.nodes.get(key);
+		if (node) {
+			callback && callback.call(this, node);
+		}
+		else {
+			const callbacks = this._requiredNodes.get(key) || [];
+			callback && callbacks.push(callback);
+			this._requiredNodes.set(key, callbacks);
+			if (!callback) {
+				this.invalidate();
```
rorticus: Why do we no longer invalidate when a node isn't found? Without the invalidate, won't we have to wait for the next render?

pottedmeat: It should be laid out in the referenced issue #639. If `requireNode` is called without a callback, `invalidate` still fires but if it is called with a callback, it is up to that callback to call `invalidate`.

rorticus: i guess i'm confused as to when we don't want it to be invalidated. if we call requireNode, regardless of what the callback does, things aren't going to work without that node, right? Would you mind posting a use case of this with the callback that doesn't necessarily call invalidate? That might help me understand! Thanks!

pottedmeat: In `_setNode`, if there is a callback in `_requiredNodes`, it sends that node to the callback. I'm updating the Intersections meta now, but basically I call `observer.observe(node)` in the callback, then call `invalidate` when the observer's intersection event handler is called.

rorticus: Oh! I get it! I just wrote up this big complicated example of why I thought we needed that invalidate when I realized that we don't 😝

src/WidgetBase.ts
```diff
@@ -284,7 +285,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	}
 
 	private _setNode(element: Element, properties: VNodeProperties): void {
-		this._nodeMap.set(String(properties.key), <HTMLElement> element);
+		const key = String(properties.key);
```
kitsonk: Being reported as uncovered in tests.

pottedmeat: I'm seeing a "50x" next to that line

src/WidgetBase.ts
```diff
@@ -284,7 +285,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+		this._nodeMap.set(key, <HTMLElement> element);
+		const callbacks = this._requiredNodes.get(key);
+		if (callbacks) {
+			for (const callback of callbacks) {
+				callback.call(this, element);
```
matt-gadd: does this not now then bleed the scope of the widget into a meta callback?

pottedmeat: Yes. Will fix. Don't need to pass element either.

matt-gadd: 👍

src/WidgetBase.ts
kitsonk: It is because the CI isn't passing and therefore the code coverage isn't being uploaded.

PR #641, opened 2017-08-22
Clean up meta instances

### LINKED ISSUE #635: Meta Base should extend from Destroyable
**Enhancement**

The meta Base class should extend from `@dojo/core/Destroyable`.  This will allow a lifecycle for meta instances to allow them to clean up after themselves (like removing event listeners).

We also need to ensure that when `widget-core` instantiates meta instances, it also destroys them as well.

pottedmeat:
**Type:** feature

meta/Base extends Destroyable. Meta instances are set to owned by the widget instance.

Resolves #635

### DIFF
--- a/src/meta/Base.ts
+++ b/src/meta/Base.ts
@@ -1,15 +1,18 @@
+import { Destroyable } from '@dojo/core/Destroyable';
 import global from '@dojo/shim/global';
 import Map from '@dojo/shim/Map';
 import Set from '@dojo/shim/Set';
 import { WidgetMetaProperties } from '../interfaces';
 
-export class Base {
+export class Base extends Destroyable {
 	private _invalidate: () => void;
 	private _invalidating: number;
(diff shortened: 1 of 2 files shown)


src/meta/Base.ts
```diff
@@ -1,15 +1,18 @@
...
 import Map from '@dojo/shim/Map';
 import Set from '@dojo/shim/Set';
 import { WidgetMetaProperties } from '../interfaces';
 
-export class Base {
+export class Base extends Destroyable {
 	private _invalidate: () => void;
 	private _invalidating: number;
 	private _requiredNodes: Set<string>;
 	protected nodes: Map<string, HTMLElement>;
```
kitsonk: Both the `_requiredNodes` and `nodes` should be cleared on destruction so they do not hold their references and can be freed for GC.

kitsonk:
Still would like to see a review by @agubler though before merging.

PR #653, opened 2017-09-01
Adds support for a decorator that defines widgets in the default widget registry

### LINKED ISSUE #652: Decorator to define registry items in the default registry
**Enhancement**

Moving away from a single global registry has reduced the developer ergonomics for defining registry items as they need to all be done at the point that the of creating the registry.

```ts
// MyWidget.ts
    render() {
        return v('div', [
            w('widget-one', {}),
            w('widget-two', {}),
            w('widget-three', {})
        ]);
    }
}

// main.ts
import MyWidget from './MyWidget';
import WidgetOne  from './widgets/one';
import WidgetTwo  from './widgets/two';
import WidgetThree  from './widgets/three';

registry.define('widget-one', WidgetOne);
registry.define('widget-two', WidgetTwo);
registry.define('widget-three', WidgetThree);

const Projector = Projector(MyWidget);
projector.append();
```

This _does_ work but is cumbersome, ideally each of the widgets could be defined against the default registry along side the widget that needs them.

```ts
// MyWidget.ts
import WidgetOne  from './widgets/one';
import WidgetTwo  from './widgets/two';
import WidgetThree  from './widgets/three';

@registryItem('widget-one', WidgetOne)
@registryItem('widget-two', WidgetTwo)
@registryItem('widget-three', WidgetThree)
    render() {
        return v('div', [
            w('widget-one', {}),
            w('widget-two', {}),
            w('widget-three', {})
        ]);
    }
}

// main.ts
import MyWidget from './MyWidget';

const Projector = Projector(MyWidget);
projector.append();
```

agubler:
**Type:** feature

**Description:**

Adds support for a decorator `@define` that accepts a registry label and widget to register in the default widget registry if the widget.

Resolves #652

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -355,7 +355,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		const { defaultRegistry } = this._registries;
 		for (let i = 0; i < registryItems.length; i++) {
 			const { label, item } = registryItems[i];
-			if (defaultRegistry && defaultRegistry.has(label) === false) {
+			if (defaultRegistry && !defaultRegistry.has(label)) {
 				defaultRegistry.define(label, item);
 			}
 		}
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -23,7 +23,7 @@ import {
 	WidgetMetaRequiredNodeCallback
 } from './interfaces';
 import RegistryHandler from './RegistryHandler';
-import { isWidgetBaseConstructor, WIDGET_BASE_TYPE, WidgetRegistry } from './WidgetRegistry';
+import { isWidgetBaseConstructor, WIDGET_BASE_TYPE, WidgetRegistry, WidgetRegistryItem } from './WidgetRegistry';
 
 /**
  * Widget cache wrapper for instance management
@@ -50,6 +50,11 @@ interface ReactionFunctionArguments {
 interface ReactionFunctionConfig {
 	propertyName: string;
 	reaction: DiffPropertyReaction;
+}
+
+interface RegisterConfig {
+	label: RegistryLabel;
+	item: WidgetRegistryItem;
 }
 
 export type BoundFunctionData = { boundFunc: (...args: any[]) => any, scope: any };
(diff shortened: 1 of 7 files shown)


src/WidgetBase.ts
```diff
@@ -320,6 +323,19 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
 		}
+		this._defineRegistryItems();
+	}
+
+	private _defineRegistryItems(): void {
+		const registryItems: DefineDecoratorConfig[] = this.getDecorator('registryItem');
+		const { defaultRegistry } = this._registries;
+		for (let i = 0; i < registryItems.length; i++) {
+			const { label, item } = registryItems[i];
+			if (defaultRegistry && defaultRegistry.has(label) === false) {
```
kitsonk: Any particular reason for `defaultRegistry.has(label) === false` instead of `!defaultRegistry.has(label)`?  Not that I have any particular concern, but `has` returns a boolean, so it is unnecessary to guard against falsey values and there is a very minor performance difference:

strict_comparison_ _jsperf

src/decorators/registry.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Decorator function to registry items that will be defined against
+ * the default registry of a widget.
+ *
+ * @param label The label of the registry item
+ * @param item The registry item to define
+ */
+export function define(label: RegistryLabel, item: WidgetRegistryItem) {
```
kitsonk: again, more discussion, but `.add` would align it better to other _collections_ from an API consistency perspective... of course a more semantic name could be `register` as well... No strong feelings, but just poking at things to see if they stand up.

src/WidgetBase.ts
agubler: Just a habit, I can change. (just curious is that true across all the browsers?)

src/decorators/registry.ts
agubler: The registry's API is `define` (as is custom elements) so it is aligned to that.

kitsonk: Ah, you trumped me on API alignment... 🙇

src/WidgetBase.ts
kitsonk: https://jsperf.com/strict-comparison-kpk/1

You gotta love JVM engines though
dashboard
:

strict_comparison_ _jsperf
strict_comparison_ _jsperf
dashboard

agubler: Surprised it's that much slower in IE11!

PR #659, opened 2017-09-06
Add support for core properties

### LINKED ISSUE #658: Core Properties for WidgetBase
**Enhancement**

`WidgetBase` currently piggybacks core properties onto the standard `WNode.properties` which means they are available on the widgets `this.properties` (even if the typings hides them at design time). It would be useful to store the core properties as a different property on `WNode` so they do not pollute the users `properties` and that `WidgetBase` can process them outside of the `__setProperties__`.

This proposal is to create a concept of `coreProperties` which currently includes `bind`, `defaultRegistry` and `registry` that will be processed before `__setProperties__`, by `__setCoreProperties__`.

This will help support a `beforeProperties` lifecycle hook and eventually a decorator that supports "inline" injection (as the `registries`/`bind` will be processed before the `beforeProperties` lifecycle is executed).

agubler:
**Type:** feature

**Description:**

Adds support for `coreProperties` and processes them before the normally `__setProperties__` phase.

Resolves #658

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -464,7 +464,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 					}
 				}
 				else {
-					const { defaultRegistry, registry } = this.getCoreProperties(node.properties);
+					const { defaultRegistry, registry } = this.__getCoreProperties__(node.properties);
 					node.coreProperties = node.coreProperties || {};
 
 					if (node.coreProperties.bind === undefined) {
@@ -478,7 +478,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		}
 	}
 
-	protected getCoreProperties(properties: any): CoreProperties {
+	protected __getCoreProperties__(properties: any): CoreProperties {
 		const {
 			registry
 		} = properties;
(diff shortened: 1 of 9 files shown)


src/WidgetBase.ts
```diff
@@ -429,26 +452,46 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-				(<any> node.properties).bind = this;
-			}
-			if (isWNode(node)) {
-				(<any> node.properties).defaultRegistry = this._registries.defaultRegistry;
-			}
-			nodes = [ ...nodes, ...node.children ];
 		}
 	}
 
+	protected getCoreProperties(properties: any): CoreProperties {
```
matt-gadd: do you think this should be double underscored?

src/WidgetBase.ts
```diff
@@ -429,26 +452,46 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+					}
 				}
+				else {
+					const { defaultRegistry, registry } = this.getCoreProperties(node.properties);
+					node.coreProperties = node.coreProperties || {};
+
+					if (node.coreProperties.bind === undefined) {
+						node.coreProperties.bind = this;
+					}
+					node.coreProperties.defaultRegistry = defaultRegistry;
```
matt-gadd: do you think we should make `coreProperties` a symbol? or bad idea?

src/WidgetBase.ts
agubler: That's a reasonable suggestion as we do not want people to override this function, even if it is just `protected`

src/WidgetBase.ts
agubler: I'm not sure what it would give?

matt-gadd: i guess it would make it transparent to the end user. but i think we agreed that didn't matter anyway.

PR #664, opened 2017-09-07
Intersection Meta

### LINKED ISSUE #644: Intersection Meta
**Enhancement** 

Intersection observers work by registering an observer instance with a callback to either a specific node or the overall viewport of the page. This observer instance may have additional options where a list of thresholds, the root node, and the root margin are all items we should support.

These options should be passed to the Intersection meta methods (`get`, `has`, etc.) where each method will ensure an observer is created, if necessary, using those options or reused if one already exists.

`get` should return an object containing `intersectionRatio` (a number) and `isIntersecting` (a boolean). This is due to `intersectionRatio` returning `0` even if the node is intersecting.

One of the pieces of Intersection that is different is that it will frequently be used to check to see whether child nodes are inside or outside of a root view and does not need to invalidate otherwise. To support this, we should add an API:

`invalidateIf(key, within|outside|custom, callback?, options)`

For each key/options pair, invalidation can be set to only occur when that node is inside the root view, outside the view view, or when a custom function - provided the previous and updated values - returns `true`.

pottedmeat:
**Type:** feature

**Description:**

Allows for observing intersections with a root node, root margin, and intersection thresholds (passed as a second argument to `get` and `has`). Returns an object with `isIntersecting` and `intersectionRatio`. 

Contains code to optionally change which intersection cause an invalidation (Within, Outside, Never, and Custom) by calling `invalidateIf`.

Resolves #644


src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+import WeakMap from '@dojo/shim/WeakMap';
+import { Base } from './Base';
+
+interface IntersectionDetail {
+	entries: WeakMap<Element, IntersectionObserverEntry>;
+	intersectionObserver?: IntersectionObserver; // attached observer
+	options: IntersectionGetOptions;
+	root: string;
+	rootMargin: string | undefined;
+	thresholds: number[]; // thresholds the observe should be attached with
```
agubler: nit: can you add this as jsdoc and not an inline comment?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export interface IntersectionGetOptions {
+	root?: string;
+	rootMargin?: string;
+	threshold?: number;
```
agubler: Do we need `threshold` and `thresholds`? If we need a convenience of passing a single threshold, can we not just type it `number | number[]`?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+const defaultIntersection: IntersectionResult = Object.freeze({
+	intersectionRatio: 0,
+	isIntersecting: false
+});
+
+export class Intersection extends Base {
+	private _details: IntersectionDetail[] = [];
+
+	private _getDetails(options: IntersectionGetOptions): IntersectionDetail {
+		let {
```
agubler: Are these all `lets`? Does `root` need a default?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+			root = '',
+			rootMargin,
+			threshold,
+			thresholds = []
+		} = options;
+		const {
+			_details: details
+		} = this;
+		if (typeof threshold === 'number' && !thresholds.length) {
```
agubler: We wouldn't need this `thresholds`, is typed as `number | number[]`. 

Could just do:

```ts
thresholds = Array.isArray(thresholds) ? thresholds : [ thresholds ];
```

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+		// Look to see if a detail record has already been created for these options
+		let cached: IntersectionDetail | undefined = undefined;
+		for (const detail of details) {
+			if (
+				options === detail.options ||
+				(
+					root === detail.root &&
+					rootMargin === detail.rootMargin &&
+					thresholds.length === detail.thresholds.length &&
+					thresholds.every(function(i) {
```
agubler: Am I reading this wrong? is this using the threshold as an index?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+			_details: details
+		} = this;
+		if (typeof threshold === 'number' && !thresholds.length) {
+			thresholds = [ threshold ];
+		}
+
+		// Look to see if a detail record has already been created for these options
+		let cached: IntersectionDetail | undefined = undefined;
+		for (const detail of details) {
+			if (
```
agubler: I think we can simplify this if condition some what....

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+			intersectionOptions.threshold = thresholds;
+		}
+		if (rootNode) {
+			intersectionOptions.root = rootNode;
+		}
+		const observer = new global.IntersectionObserver(this._onIntersect.bind(this, details), intersectionOptions);
+		details.intersectionObserver = observer;
+		this.own({
+			destroy() {
+				observer.disconnect();
```
agubler: Shows as uncovered in the test coverage

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
...
+
+		const node = this.nodes.get(key);
+		if (details && node) {
+			const entry = details.entries.get(node);
+			/* istanbul ignore else: only process entry if it exists */
+			if (entry) {
+				const {
+					intersectionRatio,
+					isIntersecting
+				} = <(IntersectionObserverEntry & { isIntersecting: boolean })> entry;
```
agubler: Do we have to cast like this?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,167 @@
+import global from '@dojo/shim/global';
+import WeakMap from '@dojo/shim/WeakMap';
+import { Base } from './Base';
+
+interface IntersectionDetail {
+	entries: WeakMap<Element, IntersectionObserverEntry>;
+	intersectionObserver?: IntersectionObserver; // attached observer
+	options: IntersectionGetOptions;
+	root: string;
+	rootMargin: string | undefined;
```
agubler: `rootMargin?: string;` over `rootMargin: string | undefined`?

src/meta/Intersection.ts
pottedmeat: My preference would be to go with thresholds and make it always `number[]`

src/meta/Intersection.ts
pottedmeat: I did this because I can't assign undefined to an optional string when declaring an object of that type.

src/meta/Intersection.ts
pottedmeat: Sure. I didn't because it's a private interface.

src/meta/Intersection.ts
pottedmeat: I'll split it up into a let and const block. root doesn't need a default, just made object/typing cleaner - same as the `rootMargin: string | undefined` above.

src/meta/Intersection.ts
pottedmeat: In terms of formatting or?

src/meta/Intersection.ts
pottedmeat: Until TypeScript is updated to include this property.

src/meta/Intersection.ts
pottedmeat: If thresholds is always an array, the `let` doesn't matter and I can allow `root` to be `undefined` in the details object.

src/meta/Intersection.ts
pottedmeat: Yeah, I'm not sure why it's not getting called.

src/meta/Intersection.ts
pottedmeat: I should have added a test for this to ensure the same set of options doesn't create additional observers.

src/meta/Intersection.ts
agubler: The logic is pretty complicated, i'd either break these logical conditions up so each one is done individually, like:

```ts
if (options !== detail.options) {
    continue;
}
```

Or perhaps even better, use a different mechanism for the options caching (like a map, using the serialised options as the key)? It would look something like this:

```ts
	private _details: Map<string, IntersectionDetail> = new Map();

	private _getDetails(options: IntersectionGetOptions): IntersectionDetail {
		const cacheKey = JSON.stringify(options);
		let cached = this._details.get(cacheKey);

		if (!cached) {
			const entries = new WeakMap<Element, IntersectionObserverEntry>();
			cached = { ...options, entries };
			this._details.set(cacheKey, cached);
		}
		return cached;
	}
```

pottedmeat:
Closed in lieu of Ant's PR

PR #666, opened 2017-09-08
Add evented Node handler to meta

### LINKED ISSUE #662: Add eventing system to meta to determine when nodes are processed
**Enhancement** 

In some circumstances, animations being one of them, we need to delay the requiring of a node until we can be sure that all node properties have been processed and all `HTMLElements` are correctly laid out. This is especially important in Animations as we need to get access to a nodes dimensions in a lot of cases which may be dependent on the styles applied to parent elements. With the current implementation, the `requireNode` callback triggers before parents are processed.

We should consider implementing an event based system similar to the way that used in `RegistryHandler`. The implementation could allow users to register for an event when a given node, or level of nodes has been processed and made available.

For example:

``` typescript
// fires when node with `myKey` is available
nodeHandler.on({ type: 'node', key: 'myKey' }, callback); 

// fires when the widget root is available
nodeHandler.on({ type: 'widget' }, callback); 

// fires when the projector is available
nodeHandler.on({ type: 'projector' }, callback); 
```

The benefit of this approach is that in the case of animations, you can wait for the projector to be processed before adding any animations. This ensures that all `HTMLElements` have been laid out and had styles applied.

```typescript
nodeHandler.on({ type: 'projector' }, () => {
   const myAnimatedNode = this.nodes('myKey');
  // calculate and add animation here
});
```

tomdye:
**Type:** feature

**Description:**
- adds `nodehandler` to meta
- tracks elements, widget roots and projector creation / updates
- changes the way nodes are acquired within meta
- fixes tests
- events for element creation

Resolves #662


src/NodeHandler.ts
```diff
@@ -0,0 +1,44 @@
+import { Evented } from '@dojo/core/Evented';
+import { VNodeProperties } from '@dojo/interfaces/vdom';
+import Map from '@dojo/shim/Map';
+
+export enum Type {
+	Projector = 'Projector',
+	Widget = 'Widget'
+}
+
+export default class NodeHandler extends Evented {
```
agubler: Can we make this a named export and export default below (a pattern for most widget-core modules)

src/NodeHandler.ts
```diff
@@ -0,0 +1,44 @@
+import { Evented } from '@dojo/core/Evented';
+import { VNodeProperties } from '@dojo/interfaces/vdom';
+import Map from '@dojo/shim/Map';
+
+export enum Type {
```
agubler: Can we have some documentation on these (and the rest of the module?), is `type` descriptive enough?

src/WidgetBase.ts
```diff
@@ -199,14 +203,9 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this._diffPropertyFunctionMap = new Map<string, string>();
 		this._bindFunctionPropertyMap = new WeakMap<(...args: any[]) => any, { boundFunc: (...args: any[]) => any, scope: any }>();
 		this._registries = new RegistryHandler();
+		this._nodeHandler = new NodeHandler();
```
matt-gadd: we should own this

src/NodeHandler.ts
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public add(element: Element, properties: VNodeProperties) {
+		const key = String(properties.key);
+		this._nodeMap.set(key, element);
+		this.emit({ type: key });
+	}
+
+	public addRoot(element: Element, properties: VNodeProperties) {
+		this.add(element, properties);
```
agubler: Do you need the key check for adding the root too?

src/WidgetBase.ts
```diff
@@ -176,16 +176,20 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-
 	private _boundRenderFunc: Render;
 
 	private _boundInvalidate: () => void;
 
 	private _defaultRegistry = new WidgetRegistry();
 
+	private _nodeHandler: NodeHandler;
+
+	private _projectorMountEvent: any;
```
agubler: is this really `any`?

src/WidgetBase.ts
agubler: Need to `own` this

src/WidgetBase.ts
```diff
@@ -230,22 +228,13 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-				/* istanbul ignore else: only checking for errors */
-				if (!this._nodeMap.has(key)) {
-					throw new Error(`Required node ${key} not found`);
-				}
-			});
-			this._requiredNodes.clear();
-			const dNodes = renderFunc();
-			this._nodeMap.clear();
-			return dNodes;
+			this._nodeHandler.clear();
```
agubler: does this still need to be in a `beforeRender` I'd be interested to know if we could remove that.

src/WidgetBase.ts
```diff
@@ -450,12 +466,26 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+			}
+		});
+
 		while (nodes.length) {
 			const node = nodes.pop();
 			if (isHNode(node) || isWNode(node)) {
 				node.properties = node.properties || {};
 				if (isHNode(node)) {
-					if (node.properties.key) {
+					if (rootNodes.indexOf(node) < 0 && node.properties.key) {
```
agubler: why not `=== -1`?

agubler: looks a bit confusing like that to me

src/interfaces.d.ts
```diff
@@ -413,16 +413,8 @@ export interface WidgetMetaConstructor<T extends WidgetMetaBase> {
...
-	requiredNodes: Map<string, ([ WidgetMetaBase, WidgetMetaRequiredNodeCallback ])[]>;
 	invalidate: () => void;
-}
-
-/**
- * Callback when asking widget meta for a required node
- */
-export interface WidgetMetaRequiredNodeCallback {
-	(node: HTMLElement): void;
+	nodeHandler: any;
```
agubler: type? might need to create an interface in here for `NodeHandler`

src/meta/Base.ts
```diff
@@ -1,46 +1,27 @@
...
-import Map from '@dojo/shim/Map';
-import { WidgetMetaBase, WidgetMetaProperties, WidgetMetaRequiredNodeCallback } from '../interfaces';
+import { WidgetMetaBase, WidgetMetaProperties } from '../interfaces';
 
 export class Base extends Destroyable implements WidgetMetaBase {
 	private _invalidate: () => void;
 	private _invalidating: number;
-	private _requiredNodes: Map<string, ([ WidgetMetaBase, WidgetMetaRequiredNodeCallback ])[]>;
-	protected nodes: Map<string, HTMLElement>;
+	protected nodeHandler: any;
```
agubler: Type please

src/mixins/Projector.ts
```diff
@@ -286,6 +288,24 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T)
 			return this._projection.domNode.outerHTML;
 		}
 
+		_afterRootCreateCallback(element: Element,
```
agubler: formatting a bit weird here (and the other functions)

tests/unit/meta/Dimensions.ts
```diff
@@ -2,13 +2,34 @@ import global from '@dojo/shim/global';
...
 import { stub } from 'sinon';
-import { v } from '../../../src/d';
-import { ProjectorMixin } from '../../../src/main';
 import Dimensions from '../../../src/meta/Dimensions';
-import { WidgetBase } from '../../../src/WidgetBase';
-import { ThemeableMixin } from './../../../src/mixins/Themeable';
+import NodeHandler, { Type } from '../../../src/NodeHandler';
 
 let rAF: any;
+const defaultDimensions = {
```
agubler: Don't we export these from dimensions?

src/NodeHandler.ts
tomdye: good catch

src/WidgetBase.ts
tomdye: unlikely

src/NodeHandler.ts
agubler: can we add a failing test for this if it needs the guard 😄

tomdye: way ahead of you 😉

src/WidgetBase.ts
tomdye: When else would you suggest doing it?

src/WidgetBase.ts
tomdye: ok

src/WidgetBase.ts
agubler: could you do before the render call in `__render__`? as all you are doing is creating an extra function that calls the real after clearing the node handler?

tests/unit/meta/Dimensions.ts
tomdye: no, i did add that earlier but removed again, makes sense

src/WidgetBase.ts
tomdye: done

tomdye: done

src/WidgetBase.ts
tomdye: good call

src/mixins/Projector.ts
tomdye: done

src/NodeHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+		return this._nodeMap.has(key);
+	}
+
+	public add(element: HTMLElement, properties: VNodeProperties) {
+		const key = String(properties.key);
+		this._nodeMap.set(key, element);
+		this.emit({ type: key });
+	}
+
+	public addRoot(element: HTMLElement, properties: VNodeProperties) {
```
agubler: I know this is a bit picky, but you could have a single function that does this, and three specific functions that pass their type. Like:

```ts
public add(element: HTMLElement, properties: VNodeProperties) {
    this._add(element, properties, Type.Widget);
}

public addRoot(element: HTMLElement, properties: VNodeProperties) {
    this._add(element, properties, Type.Widget);
}

public addProjector(element: HTMLElement, properties: VNodeProperties) {
    this._add(element, properties, Type.Projector);
}

private _add(element: HTMLElement, properties: VNodeProperties = {}, type: Type) {
    if (properties.key || Type.Widget) {
        this.add(element, properties);
    }
    this.emit({ type });
}
```

src/WidgetBase.ts
```diff
@@ -272,10 +262,38 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+		element: Element,
+		projectionOptions: ProjectionOptions,
+		vnodeSelector: string,
+		properties: VNodeProperties
+	): void {
+		this._addElementToNodeHandler(element, projectionOptions, properties);
+		this.onElementUpdated(element, String(properties.key));
+	}
+
+	_addElementToNodeHandler(element: Element, projectionOptions: ProjectionOptions, properties: VNodeProperties) {
```
agubler: is this private? let's add the visibility modifier.

src/WidgetBase.ts
```diff
@@ -272,10 +262,38 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+		this._currentRootNode += 1;
+
+		if (!this._projectorAttachEvent) {
```
agubler: minor nit: I believe `if (this._projectorAttachEvent === undefined) {` is more performant and also personally I find it clearer.

src/WidgetBase.ts
```diff
@@ -272,10 +262,38 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+			this._projectorAttachEvent = projectionOptions.nodeEvent.on('attached',
```
agubler: we should own `this._projectorAttachEvent` also

src/WidgetBase.ts
```diff
@@ -433,6 +440,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	}
 
 	public __render__(): VirtualDomNode | VirtualDomNode[] {
+		this._nodeHandler.clear();
```
agubler: is this the correct place for this? so it clears regardless of whether the widget was invalidated.... I don't think that this is the equivalent to when it was executed in the `@beforeRender`...

src/interfaces.d.ts
```diff
@@ -409,20 +409,21 @@ export interface WidgetMetaConstructor<T extends WidgetMetaBase> {
 	new (properties: WidgetMetaProperties): T;
 }
 
+export interface NodeHandler extends Evented {
```
agubler: we tend to have doc on exported interfaces (we also tend to suffix with `Interface`)

src/meta/Dimensions.ts
```diff
@@ -23,7 +25,7 @@ export interface DimensionResults {
 	scroll: TopLeft & Size;
 }
 
-const defaultDimensions = {
+export const defaultDimensions = {
```
agubler: thinking about this, doesn't this mean that something external could mess with the defaults?

src/NodeHandler.ts
tomdye: fair

agubler: Isn't that logically actually same?

agubler: disclosure, I updated the code :)

src/meta/Dimensions.ts
```diff
@@ -49,10 +51,16 @@ const defaultDimensions = {
 };
 
 export class Dimensions extends Base {
-	public get(key: string): Readonly<DimensionResults> {
-		this.requireNode(key);
+	constructor(properties: WidgetMetaProperties) {
+		super(properties);
+
+		this.nodeHandler.on(NodeEventType.Projector, () => {
```
matt-gadd: i think this is going to put us in a loop no?

src/WidgetBase.ts
```diff
@@ -231,51 +232,70 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
+	): void {
+		this._addElementToNodeHandler(element, projectionOptions, properties);
+		this.onElementUpdated(element, String(properties.key));
+	}
+
+	private _addElementToNodeHandler(element: HTMLElement, projectionOptions: ProjectionOptions, properties: VNodeProperties) {
+		this._currentRootNode += 1;
+
+		if (this._projectorAttachEvent === undefined) {
+			this._projectorAttachEvent = projectionOptions.nodeEvent.on('attached',
```
matt-gadd: this isn't really attached or mount (which it previously was), and more after dom update no?

src/WidgetBase.ts
```diff
@@ -436,6 +445,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this._renderState = WidgetRenderState.RENDER;
 		if (this._dirty === true || this._cachedVNode === undefined) {
 			this._dirty = false;
+			this._nodeHandler.clear();
```
agubler: I still think that this is in the wrong place, it will clear the map before running the renders and therefore always return back the default dimensions! This is were integration tests for using the meta with `WidgetBase` come in handy.

tomdye: moved down

src/NodeHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Enum to identify the type of event.
+ * Listening to 'Projector' will notify when projector is created or updated
+ * Listening to 'Widget' will notifiy when widget root is created or updated
+ */
+export enum NodeEventType {
+	Projector = 'Projector',
+	Widget = 'Widget'
+}
+
+export class NodeHandler extends Evented implements NodeHandlerInterface {
```
pottedmeat: There are two things I've been struggling with using meta that this could perhaps help with. The first is doing a reverse lookup and the second is knowing what the currently rendered keys are. I need the reverse lookup for intersection meta where I have a single observer that receives multiple events. I don't have a good use case for knowing all the keys, but I could see it being potentially helpful/related to whatever we'd have to do for a reverse lookup.

src/NodeHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	public get(key: string): HTMLElement | undefined {
+		return this._nodeMap.get(key);
+	}
+
+	public has(key: string): boolean {
+		return this._nodeMap.has(key);
+	}
+
+	public add(element: HTMLElement, properties: VNodeProperties) {
```
agubler: nit: can you add return type of `: void` please?

agubler: In other places through the PR also?

agubler: Maybe it's okay because it is defined on the interface, I don't know what the standards are around this. Can't hurt to add them I guess.

src/meta/Base.ts
```diff
@@ -1,45 +1,46 @@
...
-			callback && callbacks.push([ this, callback ]);
-			this._requiredNodes.set(key, callbacks);
-			if (!callback) {
-				this.invalidate();
+		if (!node) {
+			if (this._requestedNodeKeys.has(key)) {
+				throw new Error(`Required node ${key} not found`);
 			}
+
+			const handle = this.nodeHandler.on(key, () => {
```
agubler: maybe should own this

src/mixins/Projector.ts
```diff
@@ -286,6 +289,24 @@ export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T)
 			return this._projection.domNode.outerHTML;
 		}
 
+		protected afterRootCreateCallback(element: HTMLElement,
```
agubler: I think this formatting is really funky :) is it different from the formatting in widget base?

src/meta/Base.ts
```diff
@@ -1,45 +1,49 @@
...
-			callback && callback.call(this, node);
-		}
-		else {
-			const callbacks = this._requiredNodes.get(key) || [];
-			callback && callbacks.push([ this, callback ]);
-			this._requiredNodes.set(key, callbacks);
-			if (!callback) {
-				this.invalidate();
+		if (!node) {
+			if (this._requestedNodeKeys.has(key)) {
```
pottedmeat: Won't this throw an error if it's called twice? And is the assumption that `.has` would always be used to check to see if a node exists and never `.getNode`? Throwing an error as a side-effect of a getter could be unexpected.

tomdye: Good catch re the error @pottedmeat. I'll remove the error and only add to the requested keys if it does not already exist in there. Thanks

src/mixins/Projector.ts
tomdye: I think you're really funky

tomdye: Formatting is the same as in widgetbase

agubler: it's not the same, it is like this in widget base

```ts
	protected afterRootCreateCallback(
		element: HTMLElement,
  		projectionOptions: ProjectionOptions,
  		vnodeSelector: string,
  		properties: VNodeProperties
  	): void {
              // ... stuff
  	}
```

agubler: notice the first param is on the next line?

tomdye: my apologies, have fixed

src/WidgetBase.ts
```diff
@@ -274,7 +266,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this.onElementUpdated(element, String(properties.key));
 	}
 
-	protected afterRootUpdateCallback(
+	private afterRootUpdateCallback(
```
agubler: needs the `_` prefix

src/WidgetBase.ts
```diff
@@ -251,7 +243,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this.onElementCreated(element, String(properties.key));
 	}
 
-	protected afterRootCreateCallback(
+	private afterRootCreateCallback(
```
agubler: needs the `_` prefix

src/WidgetBase.ts
```diff
@@ -285,18 +277,13 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	}
 
 	private _addElementToNodeHandler(element: HTMLElement, projectionOptions: ProjectionOptions, properties: VNodeProperties) {
-		let currentRootNode = currentRootNodeMap.get(this) || 0;
-		const numRootNodes = numRootNodesMap.get(this);
-
-		currentRootNode += 1;
-		currentRootNodeMap.set(this, currentRootNode);
-		const isLastRootNode = (currentRootNode === numRootNodes);
+		this._currentRootNode += 1;
```
agubler: why not `this._currentRootNode++`

PR #668, opened 2017-09-10
Support Injector instances in Registry and add inject decorator

### LINKED ISSUE #648: External data property and children injection

agubler:
**Type:** feature

**Description:**

Adds `inject` decorator to support side loading properties from a Injector instance registered in the `Registry`.

Includes:
 * Adds additional API for define, get and has that deal with Injector instances (`defineInjector`, `getInjector`, `hasInjector`)
 * Add `inject` decorator that registers a function that will be called with the injector instances context and the widget properties in the `beforeProperties` phase.
 * Updates `Container` & `Themeable` to use mixin pattern for adding `inject` decorator
 * Removes `BaseInjector` that extends `WidgetBase` the `Injector` HOC
 * Renames `WidgetBase.getRegistries()` to `WidgetBase.registries`

Resolves #648


src/Injector.ts
```diff
@@ -1,154 +1,22 @@
...
-/**
- * Base context class that extends Evented and
- * returns the context using `.get()`.
- */
-export class Context<T = any> extends Evented {
-
-	private _context: T;
-
-	constructor(context: T = <T> {}) {
+	constructor(context: any) {
```
kitsonk: What was the reason/motivation for the context being untyped?  Not challenging, just wondering.

src/Registry.ts
```diff
@@ -29,31 +30,55 @@ export interface RegistryEvents extends BaseEventedEvents {
 /**
  * Widget Registry Interface
  */
-export interface Registry {
+export interface RegistryInterface {
```
kitsonk: We have traditionally not mentioned `Interface` in our interfaces.  Why the change?

kitsonk: Did it have something to do with name merging?  Could this be a use case for an abstract class and having an `AbstractRegistry` as an export?

src/decorators/inject.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface GetProperties<T = any> {
+	(context: any, properties: T): T;
+}
+
+export interface InjectConfig {
+	name: RegistryLabel;
+	getProperties: GetProperties;
+}
+
+export function inject({ name, getProperties }: InjectConfig) {
```
kitsonk: Some basic JSDoc?

src/Registry.ts
agubler: The reason was, that without a different name for the interface TS didn't actually error if the class didn't implement it! We've used a suffix of `Interface` in some of the interfaces exported from `interfaces.d.ts` which was the motivation.

I don't think `Abstract` works because the registry isn't actually abstract?

kitsonk: Makes sense.  Thanks.

src/Injector.ts
agubler: There was a challenge with the typing, I'll try it again and let you know (can't remember off the top of my head).

Also wasn't sure what the benefit really was, seen as it's injected at run time there's it didn't really give much to the consumer. The important typing is really on the `getProperties` function that gets passed the context, which the user can type as needed.

src/decorators/inject.ts
agubler: Ooops yup!

src/Injector.ts
agubler: I was imaging the challenge it seems, updated to accept a generic type.

kitsonk: I do get the point about the _value_ of it...

src/Registry.ts
```diff
@@ -129,18 +171,37 @@ export class Registry extends Evented implements Registry {
...
 			throw error;
 		});
 
 		return null;
 	}
+
+	public getInjector<T extends Injector>(label: RegistryLabel): T | null {
+		if (!this.hasInjector(label)) {
+			if (this.has(label)) {
+				console.warn(`Unable to find injector '${label.toString()}', did you mean to get a widget?`);
```
matt-gadd: i'm not convinced we need these warns, because the api is separate and returns discreet types (that extend Injector), the user thus shouldn't be able to get this wrong at design time.

src/decorators/inject.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Decorator retrieves a context from an available registry using the name and
+ * returns calls the `getProperties` function with the existing found context
+ * and passed properties to return the injected properties.
+ *
+ * @param InjectConfig the inject configuration
+ */
+export function inject({ name, getProperties }: InjectConfig) {
+	return handleDecorator((target, propertyKey) => {
+		beforeProperties(function(this: WidgetBase, properties: any) {
+			const context = this.registries.getInjector(name);
```
matt-gadd: should this be called `injector`, over `context`?

src/decorators/inject.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Map of instances against registered injectors.
+ */
+const registeredInjectorsMap: WeakMap<WidgetBase, any[]> = new WeakMap();
+
+/**
+ * Defines the contract requires for the get properties function
+ * used to map the injected properties.
+ */
+export interface GetProperties<T = any> {
+	(context: any, properties: T): T;
```
matt-gadd: is this `context`?

agubler: this is the `context` from the `Injector`

src/Container.ts
```diff
@@ -1,40 +1,26 @@
...
 ): Container<W> {
-	const {
-		getProperties = defaultMappers.getProperties,
-		getChildren = defaultMappers.getChildren
-	} = mappers;
-
-	return class extends WidgetBase<any> {
+	@inject({ name, getProperties })
+	class WidgetContainer extends WidgetBase<Partial<W['properties']>> {
+		public __setProperties__(properties: Partial<W['properties']>): void {
```
matt-gadd: have we thought how we can better support this in the future? this is the equivalent of `shouldComponentUpdate` set to `true` in react?

src/decorators/inject.ts
agubler: yup good shout

src/decorators/inject.ts
matt-gadd: what is a `context`? can we get a better name?

src/WidgetBase.ts
```diff
@@ -706,7 +706,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 			let child: WidgetBaseInterface<WidgetProperties>;
 
 			if (!isWidgetBaseConstructor(widgetConstructor)) {
-				const item = this.getRegistries().get(widgetConstructor);
+				const item = this.registries.get(widgetConstructor);
```
matt-gadd: is this the only place the `registries` are accessed? just noticed that we've renamed it back now.

src/Container.ts
agubler: I have thought about it, shall raise an enhancement issue - for now this is the only way I could support it.

src/decorators/inject.ts
agubler: I am receptive to better names.

src/WidgetBase.ts
agubler: I think everywhere else actually accesses the private (`_registries`) in `WidgetBase`, I'll update this to do that.

src/decorators/inject.ts
matt-gadd: I can't participate in suggesting a name without knowing what it is 😄

agubler: It's anything that you want to inject, it is what gets passed to the `getProperties` function that returns the values that get injected. It was referred to as a context in the previous implementation also.

```ts
const injector = new Injector(myThingToInject);

const myThingToInject = injector.get();
```

matt-gadd: injecting a `payload`?

agubler: If you want it to be payload, I don't have any strong attachment to `context` or objections to `payload` - they both work.

matt-gadd: the reason i don't like `context` is the baggage that comes from react with it

agubler: Yup I understand that. Let me update that for you - I think both are okay and if using `context` could lead to confusing that is it enough reason to change.

kitsonk: `value`?  Since you are injecting a _value_ of some sort that can be `any`.  Do we actually expect it to be `any` or is it an `object`?

agubler: Aye it can be a `any`, just as the `payload` of an `Injector` can be.

PR #681, opened 2017-09-14
Interface updates

### LINKED ISSUE #672: `key` on `WidgetProperties` to support `number` type
**Enhancement**

The `key` on `WidgetProperties` should support a `number` type.

Also it would be nice to provide a `KeyedWidgetProperties` where the `key` is mandatory.

bitpshr:
**Type:** feature

**Description:**

This PR updates the `WidgetProperties` interface so that `key` can be a number or a string. A new `KeyedWidgetProperties` interface was also added that extends `WidgetProperties` and makes `key` required.

Resolves #672

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -15,7 +15,6 @@ import {
 	DiffPropertyFunction,
 	DiffPropertyReaction,
 	DNode,
-	RegistryLabel,
 	Render,
 	VirtualDomNode,
 	WidgetMetaBase,
@@ -156,7 +155,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	/**
 	 * cached chldren map for instance management
 	 */
-	private _cachedChildrenMap: Map<RegistryLabel | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>;
+	private _cachedChildrenMap: Map<string | number | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>;
 
 	/**
 	 * map of specific property diff functions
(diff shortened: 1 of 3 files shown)


src/meta/Base.ts
```diff
@@ -15,15 +15,15 @@ export class Base extends Destroyable implements WidgetMetaBase {
...
 		return this.nodeHandler.has(key);
 	}
 
-	protected getNode(key: string): HTMLElement | undefined {
+	protected getNode(key: string | number): HTMLElement | undefined {
 		const node = this.nodeHandler.get(key);
 
 		if (!node && !this._requestedNodeKeys.has(key)) {
-			const handle = this.nodeHandler.on(key, () => {
+			const handle = this.nodeHandler.on(key as string, () => {
```
agubler: Do you think that we should actually convert this to a string, instead of casting the TS type?
 ```
`${key}`
```

Comment (kitsonk) [reply src/meta/Base.ts]: Yeah, better than assuming there is some sort of implicit coercion of numbers downstream.

File: src/WidgetBase.ts [hunk 2]
```diff
@@ -156,7 +155,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	/**
 	 * cached chldren map for instance management
 	 */
-	private _cachedChildrenMap: Map<RegistryLabel | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>;
+	private _cachedChildrenMap: Map<string | number | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>;
```
kitsonk: Why is it not `RegistryLabel | number |`...?

agubler: It actually should never have been registry label, the map always takes the key from properties and uses that (or the widget constructor). So I think it wasn't correct to support symbols.

kitsonk: 👍 cool...

bitpshr:
@agubler @kitsonk this should be ready for another review. Thanks!

PR #686, opened 2017-09-19
Simplify Widget Registries

agubler:
**Type:** feature

**Description:**

The more that registries have been explored, the more apparent it has become that the main use cases will be to support lazy loading widgets and for injecting values into the widget tree.

These changes remove `registry` as a supported property for all widgets, so that a `registry` is only supported for a `projector` (or app level). This `registry` will be passed as the `baseRegistry` for all widgets within the projector tree. 

As well as the `baseRegistry`, each widget will have access to define entries into a local registry created per instance. These locally defined entries will take precedence over entries with the same label defined in the `baseRegistry`.

The `registries` property (the `RegistryHandler`) on `WidgetBase` has been re-named to `registry`, as consumers will work with the handler as is they would a single registry.

In advanced cases, it is possible to invert the default precedence on a case by case basis, by passing a boolean with the label to the get (`get` / `getInjector`) functions, so that, if there is a item in the base registry it will be returned ahead of an item defined in the local registry. 

```ts
this.registry.get('widget', true);
this.registry.getInjector('widget', true);
```

This could be useful to enable a particular default to be overridden for a widget, for example a loading widget. Locally this would be registered as `this.registry.define('my-widget-loading', DefaultLoading);` but used with the extra boolean flag `this.registry.get('my-widget-loading', true);` so that consumers could add a custom widget to their application registry to override the `DefaultLoading` widget.

### DIFF
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,7 +1,7 @@
 import { Map } from '@dojo/shim/Map';
 import { Evented } from '@dojo/core/Evented';
 import { Constructor, RegistryLabel, WidgetBaseInterface } from './interfaces';
-import { Registry, RegistryEventObject } from './Registry';
+import { Registry, RegistryEventObject, RegistryItem } from './Registry';
 import { Injector } from './Injector';
 
 export class RegistryHandler extends Evented {
@@ -21,7 +21,7 @@ export class RegistryHandler extends Evented {
 		this._baseRegistry = baseRegistry;
 	}
 
-	public define(label: RegistryLabel, widget: Constructor<WidgetBaseInterface>): void {
+	public define(label: RegistryLabel, widget: RegistryItem): void {
 		this._registry.define(label, widget);
 	}
 
@@ -41,7 +41,7 @@ export class RegistryHandler extends Evented {
 		return this._get(label, globalPrecedence, 'get', this._registryWidgetLabelMap);
 	}
 
-	public getInjector(label: RegistryLabel, globalPrecedence: boolean = false): Injector | null {
+	public getInjector<T extends Injector>(label: RegistryLabel, globalPrecedence: boolean = false): T | null {
 		return this._get(label, globalPrecedence, 'getInjector', this._registryInjectorLabelMap);
 	}
 
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -1,96 +1,28 @@
-import { Handle } from '@dojo/interfaces/core';
 import { Map } from '@dojo/shim/Map';
-import { Destroyable } from '@dojo/core/Destroyable';
 import { Evented } from '@dojo/core/Evented';
 import { Constructor, RegistryLabel, WidgetBaseInterface } from './interfaces';
 import { Registry, RegistryEventObject } from './Registry';
 import { Injector } from './Injector';
 
-/**
- * Interface for the primary and secondary registry handles
- */
-interface Handles {
-	primary: Handle;
-	secondary?: Handle;
-}
-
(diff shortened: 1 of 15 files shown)


src/RegistryHandler.ts
```diff
@@ -1,79 +1,74 @@
...
+	}
+
+	public getInjector(label: RegistryLabel, globalPrecedence: boolean = false): Injector | null {
+		return this._get(label, globalPrecedence, 'getInjector', this._registryInjectorLabelMap);
+	}
+
+	private _get(label: RegistryLabel, globalPrecedence: boolean, getFunctionName: 'getInjector' | 'get', labelMap: Map<Registry, RegistryLabel[]>): any {
+		const registries = globalPrecedence ? [ this._baseRegistry, this._registry ] : [ this._registry, this._baseRegistry ];
+		for (let i = 0; i < registries.length; i++) {
+			const registry: any = registries[i];
```
kitsonk: No other decent type than `any`?

src/RegistryHandler.ts
```diff
@@ -1,79 +1,74 @@
...
+			const registeredLabels = labelMap.get(registry) || [];
 			if (item) {
 				return item;
 			}
-			else if (!registryWrapper.handle) {
-				registryWrapper.handle = registryWrapper.registry.on(widgetLabel, (event: RegistryEventObject) => {
-					if (event.action === 'loaded') {
+			else if (registeredLabels.indexOf(label) === -1) {
+				const handle = registry.on(label, (event: RegistryEventObject) => {
+					if (event.action === 'loaded' && (<any> this)[getFunctionName](label) === event.item) {
```
kitsonk: `(this as any)` and why the `any` cast?

src/RegistryHandler.ts
agubler: It's so we can use an index string function name that is passed. 

If the `getFunctionName` (which typing I realise is a bad name), is restricted to just `'get'` or `'getInjector'` then using `registry[getFunctionName]('')` works but not if is it's typed as `'get' | 'getInjector'`, any ideas?

```ts
private _get(label: RegistryLabel, globalPrecedence: boolean, getFunctionName: 'getInjector' | 'get', labelMap: Map<Registry, RegistryLabel[]>): any {
```

kitsonk: Hmmm... [Recent versions of TypeScript](https://www.typescriptlang.org/play/#src=class%20Registry%20%7B%0D%0A%20%20%20%20getInjector()%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20'foo'%3B%0D%0A%20%20%20%20%7D%0D%0A%0D%0A%20%20%20%20get()%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20'bar'%3B%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Afunction%20getThings(method%3A%20'get'%20%7C%20'getInjector'%2C%20registry%3A%20Registry)%20%7B%0D%0A%20%20%20%20return%20registry%5Bmethod%5D()%3B%0D%0A%7D%0D%0A%0D%0AgetThings('get'%2C%20new%20Registry())%3B%0D%0A) allow you to access properties by method name just fine, if the variable is a string literal, unless I am missing something here.

agubler: It was fine is I typed the argument is just `'get'` or just `'getInjector'` but not when I typed it as `'getInjector' | 'get'`...

kitsonk: Strange, the playground example types it as `'getInjector' | 'get'`... wonder what is wrong.

PR #687, opened 2017-09-20
use the global precedence when checking to invalidate

agubler:
**Type:** bug

**Description:**

Ensure that the `globalPrecedence` flag is used when checking if registered widgets need invalidation.

Also removes the `diffProperty` for `registry`, as no longer required.


src/WidgetBase.ts
```diff
@@ -122,7 +122,6 @@ const boundAuto = auto.bind(null);
 /**
  * Main widget base for all widgets to extend
  */
-@diffProperty('registry', ignore)
```
agubler: We don't need this special `diffProperty` for registry now it is not a supported property.

PR #689, opened 2017-09-20
`alwaysRender` Decorator

### LINKED ISSUE #676: Add support for "always widget render" hook
**Enhancement**

There are scenarios when a hook to always consider a component `dirty` and therefore re-render regardless of the result of the property diff'ing similar to React's `shouldComponentUpdate`.

An existing use-case is the `Container` HOC which needs to always re-render regardless of the properties that it receives.

In `WidgetBase`:

```ts
protected alwaysRender() {
    return false;
}
```

overriding this in a component:

```ts
protected alwaysRender() {
    return true;
}
```

Should returning true skip the property diff'ing? Should the function receive any arguments (like the props)?

agubler:
**Type:** feature

**Description:**

`alwaysRender` decorator that skip the entire diff properties process (as detailed in associated issue) and calls `invalidate`.  

Throws an error if the widget has registered custom diffs using the `diffProperty` decorator.

Resolves #676

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -152,7 +152,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	private _coreProperties: CoreProperties = {} as CoreProperties;
 
 	/**
-	 * cached chldren map for instance management
+	 * cached children map for instance management
 	 */
 	private _cachedChildrenMap: Map<string | number | Promise<WidgetBaseConstructor> | WidgetBaseConstructor, WidgetCacheWrapper[]>;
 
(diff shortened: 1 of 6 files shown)


src/WidgetBase.ts
```diff
@@ -318,28 +318,48 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-			const newProperty = this._bindFunctionProperty(properties[propertyName], this._coreProperties.bind);
-			if (registeredDiffPropertyNames.indexOf(propertyName) !== -1) {
-				runReactions = true;
-				const diffFunctions = this.getDecorator(`diffProperty:${propertyName}`);
-				for (let i = 0; i < diffFunctions.length; i++) {
-					const result = diffFunctions[i](previousProperty, newProperty);
+			changedPropertyKeys = Object.keys(properties);
+			diffPropertyResults = { ...properties };
+		}
+		else {
```
kitsonk: Just curious... This seems to be far more complex logic for the "not always render case".  Is this realising when we address this logic uniquily, we have a far more nuanced invalidation mechanism that we need to employ?

kitsonk: Actually it isn't specifically new LOC, it allows some sort of colocation of the code... Side note, this bit does feel a bit hefty from a complexity bit, I don't see any obvious ways of refactoring it, but just a thought...

agubler: Indeed it's the existing diff'ing logic that was moved into the `else` of the `alwaysRender` `if`check. It is pretty complicated though, as every property can have a different diff registered, return a diff value to the passed property and also have reactions to run.

kitsonk:
We should do whatever to land this or close it down.

agubler:
Closing for now

PR #696, opened 2017-09-22
Intersection Observer Meta

### LINKED ISSUE #644: Intersection Meta

agubler:
**Type:** feature

**Description:**

Builds on #664, updating to use the latest changes in meta (added by by #666).

It is highly likely the functionality will need to be enhanced to cater for the more advanced use-cases in the future.

Includes the intersection observer polyfil

Resolves #644

### DIFF
--- a/src/meta/Intersection.ts
+++ b/src/meta/Intersection.ts
@@ -1,0 +1,102 @@
+import global from '@dojo/shim/global';
+import WeakMap from '@dojo/shim/WeakMap';
+import Map from '@dojo/shim/Map';
+import { createHandle } from '@dojo/core/lang';
+import { Base } from './Base';
+
+interface ExtendedIntersectionObserverEntry extends IntersectionObserverEntry {
+	readonly isIntersecting: boolean;
+}
+
+interface IntersectionDetail extends IntersectionGetOptions {
+	entries: WeakMap<Element, ExtendedIntersectionObserverEntry>;
+	observer?: IntersectionObserver;
+}
+
+export interface IntersectionGetOptions {
+	root?: string;
+	rootMargin?: string;
+	thresholds?: number[];
+}
+
+export interface IntersectionResult {
+	intersectionRatio: number;
+	isIntersecting: boolean;
+}
+
+const defaultIntersection: IntersectionResult = Object.freeze({
+	intersectionRatio: 0,
+	isIntersecting: false
+});
+
+export class Intersection extends Base {
+	private readonly _details: Map<string, IntersectionDetail> = new Map();
+
+	public get(key: string, options: IntersectionGetOptions = {}): IntersectionResult {
+		let rootNode: HTMLElement | undefined;
+		const details = this._getDetails(options);
+
+		if (details.root) {
+			rootNode = this.getNode(details.root);
+			if (!rootNode) {
+				return defaultIntersection;
+			}
+		}
+		const node = this.getNode(key);
+		if (!node) {
+			return defaultIntersection;
+		}
+
+		this._getIntersectionObserver(details, rootNode).observe(node);
+		const { intersectionRatio, isIntersecting } = details.entries.get(node) || defaultIntersection;
+		return {
+			intersectionRatio,
+			isIntersecting
+		};
+	}
+
+	public has(key: string | number, options?: IntersectionGetOptions): boolean {
+		const node = this.getNode(key);
+		if (!node) {
+			return false;
+		}
+		return this._getDetails(options).entries.has(node);
+	}
+
+	private _getDetails(options: IntersectionGetOptions = {}): IntersectionDetail {
+		const cacheKey = JSON.stringify(options);
+		let cached = this._details.get(cacheKey);
+
+		if (!cached) {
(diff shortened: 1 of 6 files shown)


src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+
+	public has(key: string | number, options?: IntersectionGetOptions): boolean {
+		const node = this.getNode(key);
+		const details = this._getDetails(options);
+		return Boolean(details && node && details.entries.has(node));
+	}
+
+	private _createDetails(options: IntersectionGetOptions, rootNode?: HTMLElement): IntersectionDetail {
+		const entries = new WeakMap<HTMLElement, ExtendedIntersectionObserverEntry>();
+		const observer = new global.IntersectionObserver(this._onIntersect(entries), options);
```
pottedmeat: The root string from the options object (instead of the root node) is passed when creating an IntersectionObserver. I have a test for this in:	 'intersections accept root'

agubler: That makes a lot of sense, thanks (i'll add a test).

I actually tried the root node functionality yesterday but couldn't get it to work (just thought I had done it wrong.)

src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+		const details = { observer, entries, ...options };
+
+		this._details.set(JSON.stringify(options), details);
```
pottedmeat: So I was wondering about this when I was working on it. If the key order changes, will this encode the same thing? Is it a big deal if it doesn't? I had gone with the approach of using `JSON.stringify([root, rootMargin, thresholds])` to make sure the structure stays the same.

agubler: I see what your saying, feels like a pretty edge case though.

src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+		let details = this._getDetails(options);
+		if (!details) {
+			details = this._createDetails(options, rootNode);
+			details.observer.observe(node);
+		}
+
+		return details.entries.get(node) || defaultIntersection;
+	}
+
+	public has(key: string | number, options?: IntersectionGetOptions): boolean {
```
kitsonk: some JSDOC would be nice

src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+		const observer = new global.IntersectionObserver(this._onIntersect(entries), { ...options, root: rootNode });
+		const details = { observer, entries, ...options };
+
+		this._details.set(JSON.stringify(options), details);
+		this.own(createHandle(observer.disconnect));
+		return details;
+	}
+
+	private _getDetails(options: IntersectionGetOptions = {}): IntersectionDetail | undefined {
+		return this._details.get(JSON.stringify(options));
```
kitsonk: I am a bit concerned about GC with this... I can understand why stringifying is useful to create a _key_ here, but it is a map, which means it will have strong references to the values, which means that they won't ever get released.  Shouldn't here be a destruction handle in the constructor to `.clear()` the Map?

src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const defaultIntersection: IntersectionResult = Object.freeze({
+	intersectionRatio: 0,
+	isIntersecting: false
+});
+
+export class Intersection extends Base {
+	private readonly _details: Map<string, IntersectionDetail> = new Map();
+
+	public get(key: string, options: IntersectionGetOptions = {}): IntersectionResult {
```
kitsonk: some JSDOC would be nice

src/meta/Intersection.ts
```diff
@@ -0,0 +1,89 @@
...
+	isIntersecting: boolean;
+}
+
+const defaultIntersection: IntersectionResult = Object.freeze({
+	intersectionRatio: 0,
+	isIntersecting: false
+});
+
+export class Intersection extends Base {
+	private readonly _details: Map<string, IntersectionDetail> = new Map();
```
kitsonk: Small nit, but you could `private readonly _details = new Map<string, IntersectionDetail>();`

src/meta/Intersection.ts
agubler: We use `Map`s across the widget-core, should we raise an separate issue to discuss this?

PR #206, opened 2017-01-09
Enhancement: Update to latest dojo-interfaces and dojo-compose

agubler:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
* [x] Unit or Functional tests are included in the PR

**Description:**

Update `createWidgetBase` and related friends to use the latest changes to dojo-interfaces and dojo-compose.

Waiting on https://github.com/dojo/interfaces/pull/57 & https://github.com/dojo/compose/pull/113

Related to https://github.com/dojo/compose/issues/112

PR #700, opened 2017-10-02
[DO NOT MERGE] Use Prettier for auto formatting and code style rules

agubler:
**Type:** feature (do not merge)

**Description:**

Example of the changes from using Prettier for code formatting and styling.

There is formatting here that I'm not keen on, but if we were to embrace Prettier we'd have to accept these changes.... (it's pretty opinionated with only a small configuration API)


kitsonk:
Ewww... that little OCD I have is upset:

```ts
export function Container<W extends WidgetBase>(
```

Cuddling...  It does set out its stall as an opinionated formatter, so I guess we can't blame it too much.

kitsonk:
Ok, I have looked at it again and actually the item I objected to was actually contrary to our style guide.

The things that I have issues with:

* Cuddles type assertions (e.g. `<any>foo` instead of `<any> foo`).  Though we should be avoiding that syntax and instead use `foo as any`.
* Removes space from literals (e.g. `[foo, bar]` instead of `[ foo, bar ]`)
* Does not start certain blocks on a new line (e.g. `} else {` instead of `else {`)
* Does not brace single argument lambda/arrow functions (e.g. `foo => foo.bar` instead of `(foo) => foo.bar`).

Of all of them though, it is only the last one that I really have an issue with, and it seems I am not the only one as prettier/prettier#812 is a request.  The current status is that they are going to consider it because of the amount of feedback, though they have locked the issue.  Since it feels like it is something they will eventually deliver, I would not object to using prettier, though I would want to adopt that flag as soon as it becomes available.

agubler:
@kitsonk Thank you for your thoughts, the last item is indeed the most controversial divergence from our current styles.

I'd also like to add the flag when it became available from prettier 👍

kitsonk:
Refs dojo/meta#206

agubler:
Closing, this was just an example of the formatting changes

PR #711, opened 2017-10-09
Do not ignore package-lock.json

agubler:
**Type:** feature

**Description:**

We shouldn't need to ignore the package-lock.json anymore, it causes a failure in the `grunt release` task as it is using `npm version` which implicitly tries to add the file to commit (and the release fails). As the peer dependencies are installed with a pre publish script we should always get the latest version according to the semver dependency version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ coverage-final.lcov
 npm-debug.log
 /_apidoc
 deploy_key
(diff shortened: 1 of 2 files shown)


kitsonk:
Probably should also open an issue on meta for this so we can roll it out everywhere.

PR #713, opened 2017-10-09
WebAnimations Meta

### LINKED ISSUE #657: Web Animations API
This issue supersedes: https://github.com/dojo/widget-core/issues/70

**Enhancement** 
# Summary

In order to provide functional web animations for our widgets should provide a way to specify animations on the `vNode` level without needed to expose the underlying `HTMLElements` to the user.

Animations should be able to be controlled via changes in widget properties / state and should be removable.  Animation `effects` should have capability to be provided as functions such that they can access widget properties and `meta` to utilise `dimensions` in the values provided. 

A polyfill will be required for browsers that do not fully support the Web Animations API. 

This functionality can be achieved via a mixin which uses `meta` to access the underlying `HTMLElements` and the `afterRender` decorator to find `VNodes` with `animation` properties. 

## Animations Interface

``` typescript
interface AnimationProperties {
	id: string;
	effects: any[];
	controls?: AnimationControls;
	timing?: AnimationTimingProperties;
}

interface AnimationControls {
	play?: boolean;
	onFinish?: () => void;
	reverse?: boolean;
	cancel?: boolean;
	finish?: boolean;
	playbackRate?: number;
	startTime?: number;
	currentTime?: number;
}

interface AnimationTimingProperties {
	duration?: number;
	delay?: number;
	direction?: string;
	easing?: string;
	endDelay?: number;
	fill?: string;
	iterations?: number;
	iterationStart?: number;
}
```

## Example usage

``` typescript
_animateOpen() {
	const { size, scroll } = this.meta(Dimensions).get('content');
	const dim = this.meta(Dimensions).get('content');

	return [
		{ height: `${size.height}px` },
		{ height: `${scroll.height}px` }
	];
}

render() {
	const timing = {
		duration: 300,
		fill: 'forwards',
		easing: 'ease-in-out'
	};

	return v('div', {
		key: 'content',
		classes: this.classes(css.content),
		animate: [ {
			id: 'down',
			effects: this._animateOpen.bind(this),
			timing,
			controls: {
				play: true
			}
		}]
	}, [ 'Content' ]);
}
```

### VNode typing
One of the problems presented by this approach in that we are currently unable to `type` the `properties` of a `VNode` based on the mixins that have been applied to it's containing widget. The only way to do this currently is by adding to the underlying `VirtualDomProperties` as we do for the Themeable properties, but this is far from ideal.

Possible approaches to this include using a higher order function to wrap around the call to `v`, `animate(v('div', {}), animationProperties)` or exporting a decorated `v` on the widget instance that includes the animation properties in its typings.

### Meta get node timing #662 
Currently the `requireNode` callback in meta fires as soon as the requested node is available. This can happen before parent elements / widgets have had their properties processed and as such any styles applied to them will not yet have impacted the required node. When working with animations and accessing Dimensions via meta this is highly problematic.
We should look into implementing an events system that can allow callers to ensure specific nodes have been processed before dimensions etc are requested.

tomdye:
**Type:** feature

**Description:**
Adds a `WebAnimations` meta which allows programatic web animations to be applied to a `VNode`.

```ts
import WebAnimation from '@dojo/widget-core/meta/WebAnimation';

export default class AnimatedWidget extends WidgetBase {
    protected render() {
        const animate = {
            id: 'rootAnimation',
            effects: [
                { height: '10px' },
                { height: '100px' }
            ],
            controls: {
                play: true
            }
        };

        this.meta(WebAnimation).animate('root', animate);

        return v('div', {
            key: 'root'
        });
    }
}
```

Resolves #657


src/mixins/Animated.ts
```diff
@@ -0,0 +1,170 @@
...
+export function AnimatedMixin<T extends Constructor<WidgetBase>>(Base: T): T {
+	class Animated extends Base {
+
+		@afterRender()
+		protected decorateAfterRender(result: DNode): DNode {
+			decorate(result,
+				(node: HNode) => {
+					const { animate, key } = node.properties;
+					if (animate && key) {
+						this.meta(AnimationPlayer).add(key as string, animate, this);
```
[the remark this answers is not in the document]
tomdye: we can but it's potentially an array of arrays so wanted to avoid the extra loops. Can do though if you feel it's better?

src/interfaces.d.ts
```diff
@@ -414,6 +455,7 @@ export interface NodeHandlerInterface extends Evented {
 export interface WidgetMetaProperties {
 	invalidate: () => void;
 	nodeHandler: NodeHandlerInterface;
+	bind: any;
```
agubler: Is this really `any`?

src/meta/WebAnimation.ts
```diff
@@ -0,0 +1,160 @@
...
+				currentTime,
+				playState,
+				playbackRate,
+				startTime
+			};
+		}
+	}
+
+	afterRender() {
+		super.afterRender();
```
agubler: why `super.afterRender()`?

src/interfaces.d.ts
tomdye: the bind scope is any in `widgetBase` I believe? I guess it should probably be `WidgetBaseConstructor` ?

src/meta/WebAnimation.ts
tomdye: Is this not how you call the super function?

src/interfaces.d.ts
agubler: It wouldn't be the constructor, it would just be `WidgetBase` if it is able to be typed.

tomdye: ok

agubler:
package-lock.json conflict now

PR #718, opened 2017-10-11
Add start position information for a drag event

### LINKED ISSUE #715: Drag meta should report pointer position
**Enhancement** 

The Drag meta provider should provide the original pointer position when providing pointer meta data, thereby, position + delta = current position.

kitsonk:
**Type:** feature

**Description:**

This PR enhances the Drag meta provider by providing a `start` property on drag results.  This start property is a matrix of positional information from the pointer event.  It therefore makes it possible to determine where the pointer location is by taking one of the properties of the `start` property and adding the `delta` property.

This is useful in cases where a render function needs to instantiate something at a particular location of where the pointer event is occurring.

Resolves #715

### DIFF
--- a/src/meta/Drag.ts
+++ b/src/meta/Drag.ts
@@ -154,9 +154,7 @@ class DragController {
 			this._dragging = target;
 			state.last = state.start = getPositionMatrix(e);
 			state.dragResults.delta = createPosition();
-			if (!state.dragResults.start) {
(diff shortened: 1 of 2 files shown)


kitsonk:
I removed a redundant if block, the other uncovered else block should ideally have a test where there is a RAF before the up event which should cause the `start` to be populated.

PR #719, opened 2017-10-11
Fix Drag meta to properly model non-primary button or pointers

### LINKED ISSUE #717: Drag Meta keeps tracking drag after right click
**Bug** 

Package Version: current

**Expected behavior:**

When you right click the mouse the drag meta should not track the drag as the pointer event is managed by the context menu at that point.

**Actual behavior:**

Drag meta treats the right mouse click as a drag and gets confused once the context menu has been closed continuing to drag the mouse movement until left mouse button is clicked.

kitsonk:
**Type:** bug

**Description:**

This PR fixes issues with the Drag meta provider, where secondary button clicks (right clicks) would put the node into a drag state.  In addition it handles situations where a non primary pointer (e.g. two finger touch) occurs, which cancels the current drag state.

Resolves #717

### DIFF
--- a/src/meta/Drag.ts
+++ b/src/meta/Drag.ts
@@ -240,14 +240,10 @@ class DragController {
 		// we are offering up an accurate delta, so we need to take the last event position and move it to the start so
 		// that our deltas are calculated from the last time they are read
 		state.start = state.last;
+		// reset the delta after we have read, as any future reads should have an empty delta
+		state.dragResults.delta = createPosition();
 		// clear the start state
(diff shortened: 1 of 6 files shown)


kitsonk:
Also, fixed an issue where when there is an invalidation that occurs that isn't related to an event and the Drag result is read, it preserves the old results, which can cause issues when holding the pointer down and not moving it, but other invalidations are occurring.

tomdye:
all good and appears to fix the JLG issue @kitsonk 👍

PR #728, opened 2017-10-20
Integrate VDOM system into widget-core

### LINKED ISSUE #725: Integrate VDOM system into widget-core
**Enhancement**

We currently leverage maquette for our vdom management, this has served us well so far but we’re now getting to the point were it would be preferable if we had our own integrated into the widget system. There are few major reasons for this:

 * We have to do a lot of extra work in WidgetBase to manage `DNode`s and Widget instances. This is quite costly having to effectively iterate a tree to generate intermediate `VNode`s for Maquette, as well as the memory overheads.
 * Provide additional lifecycle hooks over the standard ones provided by the Maquette's existing implementation.
 * Freedom to extend / modify the vdom functionality specifically for widget-core.

The Proposal is to fork Maquette into the widget-core codebase and convert it to deal with `DNode` and widget instance management directly to achieve feature parity with the current implementation.

There are a few outstanding issues that this move will help facilitate:

[Support after/deferred property callbacks when creating VDOM](https://github.com/dojo/widget-core/issues/714) - We will have full control of the lifecycle so can implement a deferred property system

[Only require `key` for adjacent sibling widgets](https://github.com/dojo/widget-core/issues/661) - This will be fixed for free as part of the moving to one tree for widgets and nodes

[Additional options bag in v() that supports explicit props and attributes](https://github.com/dojo/widget-core/issues/650) - We will be able to support a more user friendly options bag

[Consider performance/memory improvements with .classes() in Themeable](https://github.com/dojo/widget-core/issues/616) - 
We will be able to take all the heavy lifting off `this.classes` by building in support for toggling classes based on presence

[Unify event model](https://github.com/dojo/widget-core/issues/508) - We will have full control of the event system

*Note:* This is a breaking change and will have knock on effects for downstream packages such as `@dojo/widgets` and `@dojo/test-extras` (big impact on this).

agubler:
**Type:** feature

**Description:**

Integrates a fork of Maquette into widget-core that has been converted to deal with the widget-core `DNode` abstraction. All widget instance management has been moved from `WidgetBase` into the `vdom` functionality.

Achieves at least feature and performance parity with the existing implementation.

*Includes updated variable names and code styling but maintains large amounts of the existing architecture and logic from Maquette which will be improved on over time.*

**Performance Benchmark Comparison between 0.1.2 (current) and 0.2.0 (new vdom)**

interactive_results

Using [js-framework-benchmark](https://github.com/krausest/js-framework-benchmark)

Resolves #725


src/vdom.ts
```diff
@@ -0,0 +1,803 @@
...
+		}
+		else {
+			if (isHNode(child)) {
+				if (child.properties.bind === undefined) {
+					(child.properties as any).bind = instance;
+					if (child.children && child.children.length > 0) {
+						filterAndDecorateChildren(child.children, instance);
+					}
+				}
+			} else {
```
agubler: formatting

src/WidgetBase.ts
```diff
@@ -156,6 +124,20 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 		this.own(this._nodeHandler);
 		this._boundRenderFunc = this.render.bind(this);
 		this._boundInvalidate = this.invalidate.bind(this);
+		this.own(this.on('element-created', ({ key, element }: any) => {
```
kitsonk: you can use a map here, which returns a single composite handle:

```
this.on({
  'element-create'({ key, element }: any) { /* ... */ },
  'eleement-updated'({ key, element: any) { /* ... */ },
  ...
});
```

Also it seems like the `any` could/should be more specific?

src/WidgetBase.ts
```diff
@@ -337,70 +270,20 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
-					}
-				}
-				nodes = [ ...nodes, ...node.children ];
-			}
-		}
+		return this._cachedDNode;
 	}
 
-	protected invalidate(): void {
+	public invalidate(): void {
```
kitsonk: why the change is visibility?

src/d.ts
```diff
@@ -81,29 +79,28 @@ export function w<W extends WidgetBaseInterface>(widgetConstructor: Constructor<
...
 		if (Array.isArray(propertiesOrChildren)) {
 			children = propertiesOrChildren;
 			properties = {};
 		}
 
-		let { classes } = properties;
-		if (typeof classes === 'function') {
-			classes = classes();
-			properties = assign(properties, { classes });
+		if (properties) {
```
kitsonk: Not sure I can see how properties could be undefined here?

src/util/DomWrapper.ts
```diff
@@ -1,7 +1,7 @@
 import { WidgetBase } from './../WidgetBase';
-import { Constructor, DNode, VirtualDomProperties, WidgetProperties } from './../interfaces';
+import { Constructor, DNode, HNode, VirtualDomProperties, WidgetProperties } from './../interfaces';
 import { v } from './../d';
-import { VNode } from '@dojo/interfaces/vdom';
+import { InternalHNode } from './../vdom';
```
kitsonk: We probably need to get rid of VNode in interfaces after we land this, right?  It is used in a couple places (like test-extras) which sort of had to deal with maquette nodes directly.

tests/support/loadJsdom.ts
```diff
@@ -29,6 +29,12 @@ global.requestAnimationFrame = (cb: (...args: any[]) => {}) => {
 	return true;
 };
 
+global.requestIdleCallback = (cb: (...args: any[]) => {}) => {
```
kitsonk: reminder that we would want to address this in https://github.com/dojo/test-extras/blob/master/src/support/loadJsdom.ts as well

src/util/DomWrapper.ts
agubler: Yup we can remove them afterwards for sure and yup test-extras is going to suffer the most from these changes. Otherwise updating widgets was just a matter of swapping `VNode`s for `DNodes`

src/WidgetBase.ts
agubler: It now get's called from outside of the widget instance (in vdom), previously it was called from within the instance in `dNodeToVNode`

tests/support/loadJsdom.ts
agubler: Yup! Thanks!

src/WidgetBase.ts
kitsonk: 👍

src/d.ts
agubler: Yup, I think could possibly remove that (it's because at one point I allowed undefined properties and changed my mind) - but there is a change following these that will remove this entire section so perhaps wait for that. I don't mind.

src/WidgetBase.ts
agubler: Nice I'll give that a try, also add the proper typings over `any`.

agubler: Updated

src/d.ts
agubler: Updated

tests/unit/meta/Drag.ts
```diff
@@ -54,7 +46,8 @@ registerSuite({
 		const widget = new TestWidget();
 		widget.append(div);
 
-		resolveRAF();
+		resolvers.resolve();
```
tomdye: why are you calling this twice here and only once elsewhere? It was only called once before.

kitsonk:
@agubler there is a regression from maquettes behaviour.  Previously `v('div.foo')` resulted in a `HTMLDivElement` with classes of `foo`, now it results in an `HTMLUnknownElement` with a `tagName` of `DIV.FOO`.  I suspect other things like ID selectors (`div#bar`) in the string also exhibit that behaviour.

agubler:
@kitsonk that’s correct, removing the support for ids and classes to be able to be passed as part of the selector was a conceous decision (the code logic was complicated and is pretty much considered an anti-pattern within Dojo 2).

It’s one of the few areas where the behaviour between the two implementations differ (on purpose!).

tests/unit/meta/Drag.ts
agubler: @tomdye this is because after the render we need to resolve the `requestIdleCallback`s, which with meta can cause an invalidation. So  means we need to resolve the `requestAnimationFrame` (and idle callbacks again) to reflect the outcome from the render.

kitsonk:
@agubler is it worth throwing then on strings that contain anything but `/[a-zA-Z0-9\-]+/`?  That would cover all HTML defined tags and only valid custom element tags.  I had a few test cases lying about using that behaviour, and while I would expect it being a minority, anyone familiar with HyperScript _might_ assume it is supported.

agubler:
@kitsonk I'm not sure there would be too much value, it would only prevent using `.` and `#` as part of the tag (that has never been supported when using `v()`). You could still pass an incorrect HTML element tag name and also it would be a hot path.

If it becomes an issue with consumers we can possibly look to address it later?

kitsonk:
> that has never been supported when using `v()`

Well, it wasn't supported because those were broken out and applied to the DOM node.

> If it becomes an issue with consumers we can possibly look to address it later?

Yeah, fine.

PR #734, opened 2017-10-30
Accept array of string classes in `v()` & implicitly remove classes that are not passed on subsequent renders

### LINKED ISSUE #616: Consider performance/memory improvements with .classes() in Themeable
**Enhancement** 

There are some improvements we can make in how we handle `this.classes()` in the Themeable mixin.  Specifically we are generating new functions every time.

### LINKED ISSUE #732: Remove classes that are not passed when using `v()`
**Enhancement**

Currently CSS classes need to be explicitly removed when using `classes` in `v()`, this means that `Themeable` needs keep track of all previously applied classes. Also this means that `Themeable` has more responsibility than it should, `Themeable` should only need to be used if the class can be themed. As it needs to track all the classes that are used within a widget even non themeable (aka `fixed`) classes need to be passed to `this.classes` to ensure that they can be removed if they aren't needed in a subsequent render.

Proposal:

 * Change `VirtualNodeProperties#classes` type to an array of `string`, `null` or `undefined` (`(null | undefined | string)[]`)
 * Rename `this.classes` to `this.theme` that only needs to be used if the class is themeable for the look up of the theme class name.
 * Update `vdom` to support implicitly removing `classes` if they are not passed on a subsequent render and filter `null`, `undefined` and zero length string values.

```ts
render() {
    // using `this.theme` with `css.root` to ensure that the correct theme class is returned
    return v('div', { classes: this.theme(css.root) }, [
        // Can mix using `this.theme` and classes directly by spreading the result of `this.theme`
        v('div', { classes: [ ...this.theme(css.title), css.header, 'global-css-class' ] }),
        // For classes that are not themeable, simply use the class name (previously fixed)
        // `falsy` values are supported and will be filtered
        v('div', { classes: [ this.selected ? css.selected : null ] })
    ]);
}
```

agubler:
**Type:** feature

**Description:**

Updates the `vdom` to implicitly remove classes when they are not passed on subsequent renders. `Themeable` is now only needed when a class can be themed other wise the string class name can directly be used (fixed classes as they are known now).

The `Themeable` has been changed to `Themed` and the API has been updated from `this.classes` to `this.theme` which accepts `string | null | (string | null)[]`.

Removes support for using `class` as a property with `v()`

Resolves #732
Resolves #616


kitsonk:
My biggest question is about usability.

What would the code look like if you have a class that has both fixed and themeable classes?

Like for example how would [this](https://github.com/dojo/widgets/blob/96646e4f6781396ecc207b0be2f800fa59e4970c/src/splitpane/SplitPane.ts#L152-L158) look after this change?

It is already _strange_ from a usability standandpoint if you aren't going to theme anything that you have to provide a map of `{ class: true }` instead of `...string[]` to a function.

agubler:
@kitsonk I added some examples in the associated issue but it would be:

```ts
classes: [
	...this.theme(css.root, direction === Direction.column ? css.column : css.row),
	css.rootFixed,
	direction === Direction.column ? css.columnFixed : css.rowFixed
]
```

agubler:
You never have to provide classes using the object any more, it's always just an array of strings, nulls or undefined (that latter two get filtered).

kitsonk:
> You never have to provide classes using the object any more, it's always just an array of strings, nulls or undefined (that latter two get filtered).

You have me sold at that!

agubler:
The API for `this.theme` has been changed slightly, to accept either a `string` or `string[]` and will return a `string` or `string[]` depending on the argument:

```ts
// will return 'myThemedCssClass'
this.theme('myCssClass');

// will return [ 'myThemedCssClass', 'myOtherThemedCssClass' ]
this.theme([ 'myCssClass',  'myOtherCssClass' ]); 
```

If a single class results in multiple classes returned from `this.theme` (due to `composes` or `extraClassses`), they will be joined by a space. 

The class name splitting is now dealt with when processing the classes in `vdom`.

agubler:
This means that when classes are constructed outside of `this.theme()`, it means that we don't have spread them when passing to `this.theme()`

```ts
const classes = [
    'myClass',
    'myOtherClass'
];

// means can just pass to `this.theme`, instead of having to spread again
v('div', { classes: this.theme(classes) });

// so no longer need to spread to pass into `this.theme`
v('div', { classes: this.theme(...classes) });
```

agubler:
@kitsonk @tom Have updated the PR again, added some comments to explain the changes (@tom you should hopefully be familiar)

PR #735, opened 2017-10-31
Update README for Themed changes

agubler:
**Type:** feature

**Description:**

README updates for the Themed changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ export const classNameTwo: string;
 
 **Important:** at runtime a JavaScript file is required to provide the processed CSS class names.
 
-The `ThemedMixin` provides a method available on the instance `this.theme()` that consumes string class names and returns the theme's equivalent class names.
+The `ThemedMixin` provides a method available on the instance `this.theme()` takes a single argument that is either a `string` class name or an array of `string` class names and returns the theme's equivalent class names as either a single string or array of strings.
 
 However, it is not always desirable to allow consumers to override styling that may be required for a widget to function correctly. These classes can be added directly to the `classes` array in `VirtualDomProperties`.
 
--- a/README.md
+++ b/README.md
@@ -183,9 +183,9 @@ class List extends WidgetBase<ListProperties> {
     protected render() {
         const { items } = this.properties;
 
-        return v('ul', { classes: { list: true } }, items.map((item) => {
+        return v('ul', { classes: 'list' }, items.map((item) => {
             const { id, highlighted, content } = item;
-            const classes = highlighted ? { highlighted: true } : { highlighted: false };
+            const classes = [ highlighted ? 'highlighted' : null ];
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -321,9 +321,9 @@ export const classNameTwo: string;
 
 **Important:** at runtime a JavaScript file is required to provide the processed CSS class names.
 
-The `ThemeableMixin` provides a method available on the instance `this.classes()` that consumes string class names and converts them into the expected class object when a widget is rendered. Classes that are passed to `this.classes()` are made available to be themed, as described in the [applying a theme](#applying-a-theme) section.
+The `ThemedMixin` provides a method available on the instance `this.theme()` that consumes string class names and returns the themes equivalent class names.
```
dylans: themes equivalent class names -> theme's equivalent class names

README.md
```diff
@@ -334,11 +334,13 @@ import { ThemeableMixin, theme } from '@dojo/widget-core/mixins/Themeable';
 @theme(css)
 export default class MyWidget extends ThemeableMixin(WidgetBase) {
     protected render() {
-        return v('div', { classes: this.classes(css.root).fixed(css.rootFixed) });
+        return v('div', { classes: [ this.theme(css.root), css.rootFixed ] });
     }
 }
 ```
 
+If an array is passed to `this.theme` then an array will be returned for example, `this.theme([ css.root, css.other ])` will return an array containing the theme's class names `[ 'themedRoot', 'themedOther' ]`.
```
dylans: be returned for example, -> be returned. For example,

agubler:
@dylans updated

README.md
```diff
@@ -334,11 +334,13 @@ import { ThemeableMixin, theme } from '@dojo/widget-core/mixins/Themeable';
 @theme(css)
 export default class MyWidget extends ThemeableMixin(WidgetBase) {
     protected render() {
-        return v('div', { classes: this.classes(css.root).fixed(css.rootFixed) });
+        return v('div', { classes: [ this.theme(css.root), css.rootFixed ] });
```
kitsonk: Isn't this supposed to be:

```diff
-return v('div', { classes: [ this.theme(css.root), css.rootFixed ] })
+return v('div', { classes: [ ...this.theme(css.root), css.rootFixed ] })
```

kitsonk: Nevermind... I see below about it... it does feel a bit confusing without an example maybe actually showing a code example after the return array and spreading it would be useful.  Also, IIRC from the changes, `this.theme()` only accepts a single argument?  If that is correct it might be better to be a bit more specific above when we say _`this.theme()` that consumes string class names and returns the theme's equivalent class names._ to something like _`this.theme()` takes a single argument that is either a string class name or an array of string class names and returns the theme's equivalent class names as either a single string or array of strings?_

PR #744, opened 2017-11-03
Move logic in custom element initialization

### LINKED ISSUE #531: Cannot add children to a Custom Element after creation
**Enhancement**
If you create a custom element then add children post creation, they are not consumed into the parent. We can achieve this using a `MutationObserver` to watch for child nodes.

maier49:
**Type:** feature

**Description:**
If I'm understanding the code correctly, the widget won't render until it's attached to the projector. That happens [here](https://github.com/dojo/widget-core/blob/master/src/customElements.ts#L250), in the callback that gets called when the custom element is attached to the DOM. So until that point we don't need to worry about a `MutationObserver`, because we can just capture the state of the element when it's attached and set the appropriate children at that time. This has the added benefit of letting things like `replaceChild` and `insertBefore` work as expected, without a lot of additional logic.

If we want to be able to add children after the element is attached then we will probably need to track the elements we've seen and use a `MutationObserver` to detect new children and create widgets for them. That should probably be a separate PR though.

Resolves #531

### DIFF
--- a/src/customElements.ts
+++ b/src/customElements.ts
@@ -226,8 +226,11 @@ export function initializeElement(element: CustomElement) {
 	}
 
 	const projector = ProjectorMixin(element.getWidgetConstructor());
-
 	const widgetInstance = new projector();
+
+	widgetInstance.setProperties(initialProperties);
+	element.setWidgetInstance(widgetInstance);
+
 	return function() {
 		// find children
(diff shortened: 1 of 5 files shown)


agubler:
Looks like the custom element functional tests are failing on CI.

PR #746, opened 2017-11-06
Correct run.html path

tomdye:
Fixes the incorrect file path in `run.html`

### DIFF
--- a/tests/run.html
+++ b/tests/run.html
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="utf-8">
 	<title>Intern suite</title>
-	<meta http-equiv="refresh" content="0;url=../../node_modules/intern/index.html">
+	<meta http-equiv="refresh" content="0;url=../node_modules/intern/index.html">
 </head>
 <body>
 	Redirecting to Intern client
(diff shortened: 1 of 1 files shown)


tests/run.html
```diff
@@ -3,7 +3,7 @@
 <head>
 	<meta charset="utf-8">
 	<title>Intern suite</title>
-	<meta http-equiv="refresh" content="0;url=../node_modules/intern/client.html?config=_build/tests/intern">
+	<meta http-equiv="refresh" content="0;url=../../node_modules/intern/index.html">
```
agubler: do you need the extra `..`?

tomdye: I guess it depends where you serve from. I ran my http server from the project root and this works correctly.

tomdye: it's got to step out of `_build/tests` to get to the `node_modules` folder, so yes, double `../..` required.

agubler: Are you running from the `_build` directory, because I think now you can just run from the `tests` directory

tomdye: thanks, that works

Issue #745, opened 2017-11-06
High Memory Usage


agubler:
Investigating memory usage for widgets has highlighted potential problems with the existing "own" paradigm that is used to destroy handles on dependant references and instances.

Consider the example below (repository available [here](http://blah)) that adds and removes 1,000 widgets that render a `div` element and a `TextNode` on a 5 second interval.

```ts
class Div extends WidgetBase<any> {
	render() {
		return v('div', { key: this.properties.key }, [ `Div Element number ${this.properties.key}` ]);
	}
}

class App extends WidgetBase {

	private _showRows = true;

	private _renderCount = 1;

	constructor() {
		super();
		setInterval(() => {
			this._showRows = !this._showRows;
			if (this._showRows) {
				this._renderCount++;
			}
			this.invalidate();
		}, 5000);
	}

	protected render() {
		const rows: any[] = [];

		if (this._showRows) {
			for (let i = 0; i < 1000; i++) {
				rows.push(w(Div, { key: i }));
			}
		}

		return v('div', [
			v('h2', [ `Render ${this._renderCount}. Rendered ${this._renderCount * 1000} Div widgets since startup.` ]),
			v('div', rows)
		]);
	}
}

const Projector = ProjectorMixin(App);
const projector = new Projector();

projector.append();
```

The expectation is that once widgets have been destroyed (called automatically within `vdom` by the parent widget), these widgets (and attached references including their DOM elements) will be available for garbage collection.

The reality is that the reference to the child widget by parent "owning" actually creates a handle that is never released (even after the widget has been destroyed).

This can be seen using the memory developer tools provided by Chrome, running the example provided take a heap dump after ever render and clear, filter the summary by `Div` to see all the `Div` widgets and `HTMLDivElements` ever created still in the memory footprint.

**Heap dump after 50 renders**

<img width="726" alt="memory-usage_and_task_manager_-_google_chrome_and_new_issue_ _dojo_widget-core" src="https://user-images.githubusercontent.com/1982678/32440347-a4e10290-c2ea-11e7-9793-0e7c143132b2.png">

Shows that all the `Div` widgets and `HTMLDivElements` created in the previous renders are still using memory.

**The handles on the widgets in the parent**

<img width="911" alt="memory-usage" src="https://user-images.githubusercontent.com/1982678/32440388-da8b4f86-c2ea-11e7-8bab-9496d8ff2caf.png">

[...]

PR #747, opened 2017-11-06
Do not own child widgets to ensure instances can be GC'd once they are no longer required

agubler:
**Type:** bug

**Description:**

As per the investigation in #745, when a parent owns its widget children it creates a handle reference that is never severed and therefore the widgets are never available for garbage collection.

Removing the automatic `own` on child widgets enables children to be GC'd correctly once they are no longer in use.

Related to #745

### DIFF
--- a/src/vdom.ts
+++ b/src/vdom.ts
@@ -363,7 +363,7 @@ export function filterAndDecorateChildren(children: undefined | DNode | DNode[],
 	}
 	children = Array.isArray(children) ? children : [ children ];
 
-	for (let i = 0; i < children.length;) {
+	for (let i = 0; i < children.length; ) {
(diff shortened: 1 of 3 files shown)


src/vdom.ts
```diff
@@ -363,7 +363,7 @@ export function filterAndDecorateChildren(children: undefined | DNode | DNode[],
 	}
 	children = Array.isArray(children) ? children : [ children ];
 
-	for (let i = 0; i < children.length;) {
+	for (let i = 0; i < children.length; ) {
```
tomdye: nit picking, but why the extra space?

agubler: linting error without it

PR #750, opened 2017-11-06
Fix behavior for multiple nodes with the same root in intersection meta

### LINKED ISSUE #727: Intersection Meta fails for multiple nodes
**Bug**

Intersection Meta will always return the default intersection for all but the first node observed by a root element. Lines 56-60 currently check if any observer at all exists on the root node, and only calls `.observe(node)` if not:

```ts
let details = this._getDetails(options);
if (!details) {
    details = this._createDetails(options, rootNode);
    details.observer.observe(node);
}
```

I believe this would be fixed with:
```ts
let details = this._getDetails(options);
if (!details || !details.entries.get(node)) {
    details = this._createDetails(options, rootNode);
    details.observer.observe(node);
}
```

maier49:
**Type:** bug

**Description:**
The intersection meta would not observe any nodes after the first called for a given `options` + `root` combination. This updates the logic so that `details` is only created once for each `options` + `root` combination but each new node requested is observed. A default value for details is provided for each node to avoid observing that node more than once.

Resolves #727

### DIFF
--- a/src/meta/Intersection.ts
+++ b/src/meta/Intersection.ts
@@ -53,11 +53,8 @@ export class Intersection extends Base {
 			return defaultIntersection;
 		}
 
-		let details = this._getDetails(options);
-		if (!details || !details.entries.get(node)) {
-			if (!details) {
-				details = this._createDetails(options, rootNode);
-			}
+		let details = this._getDetails(options) || this._createDetails(options, rootNode);
+		if (!details.entries.get(node)) {
 			details.entries.set(node, defaultIntersection);
 			details.observer.observe(node);
 		}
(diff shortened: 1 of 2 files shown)


src/meta/Intersection.ts
```diff
@@ -54,8 +54,11 @@ export class Intersection extends Base {
 		}
 
 		let details = this._getDetails(options);
-		if (!details) {
-			details = this._createDetails(options, rootNode);
+		if (!details || !details.entries.get(node)) {
+			if (!details) {
```
matt-gadd: am i missing a subtlety here? (presuming if there isn't any details that there won't be an entry?), could the code not just be:

```typescript
details = details || this._createDetails(options, rootNode)
if (!details.entries.get(node)) {
    details.entries.set(node, defaultIntersection);
    details.observer.observe(node);
}
```

maier49: Yeah no reason not to simplify it that way. Code's been updated

PR #756, opened 2017-11-07
Support Adding and Removing Multiple class in IE11

agubler:
**Type:** bug

**Description:**

* Support adding and removing multiple classes in IE11
* Add IE11 to browserstack configuration.
* Added two other fixes for tests in IE11

Resolves https://github.com/dojo/widgets/issues/367

### DIFF
--- a/intern.json
+++ b/intern.json
@@ -62,6 +62,7 @@
 			},
 			"environments+": [
 				{ "browserName": "edge" },
+				{ "browserName": "internet explorer", "version": "11" },
 				{ "browserName": "chrome", "platform": "WINDOWS" },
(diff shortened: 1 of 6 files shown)


kitsonk:
Also, I tried this branch and can confirm it resolves the issue.

kitsonk:
I was also thinking is there any benefit to feature detecting the issue and branching... I don't know if there is a huge performance difference between the old path and the new path, but I suspect it every bit can help at the vdom level.  Also, it could become a has flag and be built out/in of builds as well using static optimize.

agubler:
I did think about that when making the change, I have a benchmark taken before the change so can certainly run that again now but I suspect the margins will be neglible.

tomdye:
Good catch

PR #764, opened 2017-11-09
Skip the main set properties logic for the initial properties set

### LINKED ISSUE #762: Skip set properties process for initial cycle if there are no custom diffs
**Enhancement** 

There is no need to run the full set properties logic for the initial cycle when there are no registered custom diff functions. 

We should just set the incoming properties as the widgets properties and set the changed keys accordingly.

Proposal: Add private  `_initialProperties` flag to `WidgetBase` and include the logic described above.

agubler:
**Type:** feature

**Description:**

Simplify the set properties logic for the initial set properties 

Resolves #762

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -71,6 +71,11 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
 	private _dirty: boolean;
 
 	/**
+	 * Indicates if it is the initial set properties cycle
+	 */
+	private _initialProperties = true;
+
+	/**
 	 * internal widget properties
 	 */
 	private _properties: P & WidgetProperties & { [index: string]: any };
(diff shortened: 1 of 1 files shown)


src/WidgetBase.ts
```diff
@@ -210,28 +215,41 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> extends E
...
 	public __setProperties__(originalProperties: this['properties']): void {
 		const properties = this._runBeforeProperties(originalProperties);
-		const changedPropertyKeys: string[] = [];
-		const allProperties = [ ...Object.keys(properties), ...Object.keys(this._properties) ];
-		const checkedProperties: string[] = [];
-		const diffPropertyResults: any = {};
 		const registeredDiffPropertyNames = this.getDecorator('registeredDiffProperty');
-		let runReactions = false;
+		const changedPropertyKeys: string[] = [];
+		const propertyNames = Object.keys(properties);
```
kitsonk: just a reminder that on IE11 and any properties that are symbols, they will come back with `Object.keys`.  I don't know if you need symbol safe `Object.keys` here or not.

agubler: We already used `Object.keys` within `__setProperties__` to get the property names, so this hasn't really changed (i just hoisted it, so I could share)

PR #765, opened 2017-11-10
Invalidator Fixes: Inject invalidator via setter and reset renderState in scheduleRender

agubler:
**Type:** bug

**Description:**

Having an optional constructor argument for the parent invalidate function is unintuitive and horrible for end users needing to know that they have to pass the arguments on when anything is done in a widget constructor. Moves the injection of the parent invalidate function to a public setter that can only be set once.

The `WidgetBase` render state needs to be reset after running `__setProperties__()` and `__setChildren__()` in `Projector#scheduleRender()`

Related to #745

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -135,7 +135,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 			this._parentInvalidator = invalidator;
 		}
 		else {
-			console.warn('Unable to update parent invalidator after if has been initially set');
+			console.warn('Unable to update parent invalidator after it has been set');
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


src/WidgetBase.ts
```diff
@@ -134,6 +130,15 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 		this._registry.on('invalidate', this._boundInvalidate);
 	}
 
+	public set parentInvalidator(invalidator: Function) {
+		if (this._parentInvalidator === undefined) {
+			this._parentInvalidator = invalidator;
+		}
+		else {
+			console.warn('Unable to update parent invalidator after it has been set');
```
agubler: no test

agubler: added

PR #768, opened 2017-11-13
Custom element wrapper

agubler:
**Type:** feature

**Description:**

For custom elements that are written to deal with Dojo 2 custom element children only, add support for widget-core to treat them as widgets meaning that they can be authored without needing to add specific code to deal with custom elements. 

**Enabling usage as follows:**

Using as a Dojo Widget:

```ts
class Example extends WidgetBase {

  private _onSelected(data: any) {
    console.log(data);
  }

  protected render() {
    return w(Menu, { onSelected: this._onSelected }, [
      w(MenuItem, { key: 'a', title: 'Menu Item A' }),
      w(MenuItem, { key: 'b', title: 'Menu Item B', selected: true }),
      w(MenuItem, { key: 'c', title: 'Menu Item C' })
    ]);
  }
}
```

Using as a Custom Element:

```html
<demo-menu id="menu">
  <demo-menu-item id="a" title="Menu Item A"></demo-menu-item>
  <demo-menu-item id="b" title="Menu Item B"></demo-menu-item>
  <demo-menu-item id="c" title="Menu Item C" selected></demo-menu-item>
</demo-menu>
<script>
  a.data = { foo: 'bar' };
  menu.addEventListener('selected', function(event) {
    const detail = event.detail;
    console.log(detail);
  });
</script>
```

### DIFF
--- a/src/customElements.ts
+++ b/src/customElements.ts
@@ -4,6 +4,7 @@ import global from '@dojo/shim/global';
 import { Constructor, DNode, HNode, VirtualDomProperties, WidgetProperties } from './interfaces';
 import { WidgetBase } from './WidgetBase';
 import { v, w } from './d';
+import { DomWrapper } from './util/DomWrapper';
 import { ProjectorMixin } from './mixins/Projector';
 import { InternalHNode } from './vdom';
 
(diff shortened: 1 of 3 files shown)


src/customElements.ts
```diff
@@ -102,6 +108,11 @@ export interface CustomElementDescriptor {
 	 * Initialization function called before the widget is created (for custom property setting)
 	 */
 	initialization?: CustomElementInitializer;
+
+	/**
+	 *
```
agubler: jsdoc

src/customElements.ts
```diff
@@ -121,6 +132,43 @@ export interface CustomElement extends HTMLElement {
 	setWidgetInstance(instance: ProjectorMixin<any>): void;
 }
 
+export type DomToWidgetWrapperProperties = VirtualDomProperties & WidgetProperties;
+
+export type DomToWidgetWrapper = Constructor<WidgetBase<DomToWidgetWrapperProperties>>;
+
+export function DomToWidgetWrapper(domNode: CustomElement): DomToWidgetWrapper {
```
agubler: add some jsdoc

src/customElements.ts
```diff
@@ -60,6 +61,11 @@ export interface CustomElementInitializer {
 	(properties: WidgetProperties): void;
 }
 
+export enum ChildrenType {
```
kitsonk: Just a question/random thought... for HNode and VNode we use symbols to tag the types, here we are using string enums.  Conscious decision or unconscious?

agubler: It was conscious as an `enum` seemed more appropriate as part of a consumer API

PR #771, opened 2017-11-13
Add `onAttach` and `onDetach` lifecycle hooks

### LINKED ISSUE #766: Add widget unmount lifecycle method
**Enhancement** 

Currently, there doesn't appear to be a way to tie into the widget lifecycle for when a widget is unmounted. This can be useful for cleanup activities that should occur, such as cleaning up event listeners (that can't be replaced with property callbacks) when a widget is unloaded.

Package Version: 0.1.2

**Code**

```ts
class Widget extends WidgetBase {
	constructor() {
		super();
		global.addEventListener('message', this.doSomething);
	}

	doSomething: () => {};

	// There's currently not a place to cleanup this event listener
}
```

**Expected behavior:**

```ts
class Widget extends WidgetBase {
	constructor() {
		super();
		global.addEventListener('message', this.doSomething);
	}

	doSomething: () => {};

	onUmount() {
		global.removeEventListener('message', this.doSomething);
	}

}
```

**Actual behavior:**
 There currently doesn't seem to be a hook into the lifecycle to accomplish this.

agubler:
**Type:** feature

**Description:**

Adds two lifecycle methods `onAttach` and `onDetach` for when a widget has attached and when it is detached.

Resolves #766

### DIFF
--- a/src/WidgetBase.ts
+++ b/src/WidgetBase.ts
@@ -122,6 +122,12 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 			onElementUpdated: (element: HTMLElement, key: string) => {
 				this.onElementUpdated(element, key);
 			},
+			onAttach: (): void => {
+				this.onAttach();
+			},
+			onDetach: (): void => {
+				this.onDetach();
+			},
 			nodeHandler: this._nodeHandler,
 			registry: () => {
 				return this.registry;
@@ -166,6 +172,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 	 * @param key The vdom node's key.
 	 */
 	protected onElementUpdated(element: Element, key: string): void {
+		// Do nothing by default.
+	}
+
+	protected onAttach(): void {
+		// Do nothing by default.
+	}
+
+	protected onDetach(): void {
 		// Do nothing by default.
 	}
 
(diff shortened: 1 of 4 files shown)


src/WidgetBase.ts
```diff
@@ -169,6 +175,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 		// Do nothing by default.
 	}
 
+	protected onAttach(): void {
```
kitsonk: Any thought to this just being an optional method on `WidgetBaseInterface`?

kitsonk: also, you will essentially always do a noop on _most_ widgets on attachment, and I don't know the performance impact, but invoking potentially hundreds of the same function seems 😕 where as testing for presence and only calling if present feels like it would be more efficient...

agubler: We do the same for the `onElementCreated` / `onElementUpdated`, these guys get called on `requestIdleCallback` so certainly should reduce the performance impact - but I can look at the optional method.

agubler: Problem is I don't want to put it on the interface (it's protected), so I don't think we can type is as such.

kitsonk: Yeah, optional abstract isn't allowed and protected cannot be enforced on an interface... You ruin all my best ideas... 😝

PR #774, opened 2017-11-14
Typescript 2.6.1

bryanforbes:
**Type:** enhancement

**Description:**

Relates to dojo/meta#210


src/WidgetBase.ts
```diff
@@ -39,9 +39,9 @@ interface ReactionFunctionConfig {
 	reaction: DiffPropertyReaction;
 }
 
-export interface WidgetAndElementEvent extends EventTypedObject<'properties:changed'> {
-	key: string;
-	element: HTMLElement;
+export interface WidgetAndElementEvent<T extends EventType> extends EventObject<T> {
```
agubler: Don't think this get used anymore?

src/WidgetBase.ts
```diff
@@ -78,7 +78,7 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 	/**
 	 * Array of property keys considered changed from the previous set properties
 	 */
-	private _changedPropertyKeys: string[] = [];
+	private _changedPropertyKeys: (string | number)[] = [];
```
agubler: This should just be `string`, it is the actual key of the property on the properties object

src/WidgetBase.ts
```diff
@@ -165,15 +164,15 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
...
+	protected onElementUpdated(element: Element, key: string | number): void {
 		// Do nothing by default.
 	}
 
 	public get properties(): Readonly<P> & Readonly<WidgetProperties> {
 		return this._properties;
 	}
 
-	public get changedPropertyKeys(): string[] {
+	public get changedPropertyKeys(): (string | number)[] {
```
agubler: As above, this should just be `string[]`

agubler: And all the other references to `changedPropertyKeys`

agubler:
Looks reasonable, just a conflict on `vdom.ts` now, you're welcome :)

PR #790, opened 2017-11-29
Allow Injector's initial payload to be optional.

### LINKED ISSUE #753: Remove `set()` and `_payload` from Injector
**Enhancement**

[`Injector`](https://github.com/dojo/widget-core/blob/master/src/Injector.ts) is commonly used a base class that provides itself and its own data via Injector's `get()` method. This is so common that [the tutorial even uses the Injector in this way](https://dojo.io/tutorials/1010_containers_and_injecting_state/#1). It extends `Injector` then passes an empty object literal to `super({})` and overrides `get()`.

The problem is, `set()` and `_payload` mean nothing in when an Injector is extended.

**Code**

```typescript
class MyInjector extends Injector {
    updated: number;
    constructor() {
        super({});
        this.updated = Date.now();
    }
    get(): MyInjector {
        return this;
    }
}
```

```typescript
const injector: MyInjector;

// Somebody from Dojo 1 might expect this to update properties on the injector
injector.set({ updated: Date.now() })
```

I recommend removing `set()` and `_payload` from `Injector` or make an Injector abstract/interface/or base class that only has `get()` and extends `Evented` for invalidate.

mwistrand:
**Type:** feature

**Description:**

Makes the `payload` passed to the `Injector` constructor optional. When no payload is provided, `this` will be used as the initial payload. This has the benefit of simplifying custom injectors used by simpler applications that do not need a full-featured store, allowing the injector itself to be used as the store. For example, the injector created in the [state management tutorial](https://dojo.io/tutorials/1010_containers_and_injecting_state/#1) must override `get` and pass an empty object to `super`:

```
class ApplicationContext extends Injector {
    constructor() {
        super({});
        // ...
    }

    get() {
        return this;
    }
}
```

With these changes, `super` can be called without an argument and `get` will not need to be modified:

```
class ApplicationContext extends Injector {
    constructor() {
        super();
        // ...
    }
}
```

Resolves #753

### DIFF
--- a/README.md
+++ b/README.md
@@ -805,7 +805,7 @@ class MyClass extends WidgetBase {
 
 There is built in support for side-loading/injecting values into sections of the widget tree and mapping them to a widget's properties. This is achieved by registering a `@dojo/widget-core/Injector` instance against a `registry` that is available to your application (i.e. set on the projector instance, `projector.setProperties({ registry })`).
 
-Create an `Injector` instance and pass the `payload` that needs to be injected to the constructor:
+Create an `Injector` instance and pass an optional `payload` that needs to be injected to the constructor (if no `payload` is provided, then the injector itself is used as the initial payload):
 
 ```ts
 const injector = new Injector({ foo: 'baz' });
--- a/src/Injector.ts
+++ b/src/Injector.ts
@@ -2,14 +2,14 @@ import { Evented } from '@dojo/core/Evented';
 
 export class Injector<T = any> extends Evented {
 
-	private _payload: T;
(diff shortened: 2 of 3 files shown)


mwistrand:
After discussing this further with @matt-gadd and @agubler, we determined that the pattern described by the underlying issue is an abuse of the `Injector`, as it conflates the injector and the state. Rather than update the `Injector` to accommodate this pattern, it would be better to leave the `Injector` as it is and instead refactor the state into its own object, with the injector being invalidated on state changes.

PR #791, opened 2017-11-29
Update I18n mixin to use injector pattern.

### LINKED ISSUE #781: Switch `I18nMixin` to an injector pattern like `ThemedMixin`
**Enhancement**

The general pattern for injecting state into the widget tree is to use a registry and define the state by using an injector. The `ThemedMixin` uses this pattern to ensure that `themes` are passed to all widgets that have the `ThemedMixin` applied. 

Currently the the `I18nMixin` uses an observable from the `@dojo/i18n` package to subscribe to these changes and ensure that the widget is invalidated when it changes. We should consider adopting the injector pattern over the observable.

The `i18n` package, if used without widget core would still be useable with the observable and the `switchLocale` function.

mwistrand:
**Type:** feature

**Description:**

Updates the `I18n` mixin to use the injector pattern for passing locale data to children and removes the locale observer from the `I18n` constructor.

Resolves #781

### DIFF
--- a/src/mixins/I18n.ts
+++ b/src/mixins/I18n.ts
@@ -4,15 +4,19 @@ import i18n, {
 	Bundle,
 	formatMessage,
 	getCachedMessages,
-	Messages,
-	observeLocale
+	Messages
 } from '@dojo/i18n/i18n';
-import { Constructor, DNode, WidgetProperties, VirtualDomProperties } from './../interfaces';
-import { WidgetBase } from './../WidgetBase';
-import { afterRender } from './../decorators/afterRender';
 import { isHNode } from './../d';
+import { afterRender } from './../decorators/afterRender';
+import { inject } from './../decorators/inject';
+import { Constructor, DNode, WidgetProperties, VirtualDomProperties } from './../interfaces';
+import { Injector } from './../Injector';
+import { Registry } from './../Registry';
+import { WidgetBase } from './../WidgetBase';
 
-export interface I18nProperties extends WidgetProperties {
+export const INJECTOR_KEY = Symbol('i18n');
+
+export interface LocaleData {
 	/**
 	 * The locale for the widget. If not specified, then the root locale (as determined by `@dojo/i18n`) is assumed.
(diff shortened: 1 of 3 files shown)


tests/unit/mixins/I18n.ts
```diff
@@ -82,25 +83,45 @@ registerSuite('mixins/I18nMixin', {
...
 
-				switchLocale('fr');
+				const result = localized.__render__();
+				assert.strictEqual(result.properties!['lang'], 'fr');
+				assert.strictEqual(result.properties!['dir'], 'ltr');
+			},
+
+			'properties can be registered with helper'() {
+				const registry = new Registry();
+				registerI18nInjector({ locale: 'fr' }, registry);
```
matt-gadd: we should probably have a test to show that you can change the locale via the injector `set()`?

mwistrand: Done!

src/mixins/I18n.ts
```diff
@@ -72,21 +78,33 @@ export interface I18nMixin {
...
+	return injector;
+}
+
 export function I18nMixin<T extends Constructor<WidgetBase<any>>>(Base: T): T & Constructor<I18nMixin> {
-	class I18n extends Base {
+	@inject({
+		name: INJECTOR_KEY,
+		getProperties: (localeData: LocaleData, properties: I18nProperties) => {
+			const { locale, rtl } = localeData;
+			const injected: Partial<I18nProperties> = {};
```
agubler: Does this need to be a Partial?

agubler: would something like this work?
```ts
getProperties: (localeData: LocaleData, properties: I18nProperties) => {
    const { locale = localData.locale, rtl = localData.rtl } = properties;

    return {
        locale,
        rtl
    };
}
```

mwistrand: Yes, that would work. I've noticed elsewhere that an empty object is returned if nothing needs to be changed, so I kept that here.

PR #792, opened 2017-11-30
Add custom element decorator

### LINKED ISSUE #788: Support defining custom elements via a decorator
**Enhancement** 
The ergonomics for creating a custom element from a dojo widget are currently a bit fiddly via the `customElementDescriptor`, it would be nice if we could support a simplified form as a decorator on the component.

As part of the simplification we'd like to drop or do better with dealing with things automatically for the author. that would mean that they would no longer have to specify a mapping for anything. events would automatically drop the `on` and be lowercased? we would no longer want to support changing the property names or attribute names either I don't think as so far it's caused more trouble than it's worth?

Something like:
```ts
@customElement({
  tag: 'my-foo',
  properties: [ 'foo', 'bar' ],
  attributes: [ 'baz', 'qux' ],
  events: [ 'onSomething' ]
})
class Foo extends WidgetBase {
}
```

The only potential last thing we'd need is still functionality for the `initialization` function.

maier49:
**Type:** feature

**Description:**
Adds a decorator that creates a custom element
Resolves #788

### DIFF
--- a/README.md
+++ b/README.md
@@ -1183,149 +1183,50 @@ class MyWidgetWithTsx extends Themed(WidgetBase)<MyProperties> {
 Widgets can be turned into [Custom Elements](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) with
 minimal extra effort.
 
-There are two ways to turn a widget into a custom element.
+The `customElement` decorator can be used to annotate the widget class
+that should be converted to a custom element,
 
-#### Using the `CustomElementDescriptor` factory and `@dojo/cli`
+```typescript
+interface MyWidgetProperties {
+	onClick: (event: Event) => void;
+	foo: string;
+	bar: string;
+}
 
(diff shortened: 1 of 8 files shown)


kitsonk:
Additions to the README?

README.md
```diff
@@ -1183,7 +1183,11 @@ class MyWidgetWithTsx extends Themed(WidgetBase)<MyProperties> {
 Widgets can be turned into [Custom Elements](https://www.w3.org/TR/2016/WD-custom-elements-20161013/) with
 minimal extra effort.
 
-Just create a `CustomElementDescriptor` factory and use the `@dojo/cli` build tooling to do the rest of the work,
+There are two ways to turn a widget into a custom element.
```
matt-gadd: can we get rid of the description of using the CustomElementDescriptor completely? we plan to deprecate it as user facing feature in the future.

maier49: The README is updated now to only mention the decorator.

PR #801, opened 2017-12-12
652 - Registry decorator

### LINKED ISSUE #652: Decorator to define registry items in the default registry

tomdye:
**Type:** feature

**Description:**

Allows registry items to be defined via a decorator.
Also, `Registry` will automatically look for a widget constructor as the `default` import when an esModule is loaded.
Example:

```ts
@registry('reg-widget-1', Widget1) // single widget
class TestWidget1 extends WidgetBase {
	render() {
		return w('reg-widget-1', {}); // will render Widget1
	}
}

@registry({
	'reg-widget-1': Widget1,
	'reg-widget-2': Widget2
}) // multiple widgets
class TestWidget2 extends WidgetBase {
	render() {
		return [ w('reg-widget-1', {}), w('reg-widget-2', {}) ];
	} // will render Widget1, Widget2
}
```

Registry decorator can also accept promises the same way that the standard registry does ie.

```ts
@registry('reg-widget-1', () => import('path/to/Widget1')) // single widget
class TestWidget1 extends WidgetBase {
	render() {
		return w('reg-widget-1', {}); // will render Widget1
	}
}
```

Resolves #652

### DIFF
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -84,6 +84,11 @@ export function isWidgetBaseConstructor<T extends WidgetBaseInterface>(item: any
 	return Boolean(item && item._type === WIDGET_BASE_TYPE);
 }
 
+export interface ESMDefaultWidgetBase<T> {
+	default: Constructor<T>;
+	__esModule: boolean;
+}
+
 /**
  * The Registry implementation
  */
@@ -164,6 +169,11 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 		this._widgetRegistry.set(label, promise);
 
 		promise.then((widgetCtor) => {
+
+			if (this._ctorIsDefaultExport<T>(widgetCtor)) {
+				widgetCtor = ((widgetCtor as any) as ESMDefaultWidgetBase<T>).default;
+			}
+
 			this._widgetRegistry.set(label, widgetCtor);
 			this.emitLoadedEvent(label, widgetCtor);
 			return widgetCtor;
@@ -172,6 +182,16 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 		});
 
 		return null;
+	}
+
+	private _ctorIsDefaultExport<T>(module: WidgetBaseConstructor | ESMDefaultWidgetBase<T>): boolean {
+		// __esModule
+		// Object.default
+		// default is widgetbase consructor: isWidgetBaseConstructor
+
+		return (module.hasOwnProperty('__esModule') &&
+			module.hasOwnProperty('default') &&
+			isWidgetBaseConstructor((module as ESMDefaultWidgetBase<T>).default));
 	}
 
 	public getInjector<T extends Injector>(label: RegistryLabel): T | null {
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -185,11 +185,7 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 	}
 
 	private _ctorIsDefaultExport<T>(module: WidgetBaseConstructor | ESMDefaultWidgetBase<T>): boolean {
-		// __esModule
-		// Object.default
-		// default is widgetbase consructor: isWidgetBaseConstructor
-
-		return (module.hasOwnProperty('__esModule') &&
+		return (module && module.hasOwnProperty('__esModule') &&
 			module.hasOwnProperty('default') &&
 			isWidgetBaseConstructor((module as ESMDefaultWidgetBase<T>).default));
 	}
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -89,6 +89,15 @@ export interface ESMDefaultWidgetBase<T> {
 	__esModule: boolean;
 }
 
+export function isWidgetConstructorDefaultExport<T>(item: any): item is ESMDefaultWidgetBase <T> {
+	return Boolean(
+		item &&
+		item.hasOwnProperty('__esModule') &&
+		item.hasOwnProperty('default') &&
+		isWidgetBaseConstructor(item.default)
+	);
+}
+
 /**
  * The Registry implementation
  */
@@ -170,8 +179,8 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 
 		promise.then((widgetCtor) => {
 
-			if (this._ctorIsDefaultExport<T>(widgetCtor)) {
-				widgetCtor = ((widgetCtor as any) as ESMDefaultWidgetBase<T>).default;
+			if (isWidgetConstructorDefaultExport<T>(widgetCtor)) {
+				widgetCtor = widgetCtor.default;
 			}
 
 			this._widgetRegistry.set(label, widgetCtor);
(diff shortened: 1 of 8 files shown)


src/Registry.ts
```diff
@@ -164,6 +169,11 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 		this._widgetRegistry.set(label, promise);
 
 		promise.then((widgetCtor) => {
+
+			if (this._ctorIsDefaultExport<T>(widgetCtor)) {
```
agubler: Why does this have to be on the registry instance? Could you have done something like this (not on instance)? 

```ts
export function isWidgetConstructorDefaultExport<T>(item: any): item is ESMDefaultWidgetBase <T> {
   return Boolean(
       module && 
       module.hasOwnProperty('__esModule') && 
       module.hasOwnProperty('default') && 
       isWidgetBaseConstructor(module.default)
   );
}
```

and used it like this (might have to use a different variable for the assignment):

```ts
if (isWidgetConstructorDefaultExport<T>(widgetCtor)) {
    widgetCtor = widgetCtor.default;
}
```

src/WidgetBase.ts
```diff
@@ -496,6 +498,14 @@ export class WidgetBase<P = WidgetProperties, C extends DNode = DNode> implement
 
 		return dNode;
 	}
+
+	private _runAfterConstructors(): void {
+		const afterConstructors = this.getDecorator('constructor');
```
agubler: should we call this `afterConstructor`?

src/decorators/registry.ts
```diff
@@ -0,0 +1,25 @@
+import { handleDecorator } from './handleDecorator';
+import { RegistryItem } from '../Registry';
+
+export interface RegistryConfig {
+	[ name: string ]: RegistryItem;
```
agubler: do we normally add spaces like this for index types?

src/decorators/registry.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+/**
+ * Decorator that can be used to register a widget with the registry
```
agubler: Perhaps explain that it will register the widget with the widget's local registry?

src/decorators/registry.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function registry(nameOrConfig: string | RegistryConfig, loader?: RegistryItem) {
```
agubler: Do you think this would be better overloaded rather than with an optional type? For example if the first argument is `RegistryConfig` you shouldn't be able to pass a `loader` argument?

src/decorators/registry.ts
```diff
@@ -0,0 +1,25 @@
...
+	return handleDecorator((target, propertyKey) => {
+		target.addDecorator('constructor', function(this: any) {
+			if (typeof nameOrConfig === 'string') {
+				this.registry.define(nameOrConfig, loader);
+			} else {
+				Object.keys(nameOrConfig).forEach(name => {
```
agubler: Until we move to prettier, at the moment the style guide is for `name` to be surrounded by parentheses `(name)`

src/Registry.ts
tomdye: Thanks @agubler was looking for assistance with these typings

src/WidgetBase.ts
tomdye: I aligned it with `beforeRenders`, `afterRenders`. it returns an array after all.

src/decorators/registry.ts
tomdye: I thought it was a consistent styling across all single line uses of square brackets

src/decorators/registry.ts
tomdye: Not a problem, I thought we'd already dropped brackets for single params.

src/WidgetBase.ts
agubler: I meant the decorator name, `this.getDecorator('constructor')` -> `this.getDecorator('afterConstructor')`

src/Registry.ts
```diff
@@ -84,6 +84,20 @@ export function isWidgetBaseConstructor<T extends WidgetBaseInterface>(item: any
 	return Boolean(item && item._type === WIDGET_BASE_TYPE);
 }
 
+export interface ESMDefaultWidgetBase<T> {
+	default: Constructor<T>;
+	__esModule: boolean;
+}
+
+export function isWidgetConstructorDefaultExport<T>(item: any): item is ESMDefaultWidgetBase <T> {
```
agubler: weird space between `ESMDefaultWidgetBase` and the generic

src/decorators/registry.ts
```diff
@@ -0,0 +1,27 @@
...
+import { RegistryItem } from '../Registry';
+
+export interface RegistryConfig {
+	[name: string]: RegistryItem;
+}
+
+/**
+ * Decorator that can be used to register a widget with the calling widgets local registry
+ */
+export function registry(nameOrConfig: string, loader: RegistryItem): DecoratorHandler;
```
agubler: @matt-gadd when we talked about this before, did we decide not to use `registry` but `define` as the decorators name?

agubler:
Can you add to the description that the changes will now automatically peel the default for all lazy registry items?

src/decorators/registry.ts
tomdye: `define` gets picked up by webpack according to @matt-gadd , so that's a negatory

src/Registry.ts
tomdye: won't matter with `prettier`

src/decorators/registry.ts
agubler: `@define` gets picked up by webpack? In what way?

src/Registry.ts
agubler: No prettier yet in this repo ;)

src/decorators/registry.ts
matt-gadd: @agubler reserved word pretty much for AMD.

tomdye:
Raised https://github.com/dojo/widget-core/issues/810 to track readme update

PR #803, opened 2017-12-13
Move createResolvers() from widget-core to test-extras.

edhager:
NOTE: this PR needs a new version of dojo/test-extras.  For now, a tarball of test-extras is included.  Do not put this PR to master until test-extras is released.

I moved createResolvers() from widget-core to test-extras so it can be more widely used.

Fixes dojo/test-extras#76


agubler:
Widget-core can’t depend on test-extras unfortunately (as we cannot use test-extras to test in widget-core). So I think the function should be left here and just copied over to test-extras.

agubler:
It creates a horrible circular dependancy also (even if it just for a dev dep)

edhager:
On advice from @agubler, I'll close this without merging.

PR #804, opened 2017-12-14
802 - Animation Timing Properties cast correctly

### LINKED ISSUE #802: AnimationProperties interface needs changing to string enum
**Bug** 

current typing for AnimationProperties has `fill` and other fields defined as:

`fill?: 'none' | 'forwards' | 'backwards' | 'both' | 'auto';` 

But when trying to use it with `fill: 'forward'` it is throwing a type error.
There should all be changed to use string enums instead.

tomdye:
**Type:** feature

**Description:**

Moves the `AnimationProperties` used by the `WebAnimation` meta out of interfaces. Casts as the polyfilled type upon usage to avoid issues.

Breaking as it changes location of the interfaces.

Resolves #802

### DIFF
--- a/src/meta/WebAnimation.ts
+++ b/src/meta/WebAnimation.ts
@@ -1,8 +1,60 @@
 import 'web-animations-js/web-animations-next-lite.min';
 import { Base } from './Base';
-import { AnimationControls, AnimationProperties, AnimationInfo } from '../interfaces';
 import Map from '@dojo/shim/Map';
 import global from '@dojo/shim/global';
+
+/**
+ * Animation controls are used to control the web animation that has been applied
+ * to a vdom node.
+ */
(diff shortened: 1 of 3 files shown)


src/meta/WebAnimation.ts
```diff
@@ -24,7 +76,7 @@ export class WebAnimations extends Base {
 		const keyframeEffect = new KeyframeEffect(
 			node,
 			fx,
-			timing
+			timing as AnimationEffectTiming
```
agubler: I think you could have left it how it was just with this cast?

tomdye: yup, most likely, will fall back. thanks

tomdye:
@agubler will leave them out of interfaces.d.ts but will revert to the strings

PR #808, opened 2017-12-17
Focus meta provider

### LINKED ISSUE #729: Create meta for containing focus
**Enhancement**

From the widgets focus management issue (https://github.com/dojo/widgets/issues/309#issuecomment-339100993):

> **Provide meta information about whether focus is within a given node**
>
>This can be handled like a combination of Drag and Intersection meta. Similar to Drag, it will need a document-level event listener for the focus event, and like Intersection it can invalidate when focus moves into or out of a provided node.

I imagine the API would be something like:
`public get(key: string): boolean`

smhigley:
**Type:** feature

**Description:**

Addresses #729. Creates Focus meta that provides information about whether a node is currently focused or contains focus.

Updates sinon to ^4.1.3, along with syntax updates in any tests that required it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@types/jsdom": "2.0.*",
     "@types/ramda": "^0.25.5",
     "@types/selenium-webdriver": "^3.0.8",
-    "@types/sinon": "^4.1.2",
+    "@types/sinon": "~4.1.2",
     "@types/web-animations-js": "^2.2.5",
     "@types/yargs": "^8.0.2",
     "@webcomponents/custom-elements": "^v1.0.0-rc.3",
@@ -61,7 +61,7 @@
     "ramda": "^0.25.0",
     "selenium-webdriver": "3.6.0",
     "serve": "^6.4.1",
-    "sinon": "^4.1.3",
+    "sinon": "~4.1.3",
     "tslint": "5.2.0",
     "typescript": "~2.6.1"
   },
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@types/jsdom": "2.0.*",
     "@types/ramda": "^0.25.5",
     "@types/selenium-webdriver": "^3.0.8",
-    "@types/sinon": "^1.16.31",
+    "@types/sinon": "^4.1.2",
     "@types/web-animations-js": "^2.2.5",
     "@types/yargs": "^8.0.2",
     "@webcomponents/custom-elements": "^v1.0.0-rc.3",
@@ -61,7 +61,7 @@
     "ramda": "^0.25.0",
     "selenium-webdriver": "3.6.0",
     "serve": "^6.4.1",
-    "sinon": "^1.17.6",
+    "sinon": "^4.1.3",
     "tslint": "5.2.0",
     "typescript": "~2.6.1"
   },
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "grunt-dojo2": "latest",
     "grunt-tslint": "5.0.1",
     "husky": "0.14.3",
-    "intern": "~4.1.0",
+    "intern": "~4.1.5",
     "jsdom": "^9.5.0",
     "jstat": "^1.7.1",
     "lint-staged": "6.0.0",
--- a/src/meta/Base.ts
+++ b/src/meta/Base.ts
@@ -24,7 +24,7 @@ export class Base extends Destroyable implements WidgetMetaBase {
 		return this.nodeHandler.has(key);
 	}
 
-	protected getNode(key: string | number): Element | undefined {
+	public getNode(key: string | number): Element | undefined {
 		const stringKey = `${key}`;
 		const node = this.nodeHandler.get(stringKey);
 
--- a/src/meta/Base.ts
+++ b/src/meta/Base.ts
@@ -24,7 +24,7 @@ export class Base extends Destroyable implements WidgetMetaBase {
 		return this.nodeHandler.has(key);
 	}
 
-	public getNode(key: string | number): Element | undefined {
+	protected getNode(key: string | number): Element | undefined {
 		const stringKey = `${key}`;
 		const node = this.nodeHandler.get(stringKey);
 
--- a/src/meta/Focus.ts
+++ b/src/meta/Focus.ts
@@ -33,15 +33,10 @@ export class Focus extends Base {
 		};
 	}
 
-	public set(key: string | number) {
-		const node = this.getNode(key) as HTMLElement;
-		node && node.focus();
-	}
-
 	private _onFocus = () => {
 		this._activeElement = global.document.activeElement;
 		this.invalidate();
-	}
+	};
 
 	private _createListener() {
 		global.document.addEventListener('focusin', this._onFocus);
--- a/src/meta/Focus.ts
+++ b/src/meta/Focus.ts
@@ -48,11 +48,11 @@ export class Focus extends Base {
 
 	private _createListener() {
 		global.document.addEventListener('focusin', this._boundFocusHandler);
-		this.own(createHandle(this._removeListener));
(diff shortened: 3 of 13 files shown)


agubler:
Can you add back the updated `package-lock.json` please?

package.json
```diff
@@ -40,7 +40,7 @@
     "@types/jsdom": "2.0.*",
     "@types/ramda": "^0.25.5",
     "@types/selenium-webdriver": "^3.0.8",
-    "@types/sinon": "^1.16.31",
+    "@types/sinon": "^4.1.2",
```
agubler: Probably wise to pin this dependency :)

package.json
```diff
@@ -57,7 +57,7 @@
     "ramda": "^0.25.0",
     "selenium-webdriver": "3.6.0",
     "serve": "^6.4.1",
-    "sinon": "^1.17.6",
+    "sinon": "^4.1.3",
```
agubler: And this dependency!

src/meta/Focus.ts
```diff
@@ -0,0 +1,59 @@
...
+
+		return {
+			active: node === this._activeElement,
+			containsFocus: node.contains(this._activeElement)
+		};
+	}
+
+	public set(key: string | number, selector?: string) {
+		const node = this.getNode(key);
+		const target = node && selector ? node.querySelector(selector) : node;
```
agubler: Does this break encapsulation? Allowing a parent to focus on a node from a child widget?

agubler: Also you should be able to do `as HTMLElement` here rather than on usage.

src/meta/Focus.ts
```diff
@@ -0,0 +1,59 @@
...
+		target && (target as HTMLElement).focus();
+	}
+
+	private _onFocus() {
```
agubler: if you do 
```ts
private _onFocus = () => {
    // on focus stuff here
}
```

You shouldn't need to create the bound function.

src/meta/Focus.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export class Focus extends Base {
+	private _activeElement: Element;
+	private _boundFocusHandler = this._onFocus.bind(this);
+
+	public get(key: string | number): FocusResults {
+		const node = this.getNode(key);
+
+		if (!node) {
+			return assign({}, defaultResults);
```
agubler: Do we need to use `assign` from dojo/core here? Can we just do:
```ts
return { ...defaultResults };
```

src/meta/Focus.ts
```diff
@@ -0,0 +1,59 @@
...
+	containsFocus: boolean;
+}
+
+const defaultResults = {
+	active: false,
+	containsFocus: false
+};
+
+export class Focus extends Base {
+	private _activeElement: Element;
```
agubler: If another component uses `Focus` meta and sets focus, is the element set as `_activeElement` then incorrect? It looks like we set it once when it is not set, but if another meta sets focus, it doesn't get reset so the return from the `get` would be still return `true` for `node === this._activeElement`?

tests/unit/meta/Focus.ts
```diff
@@ -0,0 +1,163 @@
+const { registerSuite } = intern.getInterface('object');
```
agubler: For new tests can we use the `bdd` interfaces?

agubler:
Could you please add a section to the README for the usage?

src/meta/Focus.ts
smhigley: I think I just copied this pattern from other meta providers, but `{ ...defaultResults }` looks better

src/meta/Focus.ts
agubler: Right, gotcha.

src/meta/Focus.ts
smhigley: Having a child selector seemed like a good way to work around `DomWrapper` or wanting to call focus on a child widget, but it does break encapsulation. One possible solution might be to check focusability and send focus to the first focusable child found, which would allow a parent to send focus inside a child widget without knowing much about its DOM.

It’s not perfect, since the parent would still be assuming that something is focusable inside the child, and checking focusability is a bit fragile. I'm not sure what a better solution would look like, however.

agubler: Why does a widget need be able to focus something that it doesn't declare? I was kinda expecting the focus meta to simply return whether a node was currently focussed (or as you have done also indicate a node currently contains the focussed node) or be able to set focus on a node that it explicitly declares using `v()`.

smhigley: Some examples: The ComboBox needs to be able to focus on its TextInput widget, Select needs to focus on the Listbox root, and Calendar needs to focus on internal CalendarCell widgets.

Outside those examples, I imagine many widgets that control focus are occasionally going to want to send focus to a form field widget, or a Button widget.

agubler: All these things are widgets that should provide an API to focus and use the focus meta internally?

smhigley: I'm not sure what that API should look like. Right now we use a rather fragile pattern where the focusable widget has `callFocus` and `onFocusCalled` properties, and it's up to the parent to set `callFocus` back to false in the `onFocusCalled` callback. If the parent doesn't handle it correctly, you risk repeatedly stealing focus and completely breaking keyboard navigation. I hope we can provide a tool that allows widget authors to move focus more easily, and with fewer pitfalls.

The basic problem is calling focus is a strictly one-time thing, so it can't be controlled easily through properties. There's also the issue that controlling focus is a parent-level concern, but needs to be done on child DOM.

matt-gadd: it does seem like we need some concept of whether something is focusable at a widget level. that way we could at least signal the intent via the parent but it would still be the child widget that sets it's own focus.

src/meta/Base.ts
```diff
@@ -24,7 +24,7 @@ export class Base extends Destroyable implements WidgetMetaBase {
 		return this.nodeHandler.has(key);
 	}
 
-	protected getNode(key: string | number): Element | undefined {
+	public getNode(key: string | number): Element | undefined {
```
agubler: Why did you need to make this `public`?

smhigley: The `WebAnimation` meta stubs it in unit tests, and the updated version of sinon complained if it wasn't `public`

agubler: Ah that's a little annoying :(

smhigley:
@tomdye Focus stealing is actually more of an issue when using `focus` as a property in `render`. The way Meta currently handles it means that it should be easy to only call it once (e.g. in an event handler). If it were to be in `render`, it would be much more likely to steal if it were set to `true` on multiple renders.

The problem with having `focus` be a property is that moving focus anywhere is a single event, not an ongoing state. As soon as you try to control `focus` with properties you need to be really careful to not allow `focus` to be `true` for more than a single render. We do that now with `focus`/`onFocusCalled`, but it's fragile and the consequences for a widget user messing it up are pretty high.

src/meta/Base.ts
smhigley: yeah :/

tomdye:
@smhigley from my understanding, if using a property to control focus within the `render` function, in order to call `focus()` twice, you'd need to set the property from `true -> false -> true` due to our property diff approach. So I don't think that it is such a cause for concern.

agubler:
For now can you add the `test.skip` in each of the tests that may need to be skipped?

agubler:
@smhigley intern has been patched to support `skip` in the `beforeEach`, but there seems to be real CI failures now.

src/meta/Base.ts
tomdye: I think we need to raise an issue to address the `WebAnimation` tests rather than doing this here. Should probably have bumped `sinon` in it's own PR.

PR #819, opened 2018-01-04
Moved initPropertiesAndChildren before inserting into the DOM

### LINKED ISSUE #816: vdom.createDom creates intermediate DOM states
I am working with [A-Frame](https://aframe.io/) and Dojo 2 to render a scene in VR for a demo I've been working on called [Derpymon](https://github.com/devpaul/derpymon). I'm running into an issue with A-Frame's custom element throwing an exception due to missing required attributes (specifically A-Frame's [Asset Management System](https://aframe.io/docs/0.7.0/core/asset-management-system.html) has an element `a-asset-item` that expects a `src` attribute).

The problem is due to `vdom`'s `createDom()` function. When creating `v` elements it will [immediately attach them to the DOM before adding their attributes](https://github.com/dojo/widget-core/blob/master/src/vdom.ts#L780-L786). I was able to "fix" my issue by moving `initPropertiesAndChildren()` before attaching to the DOM. 

I'm not sure if this is a viable permanent solution or if there are reasons for attaching an element to the DOM before adding attributes; e.g. maybe some events are fired after attributes are added that may not be caught otherwise. React's provides a work-around for these cases by adding a special keyword `ref` that [allows access to the element](https://reactjs.org/docs/refs-and-the-dom.html). A-Frame's React wrapper [uses ref to ensure attributes are available](https://github.com/ngokevin/aframe-react/blob/master/src/index.js#L125), so another viable solution for Dojo may be to add a special Symbol that provides access to the element before adding it to the DOM.

As custom elements become more prevalent we'll definitely need a solution to this issue. Otherwise, we risk not being able to use custom elements with required attributes (or other similar expectations) with Dojo. FWIW, I've had a previous discussion with Maquette trying to come up with a solution in AFASSoftware/maquette#110.

Package Version: 0.2

**Code**
Here's a link to the problematic implementation of the custom-element:

https://github.com/devpaul/derpymon/blob/master/src/widgets/Assets.ts#L16-L18

**Expected behavior:**

Attributes should be attached to the element:

`<a-asset-item src="./path/to/model.obj">`

**Actual behavior:**

The element is initially rendered without attributes:

`<a-asset-item>`

devpaul:
Moved `initPropertiesAndChildren` in `vdom`'s `createDom` before the node is inserted into the DOM to prevent triggering observers and custom elements on an incomplete DOM. More details in the ticket.

Resolves #816

### DIFF
--- a/src/vdom.ts
+++ b/src/vdom.ts
@@ -744,7 +744,6 @@ function createDom(
 			} else {
 				domNode = dnode.domNode;
 			}
-			// add attributes and children so we have a fully formed node prior to attach
 			initPropertiesAndChildren(domNode! as Element, dnode, parentInstance, projectionOptions);
 			if (insertBefore !== undefined) {
 				parentVNode.domNode!.insertBefore(domNode, insertBefore);
--- a/src/vdom.ts
+++ b/src/vdom.ts
@@ -744,12 +744,13 @@ function createDom(
 			} else {
 				domNode = dnode.domNode;
 			}
+			// add attributes and children so we have a fully formed node prior to attach
+			initPropertiesAndChildren(domNode! as Element, dnode, parentInstance, projectionOptions);
 			if (insertBefore !== undefined) {
 				parentVNode.domNode!.insertBefore(domNode, insertBefore);
 			} else if (domNode!.parentNode !== parentVNode.domNode!) {
 				parentVNode.domNode!.appendChild(domNode);
 			}
-			initPropertiesAndChildren(domNode! as Element, dnode, parentInstance, projectionOptions);
 		}
 	}
 }
--- a/tests/unit/vdom.ts
+++ b/tests/unit/vdom.ts
@@ -1,6 +1,9 @@
+import Test from 'intern/lib/Test';
+
 const { afterEach, beforeEach, describe, it } = intern.getInterface('bdd');
 const { assert } = intern.getPlugin('chai');
 import { match, spy, stub, SinonStub } from 'sinon';
+import global from '@dojo/shim/global';
 import { createResolvers } from './../support/util';
 import sendEvent from './../support/sendEvent';
 
@@ -2706,6 +2709,40 @@ describe('vdom', () => {
 					projection.update(v('div', [v('span', { enterAnimation: 'fadeIn' })]));
 				});
 			});
+		});
+	});
+
+	describe('behavior', () => {
+		it('will append nodes with attributes already attached', function(this: Test) {
+			if (!global.window || !global.window.hasOwnProperty('MutationObserver')) {
+				this.skip('MutationObserver not present');
+			}
+
+			class Foo extends WidgetBase {
+				render() {
+					return [v('div', { attr: 'test' }), v('div', { attr: 'test' })];
+				}
+			}
+
+			const doc = global.window.document;
+			const log: MutationRecord[] = [];
+			const observer = new MutationObserver((mutations: MutationRecord[]) => {
+				log.push(...mutations);
+			});
+
+			const parent = doc.createElement('div');
+			observer.observe(parent, {
+				attributes: true,
+				childList: true,
+				subtree: true
+			});
+			const widget = new Foo();
+			dom.append(parent, widget.__render__() as VNode, widget);
+
+			const results = [...log, ...observer.takeRecords()];
+			observer.disconnect();
+			assert.isTrue(results.every((mutation) => mutation.type !== 'attributes'));
+			assert.lengthOf(results, 2);
 		});
 	});
 });
--- a/tests/unit/vdom.ts
+++ b/tests/unit/vdom.ts
@@ -2737,7 +2737,7 @@ describe('vdom', () => {
 				subtree: true
 			});
 			const widget = new Foo();
-			dom.append(parent, widget.__render__() as VNode, widget);
+			dom.append(parent, widget.__render__(), widget);
 
 			const results = [...log, ...observer.takeRecords()];
 			observer.disconnect();
--- a/tests/unit/vdom.ts
+++ b/tests/unit/vdom.ts
@@ -1,5 +1,3 @@
-import Test from 'intern/lib/Test';
-
 const { afterEach, beforeEach, describe, it } = intern.getInterface('bdd');
 const { assert } = intern.getPlugin('chai');
 import { match, spy, stub, SinonStub } from 'sinon';
@@ -2713,9 +2711,9 @@ describe('vdom', () => {
 	});
 
 	describe('behavior', () => {
-		it('will append nodes with attributes already attached', function(this: Test) {
+		it('will append nodes with attributes already attached', (test) => {
 			if (!global.window || !global.window.hasOwnProperty('MutationObserver')) {
-				this.skip('MutationObserver not present');
+				test.skip('MutationObserver not present');
 			}
 
 			class Foo extends WidgetBase {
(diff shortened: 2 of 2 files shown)


src/vdom.ts
```diff
@@ -744,12 +744,13 @@ function createDom(
 			} else {
 				domNode = dnode.domNode;
 			}
+			// add attributes and children so we have a fully formed node prior to attach
```
agubler: Not sure we need to add a comment for this

tests/unit/vdom.ts
```diff
@@ -2708,4 +2711,39 @@ describe('vdom', () => {
 			});
 		});
 	});
+
+	describe('behavior', () => {
+		it('will append nodes with attributes already attached', function(this: Test) {
```
agubler: I think this can use a param rather than `this` for skipping (and you shouldn't need to explicitly type `Test`), like:

```ts
it('will append nodes with attributes already attached',(test) => {
    if (!global.window || !global.window.hasOwnProperty('MutationObserver')) {
        test.skip('MutationObserver not present');
    }
    // ...
}
```

tests/unit/vdom.ts
```diff
@@ -2708,4 +2711,39 @@ describe('vdom', () => {
...
+				childList: true,
+				subtree: true
+			});
+			const widget = new Foo();
+			dom.append(parent, widget.__render__() as VNode, widget);
+
+			const results = [...log, ...observer.takeRecords()];
+			observer.disconnect();
+			assert.isTrue(results.every((mutation) => mutation.type !== 'attributes'));
+			console.log(results);
```
agubler: left a `console.log` in

tests/unit/vdom.ts
```diff
@@ -2708,4 +2711,39 @@ describe('vdom', () => {
...
+			const parent = doc.createElement('div');
+			observer.observe(parent, {
+				attributes: true,
+				childList: true,
+				subtree: true
+			});
+			const widget = new Foo();
+			dom.append(parent, widget.__render__() as VNode, widget);
+
+			const results = [...log, ...observer.takeRecords()];
```
agubler: Any reason you can't just do `log.push(...observer.takeRecords());` and use `log`?

tests/unit/vdom.ts
```diff
@@ -1,6 +1,9 @@
+import Test from 'intern/lib/Test';
```
agubler: As per the comment below, shouldn't need to import the `Test` type.

tests/unit/vdom.ts
```diff
@@ -2708,4 +2711,39 @@ describe('vdom', () => {
...
+				}
+			}
+
+			const doc = global.window.document;
+			const log: MutationRecord[] = [];
+			const observer = new MutationObserver((mutations: MutationRecord[]) => {
+				log.push(...mutations);
+			});
+
+			const parent = doc.createElement('div');
```
agubler: Everywhere else in the vdom tests we've just used `document` without needing the global shim

tests/unit/vdom.ts
```diff
@@ -2708,4 +2711,38 @@ describe('vdom', () => {
...
+			observer.observe(parent, {
+				attributes: true,
+				childList: true,
+				subtree: true
+			});
+			const widget = new Foo();
+			dom.append(parent, widget.__render__() as VNode, widget);
```
agubler: Don't need to `as VNode` here (I know there are other tests that still do)

tests/unit/vdom.ts
agubler: but if we are using it for the sniff then I guess no harm (although don't think we need to use `global` to do the `window` check either ¯\_(ツ)_/¯)

devpaul: personally I have zero love and trust for JSDOM :). I'd be happy to use bare global calls if there's any benefit.

agubler: Yeah there's no benefit so that's fine, we should probably be consistent through the all test in the future though.

src/vdom.ts
devpaul: It's probably worth a comment. Any place where order isn't apparent but required is worth mentioning.

agubler: Hmmmm. Seemed pretty obvious being that it is actually directly before the code that does the attachment, but okay.

matt-gadd: I don’t think this currently offers much value, if you feel the comment is needed can you expand upon it over “fully formed node” which doesn’t really mean anything.

tests/unit/vdom.ts
```diff
@@ -2708,4 +2709,38 @@ describe('vdom', () => {
 			});
 		});
 	});
+
+	describe('behavior', () => {
```
matt-gadd: This isn’t a great description, also is there any reason this year cannot be in the `create` or `properties` blocks elsewhere

tests/unit/vdom.ts
```diff
@@ -2708,4 +2709,38 @@ describe('vdom', () => {
...
+		it('will append nodes with attributes already attached', (test) => {
+			if (!global.window || !global.window.hasOwnProperty('MutationObserver')) {
+				test.skip('MutationObserver not present');
+			}
+
+			class Foo extends WidgetBase {
```
matt-gadd: Don’t think we need a widget to exercise this functionality do we? Any reason we can’t just use dom.create?

tests/unit/vdom.ts
```diff
@@ -2708,4 +2709,38 @@ describe('vdom', () => {
 			});
 		});
 	});
+
+	describe('behavior', () => {
+		it('will append nodes with attributes already attached', (test) => {
+			if (!global.window || !global.window.hasOwnProperty('MutationObserver')) {
+				test.skip('MutationObserver not present');
+			}
```
matt-gadd: It’s not great to skip a test due to the test using something that is potentially incompatible yet the actual code is. I’m not convinced a mutation observer is actually needed anyway, can you not just stub createElement and the associated things off it?

tests/unit/vdom.ts
```diff
@@ -2708,4 +2709,38 @@ describe('vdom', () => {
 			});
 		});
 	});
+
+	describe('behavior', () => {
+		it('will append nodes with attributes already attached', (test) => {
```
matt-gadd: This tests attributes, how about properties too?

src/vdom.ts
matt-gadd: To expand what I meant by not meaning anything, I mean as in describing why the ordering is important.

tests/unit/vdom.ts
devpaul: Yeah, I didn't know where to put it. Refactoring the test to use `dom.create()` will make it easier to house in that test suite.

tests/unit/vdom.ts
devpaul: The intention of the test is that `setProperties` was called before `attachNode`. I wanted to avoid describing internal functions in the test. Would it be more clear with another name?

src/vdom.ts
devpaul: There's a test that covers this now. If a future developer changes this they'll fail the test and can back-track through issues.

tests/unit/vdom.ts
```diff
@@ -71,6 +73,10 @@ describe('vdom', () => {
 	afterEach(() => {
 		consoleStub.restore();
 		resolvers.restore();
+		for (let spy of spys) {
```
matt-gadd: could use sinon.sandbox instead for this (then we wouldn't need the for and it's useful for other things), but not blocking this PR going in.

tests/unit/vdom.ts
```diff
@@ -1421,6 +1427,28 @@ describe('vdom', () => {
...
+					appendedHtml.push(node.outerHTML);
+					return appendChild(node);
+				});
+				return node;
+			});
+			spys.push(createElementStub);
+			const projection = dom.create(v('div', { 'data-attr': 'test' }), projectorStub);
+
+			assert.strictEqual(projection.domNode.innerHTML, expected);
+			assert.lengthOf(appendedHtml, 1);
```
matt-gadd: given this PR is about an ordering change, it would've been nice to have a slightly more expressive assertion for that. ie sinon keeps track of call orders across spies/stubs so we could've done something like:
```ts
assert.isTrue(appendChildSpy.calledAfter(setAttributeSpy));
```
anyway this isn't blocking us taking this 👍.

tests/unit/vdom.ts
devpaul: That would be a good refactor for this test suite 👍

PR #822, opened 2018-01-08
Set instance as dirty over calling invalidate

### LINKED ISSUE #821: The projector continuously re-renders when using a `Container`
**Bug**

The `Container` currently overrides `__setProperties__` and always calls `this.invalidate()` to ensure that the container renders. This used to work as the widget was not in `RenderState.IDLE` but the `RenderState` is reset at the end of `__setProperties__` this call is causing an invalidation up the widget tree every render.

agubler:
**Type:** bug

**Description:**

Set `instanceData.`dirty` to `true` over using `this.invalidate()` in the `Container`s `__setProperties__` override.

Resolves #821


agubler:
closing, favouring #824

PR #827, opened 2018-01-16
Move web-animations typings as dep and remove intern config for polyfill

tomdye:
**Type:** feature

* [ ] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

Moves the web animation types from a dev dependency to a dependency in package.json and removes the web animations config from intern.json as this will be provided by shim.

Relies on: https://github.com/dojo/shim/pull/128

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,8 +41,7 @@
     "@types/jsdom": "2.0.*",
     "@types/ramda": "^0.25.5",
     "@types/selenium-webdriver": "^3.0.8",
-    "@types/sinon": "~4.1.2",
-    "@types/web-animations-js": "^2.2.5",
+    "@types/sinon": "^1.16.31",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -69,7 +68,8 @@
     "intersection-observer": "^0.4.2",
     "pepjs": "^0.4.2",
     "tslib": "~1.8.1",
-    "web-animations-js": "^2.3.1"
+    "web-animations-js": "^2.3.1",
```
agubler: let's pin these (ideally to matching versions), could you pin the other dependencies too? Don't want any nasty surprises :)

tomdye: Done

PR #831, opened 2018-01-17
Remove replace from the vdom API

agubler:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)
* [x] Unit or Functional tests are included in the PR

**Description:**

Removes `replace` from the `vdom` and `projector` APIs.

Resolves #741

PR #838, opened 2018-01-22
Sub Tree Rendering

### LINKED ISSUE #741: Remove `replace` from `vdom` and `projector`
**Enhancement**

Replacing an existing DOM node is rarely used and not something that I think we should support from widget-core.

Proposal is to remove the capability leaving `merge`, `append` and `create`.

### LINKED ISSUE #767: Sub tree rendering
**Enhancement**

### Existing system

The existing invalidation and rendering system behaves like the following:

image

Key:
**Red**: Will render
**Orange**: Will potentially render (`setProperties` will be called)
**Arrows**: Invalidation path
**Black dot**: Source widget that invalidated

As you can see, when `Widget J` invalidates, we go all the way up the tree to the root invalidating all widgets along the way. We then come back down the tree, definitely re-rendering the branch the invalidated widget was in (in this case `Widget A`, `Widget C`, `Widget G` and `Widget J`). As well as this, the Widgets marked in orange at least have their `setProperties` ran on them also. This can be quite/very costly in tree's that have any depth, not only due to the `setProperties` and `render` calls, but ultimately the actual dom diffing below.

### Proposed system

The proposed system would be a switch to sub tree rendering. This would mean when a widget invalidates only itself is actually guaranteed to be re-rendered, there is no invalidation path up to the top widget and back down the branch, we simply render from the widget that invalidated.

image

Key:
**Red**: Will render
**Orange**: Will potentially render (`setProperties` will be called)
**Black dot**: Source widget that invalidated

The above is a good best case scenario compared to the existing system. 

In terms of this working as a system we have to take a few extra things into account: 

* If there are multiple widget invalidations across the tree for one scheduled render, we can now no longer bank on it being picked up in one whole tree render like the existing system did. To deal with this, we need to queue up the widgets that have invalidated, and render those subtrees also. 

* If there are multiple widget invalidates in the same branch then one is guaranteed to include the other (ie one is higher in the tree, so will render the subtree anyway), if the order means the lower one in the tree is first, then we potentially end up having to render it twice. We can solve this by sorting the render queue by depth, which means the higher up in the tree is guaranteed to be rendered first.

Overall in apps that have any depth of widgets this should offer some serious real world performance benefits, as well as making a more logical pipeline of what happens when a widget is invalidated. This will also enable is in the future to more smartly optimise and schedule rendering on a widget level.

agubler:
**Type:** feature

**Description:**

Implements the concept of "Sub Tree Rendering" as described in the associated issue #767. 

The current invalidation mechanism propagates up through the entire widget tree to the top level projector and signals a re-render down this tree. This can be inefficient for scenarios that do not require any changes to their parent widgets, where a single widget (and possibly children) only require re-rendering.

This change removes the existing propagation model (eventing) to a model that registers widgets that have signalled that they require re-rendering onto a queue that get reconciled on the animation frame ordered by the depth they exist in the widget tree.

Tested in example applications: `todo-mvc`, `todo-mvc-kitchensink` and `realworld`

**Note:** These changes will require significant changes within the `@dojo/text-extras` module.
**Double Note:** Will impact the current PoC for `@dojo/diagnostics`

Resolves #767 
Resolves #741  

Pending specific unit tests for the subtree rendering (all existing tests are passing)


src/mixins/Projector.ts
```diff
@@ -1,15 +1,14 @@
...
-import 'pepjs';
 import cssTransitions from '../animations/cssTransitions';
 import { Constructor, DNode, Projection, ProjectionOptions } from './../interfaces';
 import { WidgetBase } from './../WidgetBase';
 import { afterRender } from './../decorators/afterRender';
 import { v } from './../d';
 import { Registry } from './../Registry';
-import { dom, widgetInstanceMap } from './../vdom';
+import { dom } from './../vdom';
+import 'pepjs';
```
tomdye: I believe this import can be removed as it's handled in [shim](https://github.com/dojo/shim/blob/f2594e2d6f91865464d8a10ba80b76e1aa4e16f1/src/browser.ts#L1)

src/vdom.ts
```diff
@@ -543,7 +578,7 @@ function updateChildren(
 	const oldChildrenLength = oldChildren.length;
 	const newChildrenLength = newChildren.length;
 	const transitions = projectionOptions.transitions!;
-
+	projectionOptions = { ...projectionOptions, ...{ depth: projectionOptions.depth + 1 } };
```
tomdye: No need for the second spread here or on line 671.
`projectionOptions = { ...projectionOptions, depth: projectionOptions.depth + 1 };`

agubler:
Includes removing the replace API from the projector #831

PR #848, opened 2018-01-31
Support `focus` on `VNodeProperties`

### LINKED ISSUE #847: Support focus property on VNodeProperties
**Enhancement**

Add support in vdom for a focus property that accepts a `boolean` value and diffs like normal properties and calls focus (when value is `true` and was `false`) on the node in the next request animation frame.

Additionally support a function that returns a boolean to indicate if focus is needed for a node, this will not be diff'd as a normal property, instead it will run the function every render and always call focus if the return is `true`.

This support should enable future enhancements wrt to managing focus.

agubler:
**Type:** feature

**Description:**

Adds support in `vdom` for a focus property that calls focus on the next rAF if the property is `true` and was `false` in the previous render or the function returns `true`

Resolves #847

### DIFF
--- a/src/interfaces.d.ts
+++ b/src/interfaces.d.ts
@@ -89,6 +89,8 @@ export type SupportedClassName = string | null | undefined;
 
 export type DeferredVirtualProperties = (inserted: boolean) => VNodeProperties;
 
+export type FocusFunction = () => boolean;
+
 export interface VNodeProperties {
 	/**
 	 * The animation to perform when this node is added to an already existing parent.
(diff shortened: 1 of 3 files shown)


schontz:
Is this the correct usage?

```typescript
v('div', { focus: true }, [...]);
```
And does it work on widgets as well?

agubler:
@schontz That is the correct usage and it doesn't work for widgets.

schontz:
Thank you. FYI the [sandbox demo](https://codesandbox.io/s/ox5y97vqz5) is broken. Change line 4 of `Demo.ts` to:
```typescript
import Focus from '@dojo/widget-core/mixins/Focus';
```

agubler:
@schontz I had done that, but it didn't save for some reason 😞 Thanks!

PR #854, opened 2018-02-02
Transitioning to @dojo/scripts

rorticus:
What's better than widget-core? A widget-core without GRUNT!

This PR removes `grunt-dojo2` and adds npm scripts.

- `test` - run intern unit tests in node
- `test:local` - run intern tests in browser
- `test:browserstack` - run tests in browserstack
- `lint` - run tslint
- `build` - build project as UMD, ESM, and a release distribution (checkout `dist/` directory)
- `watch` - Run the build in watch mode. Watch mode will rebuild all three directories mentioned above when a typescript file changes
- `release` - Run the publish checks and the release command. Note that it expects more arguments, so `npm run release -- --version=2.1.1 --next=2.1.2-pre`


tslint.json
```diff
@@ -1,60 +1,6 @@
...
-			"variable-declaration": "nospace"
-		}, {
-			"call-signature": "onespace",
-			"index-signature": "onespace",
-			"parameter": "onespace",
-			"property-declaration": "onespace",
-			"variable-declaration": "onespace"
-		} ],
-		"variable-name": [ true, "check-format", "allow-leading-underscore", "ban-keywords", "allow-pascal-case" ]
+		"whitespace": [ true, "check-branch", "check-decl", "check-operator", "check-module", "check-separator", "check-type", "check-typecast" ]
```
matt-gadd: how come we need an override for this?

tsconfig.json
```diff
@@ -1,29 +1,11 @@
 {
-	"version": "2.1.5",
+	"extends": "./node_modules/@dojo/scripts/tsconfig/umd.json",
 	"compilerOptions": {
 		"declaration": false,
```
matt-gadd: whats the default for this?

tsconfig.esm.json
```diff
@@ -0,0 +1,17 @@
+{
+	"extends": "./node_modules/@dojo/scripts/tsconfig/esm.json",
+	"compilerOptions": {
+		"declaration": false,
+		"emitDecoratorMetadata": true,
+		"jsx": "react",
+		"jsxFactory": "tsx",
+		"noUnusedLocals": true,
```
matt-gadd: this should be common right?

tsconfig.esm.json
```diff
@@ -0,0 +1,17 @@
+{
+	"extends": "./node_modules/@dojo/scripts/tsconfig/esm.json",
+	"compilerOptions": {
```
matt-gadd: seems like the only options that would need adding here are the jsx ones and the types?

tsconfig.json
rorticus: It defaults to true, probably because it is based off the shim/core tsconfig.json which both are set to true.

tsconfig.esm.json
rorticus: I added `noUnusedLocals` to the base config and removed the stuff not specific to widget-core

rorticus:
This is rebased on master. The build is failing, but that's because @types/node updated and broke things... it's failing on master too at the moment.

Also note that the benchmarks are broken, but they too are broken on master.

JamesMilnerUK:
@rorticus benchmarks should be passing now

agubler:
One day we'll get this in! What is the current status (apart from the conflicted files?)

PR #856, opened 2018-02-02
Allow generic for properties on custom element to not extend WidgetProperties

### LINKED ISSUE #855: customElement generic for property type should not need to extend WidgetProperties
**Bug**

Within widget-core we apply `WidgetProperties` implicitly so that end consumers do not have to explicitly extend the interface. We should apply this to the properties generic that the `customElement` decorator requires and also default to any object.

agubler:
**Type:** bug

**Description:**

Update the types so that `WidgetProperties` is implicitly added to the allowed keys when using the custom element decorator.

Resolves #855

### DIFF
--- a/tests/unit/vdom.ts
+++ b/tests/unit/vdom.ts
@@ -3029,7 +3029,6 @@ describe('vdom', () => {
 
 		it('Should focus if function for focus returns true', () => {
 			const shouldFocus = () => {
-				console.log('here2');
 				return true;
 			};
 			const widget = getWidget(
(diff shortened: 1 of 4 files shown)


tests/unit/vdom.ts
```diff
@@ -3029,7 +3029,6 @@ describe('vdom', () => {
 
 		it('Should focus if function for focus returns true', () => {
 			const shouldFocus = () => {
-				console.log('here2');
```
agubler: Ooops left this in from a previous pull request, so removing it :)

PR #858, opened 2018-02-07
Allow consumer to break from the decorate function and support shallow

### LINKED ISSUE #685: The d/decorate function should allow choosing between deep/shallow decoration
**Enhancement**

The `d/decorate` function:

https://github.com/dojo/widget-core/blob/de035a8a6a182aa00d9cbc79dffb916e02c7d5d8/src/d.ts#L48-L65

Should allow the user to choose between deep or shallow decoration, as there are use cases where only the direct children of a DNode should be decorated, but currently is always goes deep.

### LINKED ISSUE #846: I18n mixin doesn't add the right properties if the top node is not a VNode
**Bug**

Potentially the i18n mixin should add the i18n properties to the first found `VNode` in the returned render tree, instead of only considering the top `VNode`.

Also consider what should happen if a widget returns an array of `DNode`s at the top level?

agubler:
**Type:** feature

**Description:**

* Support for a `breaker` function that can be called in a modifier to drain the queued nodes and exit decoration.
* Supports a `shallow` flag passed in the `DecorateOptions` (new API) to indicate that the `predicate` and `modifier` should only be run against the top level node or nodes.
* Backwards compatible with the existing API, however the existing API does not support the new `shallow` functionality.

Utilises the breaker for `i18n` to break after adding the `i18n` properties to the first `VNode` found.

Resolves #685 
Resolves #846

### DIFF
--- a/src/d.ts
+++ b/src/d.ts
@@ -10,6 +10,7 @@ import {
 	WidgetBaseInterface,
 	WNode
 } from './interfaces';
+import { RenderResult } from './vdom';
 
 /**
  * The symbol identifier for a WNode type
@@ -45,9 +46,13 @@ export interface Modifier<T extends DNode> {
 	(dNode: T, breaker: DecorateBreak): void;
 }
 
+export interface Predicate<T extends DNode> {
+	(dNode: DNode): dNode is T;
+}
+
 export interface DecorateOptions<T extends DNode> {
 	modifier: Modifier<T>;
-	predicate?: (dNode: DNode) => dNode is T;
+	predicate?: Predicate<T>;
 	shallow?: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


src/d.ts
```diff
@@ -37,41 +38,82 @@ export function isVNode(child: DNode): child is VNode {
...
  * Generic decorate function for DNodes. The nodes are modified in place based on the provided predicate
  * and modifier functions.
  *
  * The children of each node are flattened and added to the array for decoration.
  *
- * If no predicate is supplied then the modifier will be executed on all nodes.
+ * If no predicate is supplied then the modifier will be executed on all nodes. A `breaker` function is passed to the
+ * modifier which will drain the nodes array and exit the decoration.
+ *
+ * When the `shallow` options is set to `true` the only the top node or nodes will be decorated (only supported using
```
tomdye: extra `the` in this comment between `'true'` and `only the top`

PR #863, opened 2018-02-10
Add `dom` pragma to create VNodes using a DOM node.

### LINKED ISSUE #862: Support for existing dom nodes without using DomWrapper
**Enhancement**

Currently to use an existing dom node (not created by widget-core) in the widgeting system you need to use the `DomWrapper` widget. This is a widget that under the hood overrides `__render__` and decorates the dom node on to a `VNode` which isn't ideal as our double underscored functions aren't really meant to be overridden. As a consumer the end result is also not ideal, as the node they get back from a `DomWrapper` is a `WNode` rather than a `VNode`.

Given we now have control of the vdom system, I think we can now drop the `DomWrapper`, and instead provide a `dom()` pragma that is like `v()` but can take an existing dom node. We should also probably allow control over how the element is diffed also, because an existing element may already have properties/attributes on it, or may gain properties/attributes in the future from a different source.

agubler:
**Type:** feature

**Description:**

Retires the `DomWrapper` in favour of a new `dom` pragma that creates a `VNode` for the DOM node passed with explicit `props` and `attrs` and an optional `diffType` argument that describes how the `props` and `attrs` should be diff'd by the vdom system.

The diff types available are `vdom`, `none` and `dom`:

* `vdom`: The default diff'ing algorithm that is uses the previous `VNode` to determine what properties/attributes need to be updated on the DOM node.
* `none`: Will always compare the `properties` and `attributes` against an empty object and therefore they will always be set when processed through the vdom system. This is the default for the `dom` pragma.
* `dom`: Will compare the `VNode` `properties` and `attributes` against the values on the DOM node.

```ts
import { dom } from '@dojo/widget-core/d';

const node = document.createElement('div');
const vnode = dom({ 
    node, 
    // `foo` will be added as property even though it is a string type
    props: { foo: 'bar', bar: 1 },
    // baz will be added as an attribute
    attrs: { baz: 'bar' },
    diffType: 'vdom' | 'none' | 'dom' 
})
```

Resolves #862


agubler:
Need to add section in the README for the new `dom` function

README.md
```diff
@@ -1170,6 +1171,22 @@ class IsTallMeta extends MetaBase {
 }
 ```
 
+### Inserting DOM nodes into the VDom Tree
+
+The `dom()` function is used to wrap DOM that is created outside of Dojo 2, this is the only mechanism to integrate foreign DOM nodes into the virtual DOM system.
```
dylans: Dojo 2, this -> Dojo 2. This

src/customElements.ts
```diff
@@ -326,7 +318,7 @@ export function initializeElement(element: CustomElement) {
 			if (childrenType === ChildrenType.DOJO) {
 				children.push(w(DomToWidgetWrapper(childNode), properties));
 			} else {
-				children.push(w(DomWrapper(childNode), properties));
+				children.push(dom({ node: childNode, props: properties }));
```
agubler: This means that only properties and no attributes or events will be added to custom element children...

agubler: Is only a single key property anyway...

PR #873, opened 2018-02-20
Remove web-animations polyfill dependency and import

### LINKED ISSUE #770: WebAnimations should be moved to shim
**Enhancement**

The shim/polyfill for WebAnimations should be migrated to `@dojo/core/shim` in the same fashion as PEP is provided.

tomdye:
**Type:** bug / feature

* [ ] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

Dependent on `@dojo/shim` changes

Resolves #770 

Shim pr: https://github.com/dojo/shim/pull/132

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,8 +65,6 @@
     "typescript": "~2.6.1"
   },
   "dependencies": {
-    "intersection-observer": "0.4.3",
-    "pepjs": "0.4.3",
     "tslib": "~1.8.1",
     "@types/web-animations-js": "2.2.5"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -68,7 +68,6 @@
     "intersection-observer": "0.4.3",
     "pepjs": "0.4.3",
     "tslib": "~1.8.1",
-    "web-animations-js": "2.3.1",
```
agubler: Should this still be a dependency? intersection and pepjs are also handled by `browser.ts` from shim but they are still dependencies or should they be removed from here? Noticed that the versions are pinned here but in shim they are using `^`

PR #874, opened 2018-02-20
Minor readme grammar cleanup [ci skip]

dylans:
grammar cleanup

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ You can also use the [dojo cli](https://github.com/dojo/cli) to create a complet
 
 Dojo 2 applications use the Virtual DOM (vdom) paradigm to represent what should be shown on the view. These vdom nodes are plain JavaScript objects that are more efficient to create from a performance perspective than browser DOM elements. Dojo 2 uses these vdom elements to synchronize and update the browser DOM so that the application shows the expected view.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,7 @@ You can also use the [dojo cli](https://github.com/dojo/cli) to create a complet
 
 Dojo 2 applications use the Virtual DOM (vdom) paradigm to represent what should be shown on the view. These vdom nodes are plain JavaScript objects that are more efficient to create from a performance perspective than browser DOM elements. Dojo 2 uses these vdom elements to synchronize and update the browser DOM so that the application shows the expected view.
 
-There are two types of vdom within Dojo 2, the first are pure representations of DOM elements and are the fundamental building blocks of all Dojo 2 applications. These are called `VNode`s and are created using the `v()` function available from the `@dojo/widget-core/d` module.
+There are two types of vdom within Dojo 2, the first is pure representations of DOM elements and are the fundamental building blocks of all Dojo 2 applications. These are called `VNode`s and are created using the `v()` function available from the `@dojo/widget-core/d` module.
```
tomdye: I don't think this change reads well if I'm honest. How about, `the first is a pure`, there's another `are` later in the sentence also that no longer makes sense with this `is`.

dylans: Yeah, I think this sentence is actually a pretty poor mix of singular and plural tense and should probably be rewritten. Because grammatically is actually should be is (as it is one of two types), but it sounds awkward.

PR #896, opened 2018-03-16
Default base registry and enable defining/getting injectors on base registry

### LINKED ISSUE #895: Create base registry by default and only define injectors against the base registry
**Enhancement**

It would be useful to be able to set an injector on the base registry within a widget. To support the functionality we'd create a base registry by default for the projector and set it on the core properties such that it will be passed down to the widget tree. The default registry will be able to be overridden by passing a registry as a property on the projector as normal.

The `RegistryHandler` will only allow setting and getting injectors on the base registry as setting an injector on a locally scoped registry isn't useful.

agubler:
**Type:** feature

**Description:**

Create a default base registry and enhance registry functionality to allow injectors to be set against the base registry.

Resolves #895

### DIFF
--- a/src/RegistryHandler.ts
+++ b/src/RegistryHandler.ts
@@ -41,7 +41,7 @@ export class RegistryHandler extends Evented<RegistryHandlerEventMap> {
 	}
 
 	public defineInjector(label: RegistryLabel, injector: Injector): void {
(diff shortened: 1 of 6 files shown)


agubler:
This is quite right to support the requirements. Will need to revisit this in the future.

PR #899, opened 2018-03-22
TS2.7 forwards compatibility

agubler:
**Type:** feature

**Description:**

Forwards compatibility for TS2.7 and upgrade @dojo/core.

Requires release and update of the dojo/i18n version

### DIFF
--- a/src/mixins/Projector.ts
+++ b/src/mixins/Projector.ts
@@ -109,16 +109,16 @@ export interface ProjectorMixin<P> {
 }
 
 export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T): T & Constructor<ProjectorMixin<P>> {
-	class Projector extends Base {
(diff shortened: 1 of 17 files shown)


src/mixins/Projector.ts
```diff
@@ -109,16 +109,16 @@ export interface ProjectorMixin<P> {
 }
 
 export function ProjectorMixin<P, T extends Constructor<WidgetBase<P>>>(Base: T): T & Constructor<ProjectorMixin<P>> {
-	class Projector extends Base {
+	abstract class Projector extends Base {
 		public projectorState: ProjectorAttachState;
-		public properties: Readonly<P> & Readonly<ProjectorProperties>;
 
-		private _root: Element;
+		private _root: Element = document.body;
```
dylans: At some point we should move things like this to use a reference to global, but it's consistent with what we're doing elsewhere.

PR #902, opened 2018-03-23
Allow widget i18n bundles to be overridden

### LINKED ISSUE #901: User-provided i18n bundles
**Enhancement** 

We need to determine a workflow to apply user provided resource bundles for our widgets. One possible option is to provide an additional `overrideBundle` property that, when specified, will be used over the default bundle.

mwistrand:
**Type:** feature

**Description:**

Add an `overrideBundle` property to the `I18n` mixin properties allowing the base bundle to be overridden, or specific bundles to be overridden in the event that a widget derives its messages from multiple bundles. `overrideBundle` can a complete bundle, an object with a `locales` loader object that will be merged with the widget's default bundle, or a `Map` of default bundles to override values.

Resolves #901

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,7 +539,7 @@ In the above example, the tabPanel will receive its original `root` class in add
 
 ### Internationalization
 
-Widgets can be internationalized by adding the `I18nMixin` mixin from `@dojo/widget-core/mixins/I18n`. [Message bundles](https://github.com/dojo/i18n) are localized by passing them to `localizeBundle`. Note that with this pattern it is possible for a widget to obtain its messages from multiple bundles; however, we recommend limiting widgets to a single bundle whenever possible.
+Widgets can be internationalized by adding the `I18nMixin` mixin from `@dojo/widget-core/mixins/I18n`. [Message bundles](https://github.com/dojo/i18n) are localized by passing them to `localizeBundle`. Note that with this pattern it is possible for a widget to obtain its messages from multiple bundles; however, we strongly recommend limiting widgets to a single bundle whenever possible.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -582,6 +582,36 @@ class I18nWidget extends MyWidgetBase<I18nWidgetProperties> {
...
+const overrideMapForWidgetB = new Map<Bundle<any>, Bundle<any>>();
+map.set(widgetB1, overrideBundleForWidgetB);
+
+export class MyWidget extends WidgetBase {
+	protected render() {
+		return [
+			w(WidgetA, {
+				i18nBundle: overrideBundleForWidgetA
+			}),
+			w(WidgetB) {
```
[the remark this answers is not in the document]
mwistrand: Ah, thanks!

tests/unit/mixins/I18n.ts
```diff
@@ -119,6 +135,62 @@ registerSuite('mixins/I18nMixin', {
 				});
 			}
 		},
+		'.localizeBundle() with an override': {
+			'Uses the `i18nBundle` property'(this: any) {
+				const dfd = this.async();
```
agubler: How come we are using `this.async` here and not dealing with promises as with other async tests in widget-core? It would be super if we didn't need to use a `setTimeout` also.

mwistrand: This was originally needed to workaround the fact that `localizeBundle` asynchronously loads locale-specific messages but synchronously returns either the default messages or the cached messages. However, the underlying functionality has changed between commits, so I should be able to just use `@dojo/i18n/i18n.default()` directly.

PR #910, opened 2018-04-16
Functional Factory Injector

### LINKED ISSUE #911: Change Registry to accept a functional Injector Factory over the existing Injector Instance
**Enhancement**

Currently to define an injector within a `Registry` a consumer is required to create an `Injector` instance and if any custom behaviour is required they need to create a custom `Injector` class by extending the base `Injector`. The `Injector` was only designed to be a wrapper that connects consumers payloads to connected widgets in the tree, but this has proven to be confusing and create issues with developer ergonomics and the uses that it inadvertently promotes:

* It requires consumers to deal with `Evented.emit()`, unlike everywhere else in Dojo 2, where we tend to provide abstractions on top of listening and emitting events.
* It was awkward to just provide a static payload that never needed to be reset.
* Encourages consumers to extend the `Injector` class and make it stateful which was never the intention.

This awkward and restrictive pattern, could be improved by changing the registry to accept a factory that returns an "injector" function instead of `Injector` class instance. This is a more flexible approach and enables more powerful uses of the registry without restricting consumers to be required to work with an Injector class.

agubler:
**Type:** feature

* [ ] Any required updates included in the readme

**Description:**

It has been identified that the ergonomics around using an `Injector` class as the base type when defining an injector in a registry can be awkward. 

* It requires consumers to deal with `Evented.emit()`, unlike everywhere else in Dojo 2, where we tend to provide abstractions on top of listening and emitting events.
* Additionally it was awkward to just provide a static payload that never needed to be reset.
* It encouraged consumers to extend the `Injector` class and make it stateful which was never the intention.

This change moves to accepting a functional factory that gets provided an invalidator and returns a function that when called will return the intended payload to be injected into the registered widgets.

```ts
registry.defineInjector('state', (invalidator: () => void) => {
    const applicationContext = new ApplicationContext(invalidator);
    return () => applicationContext;
});
```

If the payload never changes then the `invalidator` function can just be ignored.

```ts
registry.defineInjector('state', () => {
    return () => ({ foo: 'bar' });
});
```

Resolves #911

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,7 +854,7 @@ class MyClass extends WidgetBase {
 
 ### Containers & Injectors
 
-There is built-in support for side-loading/injecting values into sections of the widget tree and mapping them to a widget's properties. This is achieved by registering an injector factory with a `registry` and setting the registry as a property on the application's `projector` to ensure that is available to your application.
+There is built-in support for side-loading/injecting values into sections of the widget tree and mapping them to a widget's properties. This is achieved by registering an injector factory with a `registry` and setting the registry as a property on the application's `projector` to ensure the registry instance is available to your application.
 
 Create a factory function for a function that returns the required `payload`.
 
@@ -864,12 +864,12 @@ registry.defineInjector('my-injector', () => {
 });
 ```
 
-The injector factory gets passed an `invalidator` function that can be called when something has changed that requires connected widgets to `invalidate`.
+The injector factory gets passed an `invalidator` function that can get called when something has changed that requires connected widgets to `invalidate`.
 
 ```ts
 registry.defineInjector('my-injector', (invalidator) => {
-    // This could be a store, but for this example is simply an instance
-    // that accepts the injector and calls it when any of it's internal
+    // This could be a store, but for this example it is simply an instance
+    // that accepts the `invalidator` and calls it when any of its internal
     // state has changed.
     const appContext = new AppContext(invalidator);
     return () => appContext;
--- a/README.md
+++ b/README.md
@@ -688,11 +688,14 @@ A main registry can be provided to the `projector`, which will be automatically 
 import { Registry } from '@dojo/widget-core/Registry';
 
 import { MyWidget } from './MyWidget';
-import { MyInjector } from './MyInjector';
+import { MyAppContext } from './MyAppContext';
 
 const registry = new Registry();
 registry.define('my-widget', MyWidget);
-registry.defineInjector('my-injector', MyInjector);
+registry.defineInjector('my-injector', (invalidator) => {
+	const appContext = new MyAppContext(invalidator);
+	return () => appContext;
+});
 // ... Mixin and create Projector ...
 
 projector.setProperties({ registry });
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -851,18 +854,32 @@ class MyClass extends WidgetBase {
 
 ### Containers & Injectors
 
-There is built-in support for side-loading/injecting values into sections of the widget tree and mapping them to a widget's properties. This is achieved by registering a `@dojo/widget-core/Injector` instance against a `registry` that is available to your application (i.e. set on the projector instance, `projector.setProperties({ registry })`).
+There is built-in support for side-loading/injecting values into sections of the widget tree and mapping them to a widget's properties. This is achieved by registering an injector factory with a `registry` and setting the registry as a property on the application's `projector` to ensure that is available to your application.
```
dylans: "to ensure that is available"   please be more specific than "that is"

README.md
```diff
@@ -851,18 +854,32 @@ class MyClass extends WidgetBase {
...
+Create a factory function for a function that returns the required `payload`.
+
+```ts
+registry.defineInjector('my-injector', () => {
+    return () => ({ foo: 'bar' });
+});
+```
 
-Create an `Injector` instance and pass the `payload` that needs to be injected to the constructor:
+The injector factory gets passed an `invalidator` function that can be called when something has changed that requires connected widgets to `invalidate`.
```
dylans: can be called -> can get called

README.md
```diff
@@ -851,18 +854,32 @@ class MyClass extends WidgetBase {
...
 
 ```ts
-const injector = new Injector({ foo: 'baz' });
-registry.defineInjector('my-injector', injector);
+registry.defineInjector('my-injector', (invalidator) => {
+    // This could be a store, but for this example is simply an instance
+    // that accepts the injector and calls it when any of it's internal
```
dylans: it's -> its

src/Registry.ts
```diff
@@ -152,7 +157,7 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
 		}
 	}
 
-	public defineInjector(label: RegistryLabel, item: Injector): void {
+	public defineInjector(label: RegistryLabel, item: InjectorFactory): void {
```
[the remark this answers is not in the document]
agubler: I can defo change that

agubler:
@maier49 Think i've addressed you're feedback 👍

src/Registry.ts
```diff
@@ -201,12 +213,12 @@ export class Registry extends Evented<{}, RegistryLabel, RegistryEventObject> im
...
 	}
 
-	public getInjector<T extends Injector>(label: RegistryLabel): T | null {
+	public getInjector<T>(label: RegistryLabel): InjectorItem<T> | null {
 		if (!this._injectorRegistry || !this.hasInjector(label)) {
 			return null;
 		}
 
-		return this._injectorRegistry.get(label) as T;
+		return this._injectorRegistry.get(label)!;
```
devpaul: In the return were `InjectorItem<T> | undefined` wouldn't we be able to avoid the `!`?

agubler: We already check whether the item exists, so I know at this point it is not undefined and I wanted to avoid changing the API any more than required.

agubler: @devpaul ^

src/decorators/inject.ts
```diff
@@ -44,21 +43,22 @@ export interface InjectConfig {
...
-			if (injector) {
+			const injectorItem = this.registry.getInjector(name);
+			if (injectorItem) {
+				const { injector, invalidator } = injectorItem;
 				const registeredInjectors = registeredInjectorsMap.get(this) || [];
 				if (registeredInjectors.length === 0) {
 					registeredInjectorsMap.set(this, registeredInjectors);
 				}
-				if (registeredInjectors.indexOf(injector) === -1) {
+				if (registeredInjectors.indexOf(injectorItem) === -1) {
```
devpaul: Can we use a `WeakSet<injectorItem>` here?

PR #927, opened 2018-05-21
Fixes Svg(path, rect, ellipse, line) is not draggable

SagarKharche:
Bug:- SVG element is not able to drag in IE 11.

Fixed the svg dragging issue in IE11. IE 11 does not support parentElement for SVGElement, so in case of parentElement undefined now checking for parentNode.

### DIFF
--- a/src/meta/Drag.ts
+++ b/src/meta/Drag.ts
@@ -153,9 +153,12 @@ class DragController {
 		if (this._nodeMap.has(target)) {
 			return { state: this._nodeMap.get(target)!, target };
 		}
-		if (target.parentElement) {
-			return this._getData(target.parentElement);
-		}
+		// IE 11 does not support parentElement for SVGElement, so in case of parentElement undefined
+        // checking for parentNode.
+        const parent = (target.parentElement || target.parentNode) as HTMLElement;
+        if (parent) {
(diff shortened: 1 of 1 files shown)


src/meta/Drag.ts
```diff
@@ -153,9 +153,12 @@ class DragController {
 		if (this._nodeMap.has(target)) {
 			return { state: this._nodeMap.get(target)!, target };
 		}
-		if (target.parentElement) {
-			return this._getData(target.parentElement);
-		}
+		// IE 11 does not support parentElement for SVGElement, so in case of parentElement undefined
+        // checking for parentNode.
```
dylans: Need to set your editor's linting rules to match what Dojo provides... need tabs instead of spaces for this PR please. Otherwise this change looks fine to me!

agubler:
Surprising as prettier should have formatted the file on commit. 

We do need a test to cover these changes please.

src/meta/Drag.ts
SagarKharche: Closing this PR because of author email address is my wrong in commit. I will raise another PR with above changes.

SagarKharche:
Closing this PR because of author email address is my wrong in commit

PR #939, opened 2018-06-07
Feature: Focus Mixin

agubler:
**Type:** feature

**Description:**

`FocusMixin` builds on the primitive `focus` functionality provided by `vdom` and enables management of focus across widget boundaries.

The `FocusMixin` provides a function called `shouldFocus` that manages when a widget or their child widgets can apply focus, `shouldFocus` will only return `true` the first time it is called until `focus` is called again on the widget.

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,27 +708,27 @@ An example usage controlling focus across child VNodes (DOM) and WNodes (widgets
 					// VNode child.
 					w(FocusInputChild, {
 						key: 0,
-						focus: this._focusedInputKey === 0 && this.shouldFocus,
+						focus: this._focusedInputKey === 0 ? this.shouldFocus : undefined,
 						onFocus: () => this._onFocus(0)
 					}),
 					w(FocusInputChild, {
 						key: 1,
-						focus: this._focusedInputKey === 1 && this.shouldFocus,
+						focus: this._focusedInputKey === 1 ? this.shouldFocus : undefined,
 						onFocus: () => this._onFocus(1)
 					}),
 					w(FocusInputChild, {
 						key: 2,
-						focus: this._focusedInputKey === 2 && this.shouldFocus,
+						focus: this._focusedInputKey === 2 ? this.shouldFocus : undefined,
 						onFocus: () => this._onFocus(2)
 					}),
 					w(FocusInputChild, {
 						key: 3,
-						focus: this._focusedInputKey === 3 && this.shouldFocus,
+						focus: this._focusedInputKey === 3 ? this.shouldFocus : undefined,
 						onFocus: () => this._onFocus(3)
 					}),
 					v('input', {
 						key: 4,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -629,6 +630,114 @@ These are some of the **important** principles to keep in mind when creating and
...
+```
+
+```ts
+// Call focus for the node on every render
+v('input', { type: 'text', focus: () => true) })
+```
+
+This primitive is a base that enables building further abstractions to handle more complex behaviors, one of which is handling focus across the boundaries of encapsulated widgets. The `FocusMixin` is designed to provide support for these scenarios and should be used by widgets that want to provide `focus` to its children or can accept `focus` from a parent widget.
+
+The mixin enhances a widgets API adding `focus` and `shouldFocus` methods, `shouldFocus` checks whether the widget is in a state to perform a focus action and will only return `true` once until the the widget's `focus` method has been called. This `shouldFocus` method is designed to be passed to child widgets or nodes as the value of the `focus` property.
```
tomdye: change:
The mixin enhances a widgets API adding `focus` and `shouldFocus` methods,

to:
The mixin enhances a widgets API by adding `focus` and `shouldFocus` methods.

README.md
```diff
@@ -629,6 +630,114 @@ These are some of the **important** principles to keep in mind when creating and
...
+// Call focus for the node on the only first render
+v('input', { type: 'text', focus: true })
+```
+
+```ts
+// Call focus for the node on every render
+v('input', { type: 'text', focus: () => true) })
+```
+
+This primitive is a base that enables building further abstractions to handle more complex behaviors, one of which is handling focus across the boundaries of encapsulated widgets. The `FocusMixin` is designed to provide support for these scenarios and should be used by widgets that want to provide `focus` to its children or can accept `focus` from a parent widget.
```
tomdye: This primitive is a base that enables further abstractions to be built to handle more complex behaviours. One of these is handling focus across the boundaries of encapsulated widgets. The `FocusMixin` should be used by widgets to provide `focus` to their children or to accept `focus` from a parent widget.

README.md
tomdye: the `FocusMixin` adds `focus` and `shouldFocus` properties to a widget's API. `shouldFocus` checks if the widget is in a state to perform a focus action and will only return `true` once the widget's `focus` method has been called. This `shouldFocus` method is designed to be passed to child widgets or nodes are the value of their `focus` property.

src/mixins/Focus.ts
```diff
@@ -0,0 +1,50 @@
+import { Constructor } from './../interfaces';
+import { WidgetBase } from './../WidgetBase';
+import { diffProperty } from './../decorators/diffProperty';
+
+export interface FocusProperties {
+	focus?: (() => boolean) | boolean;
```
agubler: this should be `false` we shouldn't allow `true` here

PR #946, opened 2018-06-14
Updating Dimensions meta to include client dimensions

rorticus:
**Type:** feature

**Description:**

I recently came across a need for `clientWidth` but realized that the `Dimensions` meta does not return it. This simply adds `clientLeft`, `clientTop`, `clientWidth`, and `clientHeight` to the results returned by the `Dimensions` meta.

### DIFF
--- a/src/meta/Dimensions.ts
+++ b/src/meta/Dimensions.ts
@@ -21,9 +21,16 @@ export interface DimensionResults {
 	offset: TopLeft & Size;
 	size: Size;
 	scroll: TopLeft & Size;
+	client: TopLeft & Size;
 }
 
 const defaultDimensions = {
+	client: {
+		height: 0,
+		left: 0,
+		top: 0,
+		width: 0
+	},
 	offset: {
 		height: 0,
 		left: 0,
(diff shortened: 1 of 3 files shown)


src/meta/Dimensions.ts
```diff
@@ -21,9 +21,16 @@ export interface DimensionResults {
 	offset: TopLeft & Size;
 	size: Size;
 	scroll: TopLeft & Size;
+	client: TopLeft & Size;
```
agubler: Could this be breaking, if a consumer is importing and creating an object using this interface?

rorticus: I guess I'm not sure how since we are only adding data. As long as they are using the DimensionResults type I think they should be OK?

agubler: Well we export this interface, so if someone decided to import it and use it to construct their own object... adding this non optional property would cause compilation errors. An edge case for sure (and one maybe we don't need to consider) as consumers should just be getting the object from dimensions

```ts
const myDimensions: DimensionResults = {
    size: { /* stuff */},
    offset: { /* stuff */},
    scroll: { /* stuff */},
    position: { /* stuff */}
}
```

rorticus: Oh! Yeah, I get it.  That case would indeed create an error.

rorticus: I should mention that there is no real rush on this, we can wait until we add more breaking changes to include this!

PR #951, opened 2018-06-21
I18nMixin: render attributes only when set

### LINKED ISSUE #904: Do not default `lang` attribute to null as this causes failing a11y audits
**Bug**

Passing `null` as the default `lang` property in the `I18nMixin` causes a string of `null` to be applied as the `lang` attribute. This causes a11y audits to fail (using chrome dev tools). This should probably be defaulted to a zero length string.

mwistrand:
**Type:** bug

**Description:**

Resolves #904 

Update the I18nMixin to render the `dir` and `lang` attributes only when they are explicitly set.

Tested both with a fresh `dojo create` app and with [todo-mvc-kitchensink](https://github.com/dojo/examples/tree/master/todo-mvc-kitchensink).

### DIFF
--- a/src/mixins/I18n.ts
+++ b/src/mixins/I18n.ts
@@ -41,7 +41,7 @@ export interface I18nProperties extends LocaleData, WidgetProperties {
  */
 interface I18nVNodeProperties extends VNodeProperties {
 	dir: string;
-	lang: string | null;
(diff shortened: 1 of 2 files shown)


mwistrand:
Looks like @rorticus beat me to it.

